PR #5, opened 2023-07-26
Support ending sessions with Ctrl-D

chadwhitacre:
Mentat currently supports using Ctrl-C or 'q' to end a session:

Screenshot 2023-07-26 at 5 17 22 PM

However, these ol' fingers are long habituated to using Ctrl-D to end interactive sessions. This PR adds support for `EOFError` to end a session.

### DIFF
--- a/.github/workflows/lint_and_test.yml
+++ b/.github/workflows/lint_and_test.yml
@@ -3,7 +3,7 @@
 
 name: lint and test
 
-on: [push]
+on: [push, pull_request]
 
 jobs:
   ruff-black-isort:
--- a/mentat/app.py
+++ b/mentat/app.py
@@ -33,7 +33,7 @@ def run(paths: Iterable[str]):
     cost_tracker = CostTracker()
     try:
         loop(paths, cost_tracker)
-    except KeyboardInterrupt as e:
+    except (EOFError, KeyboardInterrupt) as e:
         print(e)
     finally:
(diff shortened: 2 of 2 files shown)


mentat/parsing.py
```diff
@@ -132,7 +133,7 @@ def run_async_stream_and_parse_llm_response(
     except openai.error.InvalidRequestError as e:
         cprint(e, "red")
         exit()
-    except KeyboardInterrupt:
+    except UserInterrupt:
```
chadwhitacre, line 136: Might want both `KeyboardInterrupt` and `UserInterrupt` here?

Probably at least `KeyboardInterrupt`, actually.

chadwhitacre:
Sadly couldn't use Mentat for this because of #6.

mentat/parsing.py
chadwhitacre: Merp, this is overcomplicated. Simplified greatly in 110fffef49f1f74ad510d8eb84cbccf807fbae42. :)

biobootloader:
Yay, first PR! 🎉 

Let me figure out why checks aren't running and we'll merge it!

Ctrl-C is also be used to interrupt the models response. `EOFError` wouldn't apply there since we aren't getting user input. But that's a fairly rare thing to need to do anyway.

biobootloader (approved):

PR #7, opened 2023-07-27
Fix rel_path vs abs_path typo

yoDon:
code_file_manager _read_file() expects to receive an absolute path but was accidentally being handed a relative path immediately after the absolute path was computed. 

This caused mentat to be unable to apply changes to files if mentat was launched with a path other than "."


PCSwingle (approved):
Thanks for catching this!

PR #8, opened 2023-07-27
Improve handling of large projects

yoDon:
Summary:

Adds several optional configuration properties to improve handling of large projects to work around limits in the number of tokens that can be handled by GPT-4 and limits in the number of files that can be passed to the git `repo.ignored()` function.

From the modified `docs/configuration.md`:

### Filepath Include and Exclude list
Optional lists of regex patterns that determine which files or directories will be included 
or excluded from context. Either or both lists can be provided. If the file path matches 
both an inclusion rule and an exclusion rule, the exclusion rule always wins (including 
for example filetype exclusion wins over filepath inclusion).
```
{
    "filepath-include-only-these-regex-patterns": [".*/include_this"],
    "filepath-exclude-these-regex-patterns": [".*/exclude_this"]
}
```

### Disable .gitignore processing
There is currently an issue where large projects can overwhelm the test for .gitignore exclusion. This option
disables .gitignore testing to support working with those larger projects. Default is false, which
allows mentat to make use of git ignore testing.
```
{
    "do-not-check-git-ignored": true,
}
```

### DIFF
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -22,25 +22,11 @@ Determines which file types inside of a directory will be automatically included
 }
 ```
 
-### Filepath Include and Exclude list
-Optional lists of regex patterns that determin which files or directories will be included 
-or excluded from context. Either or both lists can be provided. If the file path matches 
-both an inclusion rule and an exclusion rule, the exclusion rule always wins (including 
-for example filetype exclusion wins over filepath inclusion).
+### File Exclude Glob list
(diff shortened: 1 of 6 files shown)


PCSwingle:
Looks good! I made a few changes:
1. Removed the git ignore change; #13 fixes this now.
2. Changed to using glob instead of regex (makes more sense for file matching)
3. Removed the regex include config; it felt unecessary.

yoDon:
I like it - these edits feel much more pythonic in style

biobootloader (approved):
Thank you for adding this feature!! 🚀

PR #9, opened 2023-07-27
make required python version 3.10 or higher

PCSwingle:
Tested with pyenv, pip won't let you install packages if your python version is too low. There is no point in making a version check in mentat since it won't even run below 3.10 because of the match statements.


biobootloader:
I added a version check so it doesn't crash like below, which is a lot less clear:
image

biobootloader (approved):

PR #13, opened 2023-07-27
Fix bug filter gitignore sooner

swinglejohn:
Fix the bug with too many files in large directories. This bug was caused because we traversed the entire file path tree without ignoring .gitignore files. Now we do those together.


PCSwingle (approved):
LGTM :rock: :rocket:

PR #28, opened 2023-07-29
Git migration

PCSwingle:
Move git root handling outside of code_file_manager


biobootloader:
Good stuff! 

I have a thought on improvement:

Symptom: `git_root` getting passed around everywhere (i.e., stored in each CodeChange)

Root Problem: CodeFileManager.file_lines maps full paths to files.

Solution: Change CodeFileManager to map just relative paths, so the only place the git_root is needed is when files are written and read, which is only done internally by the CodeFileManager

biobootloader (approved):
my suggested change would be best done in a separate PR

PR #29, opened 2023-07-29
The great config migration

PCSwingle:
Change config name to .mentat_config.json and allow project specific config


biobootloader:
image

biobootloader (approved):
nice!

Would it be easy to add a depreciation warning if the user has an old `config.json`?

PR #32, opened 2023-07-29
Remove filetype checking

PCSwingle:
Remove filetype checking, add glob includes, and fix a few bugs. To test include globs in zsh, put quotes around paths.

In case you're wondering why the glob path expansion is at the very beginning (outside of codefilemanager), there are 2 reasons:
1. We need it to resolve the git root
2. Because some terminals (such as zsh) automatically expand glob patterns, I think it's best to do it at the very beginning so there is no difference between terminals.

### DIFF
--- a/README.md
+++ b/README.md
@@ -77,4 +77,4 @@ Run Mentat from within your project directory. Mentat uses git, so if your proje
 
 `mentat <paths to files or directories>`
 
-If you provide a directory, Mentat will add all non-hidden text files in that directory to it's context. If this exceeds the GPT-4 token context limit, try running Mentat with just the files you need it to see.
+If you provide a directory, Mentat will add all non git-ignored text files in that directory to it's context. If this exceeds the GPT-4 token context limit, try running Mentat with just the files you need it to see.
--- a/README.md
+++ b/README.md
@@ -77,4 +77,4 @@ Run Mentat from within your project directory. Mentat uses git, so if your proje
 
 `mentat <paths to files or directories>`
 
-If you provide a directory, Mentat will add all non git-ignored text files in that directory to it's context. If this exceeds the GPT-4 token context limit, try running Mentat with just the files you need it to see.
+If you provide a directory, Mentat will add all non git-ignored text files in that directory to it's context. If you provide a glob pattern, Mentat will add all files matching the pattern to it's context. If this exceeds the GPT-4 token context limit, try running Mentat with just the files you need it to see.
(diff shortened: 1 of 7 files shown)


docs/configuration.md
```diff
@@ -13,20 +13,11 @@ Allow Mentat to use OpenAI's gpt-4 32k context window model. Your API key must a
...
-```
-{
-    "filetype-include-list": [".include_this"],
-    "filetype-exclude-list": [".exclude_this"]
-}
-```
-
 ### File Exclude Glob list
-List of [glob patterns](https://docs.python.org/3/library/glob.html) that will exclude all files from context that it matches starting from the given directory.
+List of [glob patterns](https://docs.python.org/3/library/glob.html) that will exclude all files from context that it matches starting from the git root. Importantly, this means that the pattern to exclude all files ending with `.py` would be `**/*.py` rather than `*.py`. Here is an example that would exclude all hidden directories and files:
```
biobootloader, line 17:
```suggestion
List of [glob patterns](https://docs.python.org/3/library/glob.html) to exclude files from being read/edited by Mentat. These take effect whenever you provide Mentat with a directory as an argument. Mentat will add all files in the directory that are not in your `.gitignore` and do not match these glob patterns. Glob patterns are interpreted from the git root location. If you wanted to exclude all files ending in `.py`, the pattern to use would be `**/*.py` rather than `*.py`. Here is an example that would exclude all hidden directories and files:
```

mentat/code_file_manager.py
```diff
@@ -22,41 +21,6 @@
...
-    ".r",       # R
-    ".m",       # objective-c
-    ".sh",      # shell scripts
-    ".f",       # fortran
-    ".jsx",     # javascript react
-    ".tsx",     # typescript react
-]
-default_filetype_exclude_list = []
-# fmt: on
-
```
biobootloader, line 59: 🥳

biobootloader (approved):

biobootloader:
Massive.

README.md
```diff
@@ -77,4 +77,4 @@ Run Mentat from within your project directory. Mentat uses git, so if your proje
 
 `mentat <paths to files or directories>`
 
-If you provide a directory, Mentat will add all non-hidden text files in that directory to it's context. If this exceeds the GPT-4 token context limit, try running Mentat with just the files you need it to see.
+If you provide a directory, Mentat will add all non git-ignored text files in that directory to it's context. If you provide a glob pattern, Mentat will add all files matching the pattern to it's context. If this exceeds the GPT-4 token context limit, try running Mentat with just the files you need it to see.
```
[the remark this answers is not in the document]
PCSwingle: Why are we not mentioning that you an use globs as file paths?

PR #33, opened 2023-07-30
Exclude globs

PCSwingle:
Adds -e --exclude option. Merge after #32

### DIFF
--- a/mentat/code_file_manager.py
+++ b/mentat/code_file_manager.py
@@ -142,8 +142,7 @@ class CodeFileManager:
         excluded_files, excluded_files_from_dir = _abs_file_paths_from_list(
             exclude_paths, check_for_text=False
         )
-        # excluded paths should have already been expanded
-        assert not excluded_files_from_dir
+
         glob_excluded_files = set(
             os.path.join(self.git_root, file)
             for glob_path in self.config.file_exclude_glob_list()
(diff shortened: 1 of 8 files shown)


biobootloader (approved):
@PCSwingle I made some changes, view changes from just my commits [here](https://github.com/biobootloader/mentat/pull/33/files/6a545cdf758c75243208f498e06cae2d7084961f..87690ae74a544c5d456a26329e62b9600897339d).
- made `exclude_paths` optional in run and loop, cleaning up tests.
- changed precedence so the config exclude globs only exclude things added via directories.
- updated readme and configuration.md to clarify

If my changes look good go ahead and merge! 🚀

mentat/code_file_manager.py
```diff
@@ -138,7 +139,11 @@ def _set_file_paths(
             print("Exiting...")
             exit()
 
-        excluded_files = _abs_file_paths_from_list(exclude_paths, check_for_text=False)
+        excluded_files, excluded_files_from_dir = _abs_file_paths_from_list(
+            exclude_paths, check_for_text=False
+        )
+        # excluded paths should have already been expanded
+        assert not excluded_files_from_dir
```
PCSwingle, line 146: ? Why is this? What if the user tries to --exclude directory?

biobootloader: Oh oops. You want to fix?

PCSwingle: Yeah i'll get it

PR #37, opened 2023-07-30
Model error handling

PCSwingle:
Adds error handling for if the model messes up; if it's a non-fatal error, will keep accepting the models changes, otherwise it will stop and only use the existing changes.

I think the error code for codechange feels very clunky and messy; for the non fatal errors we can't throw an exception though, and this was the best I thought of. Let me know if you have any ideas to make it cleaner.

### DIFF
--- a/mentat/code_change.py
+++ b/mentat/code_change.py
@@ -65,30 +65,43 @@ class CodeChange:
                 f"Model created change with unknown action {self.json_data['action']}",
                 unfinished_change=False,
             )
-        match self.action:
-            case CodeChangeAction.Insert:
-                if "insert-before-line" in self.json_data:
-                    self.first_changed_line = self.json_data["insert-before-line"]
-                    if "insert-after-line" in self.json_data:
-                        if (
-                            self.first_changed_line - 1
-                            != self.json_data["insert-after-line"]
-                        ):
-                            self.error = "Insert line numbers invalid"
-                elif "insert-after-line" in self.json_data:
-                    self.first_changed_line = self.json_data["insert-after-line"] + 1
-                else:
-                    self.error = "Insert line number not specified"
-                self.first_changed_line -= 0.5
-                self.last_changed_line = self.first_changed_line
 
-            case CodeChangeAction.Replace:
-                self.first_changed_line = self.json_data["start-line"]
-                self.last_changed_line = self.json_data["end-line"]
+        try:
+            match self.action:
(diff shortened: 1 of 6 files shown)


mentat/model_error.py
```diff
@@ -0,0 +1,5 @@
+# Raised when model messes up creating a change
```
biobootloader, line 1:
```suggestion
# Raised when model output doesn't adhere to the specified format for changes
```

mentat/code_change.py
```diff
@@ -59,52 +58,64 @@ def __init__(
...
+
+        if self.action != CodeChangeAction.CreateFile:
             file_path = os.path.join(self.git_root, self.file)
             try:
                 self.file_lines = code_file_manager.file_lines[file_path]
+                self.line_number_buffer = len(str(len(self.file_lines) + 1)) + 1
             except KeyError:
-                raise Exception(f"File {file_path} not found!")
-            self.line_number_buffer = len(str(len(self.file_lines) + 1)) + 1
+                self.error = f"File {file_path} not included in context"
```
biobootloader, line 112:
```suggestion
                self.error = f"Model attempt to edit {file_path}, which isn't in current context or doesn't exist"
```

mentat/code_change.py
```diff
@@ -59,52 +58,64 @@ def __init__(
...
         else:
+            if os.path.exists(self.file):
+                self.error = (
+                    f"Model attempted to create file {self.file} that already exists"
```
biobootloader, line 116:
```suggestion
                    f"Model attempted to create file that already exists: {self.file}"
```

mentat/parsing.py
```diff
@@ -155,11 +190,21 @@ async def stream_and_parse_llm_response(
...
-
-    logging.debug(f"LLM response:\n{state.message}")
+    try:
+        await _process_response(state, response, printer, code_file_manager)
+        printer.wrap_it_up()
+        await printer_task
+    except ModelError as e:
+        logging.info(f"Model created error {e}")
+        printer.wrap_it_up()
+        await printer_task
```
biobootloader, line 200:
```suggestion
        # make sure we finish printing everything model sent before we encountered the crash
        await printer_task
```

biobootloader (commented):
maybe we could have a more descriptive name for `unfinished_change`?

biobootloader (approved):

PR #40, opened 2023-07-31
Fix typo in readme for --exclude

yoDon:
The --exclude flag needs to be preceded by a directory path. Without a directory path the mentat command returns a no context provided error.


biobootloader (approved):
Thanks for catching this! 🙏

PR #42, opened 2023-08-01
Check early if file tokens exceed or near limit

biobootloader:


mentat/app.py
```diff
@@ -57,7 +57,7 @@ def run(paths: Iterable[str], exclude_paths: Optional[Iterable[str]] = None):
     try:
         loop(paths, exclude_paths, cost_tracker)
     except (EOFError, KeyboardInterrupt) as e:
-        print(e)
+        cprint("\n" + str(e), color="red")
```
PCSwingle, line 60: ? Why is this here? All it does for me is print a lot of newlines

PCSwingle (approved):
LGTM; feel free to merge

biobootloader:
I just realized we need to take into account the prompt tokens when doing this check, I'll update before merging

mentat/app.py
biobootloader: If the error has a message (like the new ones I introduced) then it prints it here - you shouldn't be getting more than 2 new lines even if there error has no message, right?

The "\n" is there to separate the error from the file tree

PCSwingle: But the error is EOF or KeyboardInterrupt; why would it have a message?

PR #44, opened 2023-08-01
Fix a type mismatch

chadwhitacre:
Without this I'm getting:

```
  File "/Volumes/workbench/biobootloader/mentat/mentat/code_change.py", line 77, in __init__
    self.first_changed_line - 1
    ~~~~~~~~~~~~~~~~~~~~~~~~^~~
TypeError: unsupported operand type(s) for -: 'str' and 'int'
```

https://github.com/biobootloader/mentat/blob/c91548cd4d2f45e944428c79e28a71091431641d/mentat/code_change.py#L77

I don't have a good test case, sorry, just had time for a quick PR.

### DIFF
--- a/mentat/code_change.py
+++ b/mentat/code_change.py
@@ -44,10 +44,15 @@ class CodeChange:
         code_file_manager,
     ):
         self.json_data = json_data
-        # Scab on some simple type coercion.
-        for k in "insert-before-line insert-after-line start-line end-line".split():
-            if k in self.json_data:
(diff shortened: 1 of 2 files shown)


PCSwingle:
I've never seen this before, but I assume it's because sometimes gpt puts quotes around the numbers? I'll add a test for this and then feel free to merge.

PCSwingle (approved):
:partying_face:

PR #45, opened 2023-08-02
Use custom exceptions instead of KeyboardInterrupts and exit()

swinglejohn:
Don't raise KeyboardInterrupts. Also raise exceptions instead of exit() so we can print the final cost.

### DIFF
--- a/mentat/code_file_manager.py
+++ b/mentat/code_file_manager.py
@@ -125,17 +125,6 @@ class CodeFileManager:
     def _set_file_paths(
         self, paths: Iterable[str], exclude_paths: Iterable[str]
     ) -> None:
-        invalid_paths = []
-        for path in paths:
-            if not os.path.exists(path):
-                invalid_paths.append(path)
-        if invalid_paths:
-            cprint("Error:", "red", end=" ")
-            cprint("The following paths do not exist:")
-            print("\n".join(invalid_paths))
-            print("Exiting...")
-            raise UserError()
-
         excluded_files, excluded_files_from_dir = _abs_file_paths_from_list(
             exclude_paths, check_for_text=False
         )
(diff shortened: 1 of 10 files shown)


mentat/code_file_manager.py
```diff
@@ -137,7 +134,7 @@ def _set_file_paths(
             cprint("The following paths do not exist:")
             print("\n".join(invalid_paths))
             print("Exiting...")
-            exit()
+            raise UserError()
```
PCSwingle, line 137: We could change the error to be the following paths don't exist instead of cprinting, and maybe get rid of the 'exiting...'? Since that should be standard (either always have it, or always don't have it).

mentat/code_file_manager.py
```diff
@@ -249,7 +246,7 @@ def _get_new_code_lines(self, changes) -> Iterable[str]:
         min_changed_line = largest_changed_line + 1
         for i, change in enumerate(changes):
             if change.last_changed_line >= min_changed_line:
-                raise ValueError(f"Change line number overlap in file {change.file}")
+                raise UserError(f"Change line number overlap in file {change.file}")
```
PCSwingle, line 249: This should be a MentatError; the collision handling is supposed to handle this

mentat/llm_api.py
```diff
@@ -24,12 +25,10 @@ def setup_api_key():
         openai.api_key = key
         openai.Model.list()  # Test the API key
     except openai.error.AuthenticationError:
-        cprint(
+        raise UserError(
```
PCSwingle, line 28: We call this function before the try except in app.py; we should move that call into loop() so this error gets properly caught

swinglejohn: Good catch

mentat/code_file_manager.py
swinglejohn: It turns out the warning for bad file paths got broken when we switched to globbing so I moved this chunk of code to expand_paths() in app.py.

PCSwingle (approved):
:rocket: :rock:

PR #51, opened 2023-08-05
Add a benchmark test for reading/ parsing data

swinglejohn:
Adds a benchmark test requiring the Mentat to create a class and function to read, parse, and store data from a json file. The data is a set of people with sever attributes. The hardest part for Mentat turned out to be the parsing (originally I thought it would be the class architecture). The hard part about the parsing for Mentat is people in a family are represented differently than single people. Most of the mistakes Mentat makes are just 5 or 6 different things.

I created 4 different prompts with varying levels of prompt engineering. The first three I tested 20 times each and they had these pass rates:
easy - 95%
easy_medium - 75%
medium - 40%
hard - probably 0% (I didn't record the tests)

I've set the test to use the easy_medium prompt for now because I think we could improve its performance just by improving the general Mentat prompt. But one thing I was considering was moving this test to another file and breaking it out into 1-3 tests so we could see how all of the prompts do across updates to the general Mentat prompt.

### DIFF
--- a/testbed/person_data/data.json
+++ b/testbed/person_data/data.json
@@ -1,0 +1,20 @@
+{
+    "People": [
+        {
+            "parents": [
+                {"name": "Henry", "age": 35, "weight": 235, "married": true},
+                {"name": "Martha", "age": 36, "weight": 140, "married": true}
+            ],
+            "children": [
+                {"name": "Steve", "age": 15, "weight": 120, "married": false},
+                {"name": "Alex", "age": 12, "weight": 110, "married": false},
+                {"name": "Sarah", "age": 9, "weight": 70, "married": false}
+            ]
+        },
+        { "name": "Fridge", "age": 42, "weight": 180, "married": false },
+        { "name": "Jane", "age": 57, "weight": 140, "married": true },
+        { "name": "Jane", "age": 58, "weight": 140, "married": true },
+        { "name": "Jane", "age": 58, "weight": 141, "married": true },
+        { "name": "Jane", "age": 58, "weight": 141, "married": false}
+    ]
+}
--- a/testbed/person_data/data.json
+++ b/testbed/person_data/data.json
@@ -2,7 +2,7 @@
     "People": [
         {
             "parents": [
-                {"name": "Henry", "age": 35, "weight": 235, "married": true},
+                {"name": "Ben", "age": 35, "weight": 235, "married": true},
                 {"name": "Martha", "age": 36, "weight": 140, "married": true}
             ],
             "children": [
@@ -11,10 +11,10 @@
                 {"name": "Sarah", "age": 9, "weight": 70, "married": false}
             ]
         },
-        { "name": "Fridge", "age": 42, "weight": 180, "married": false },
-        { "name": "Jane", "age": 57, "weight": 140, "married": true },
-        { "name": "Jane", "age": 58, "weight": 140, "married": true },
-        { "name": "Jane", "age": 58, "weight": 141, "married": true },
-        { "name": "Jane", "age": 58, "weight": 141, "married": false}
+        { "name": "Fridge", "age": 57, "weight": 140, "married": false},
+        { "name": "Jane", "age": 57, "weight": 140, "married": false},
+        { "name": "Jane", "age": 58, "weight": 140, "married": false},
+        { "name": "Jane", "age": 58, "weight": 141, "married": false},
+        { "name": "Jane", "age": 58, "weight": 141, "married": true}
     ]
 }
--- a/testbed/person_data/test_data.json
+++ b/testbed/person_data/test_data.json
@@ -1,0 +1,20 @@
+{
+    "People": [
+        {
+            "parents": [
+                {"name": "Henry", "age": 35, "weight": 235, "married": true},
+                {"name": "Martha", "age": 136, "weight": 140, "married": true}
+            ],
+            "children": [
+                {"name": "Steve", "age": 15, "weight": 120, "married": false},
+                {"name": "Alex", "age": 12, "weight": 110, "married": false},
+                {"name": "Sarah", "age": 9, "weight": 70, "married": false}
+            ]
+        },
+        { "name": "Fridge", "age": 42, "weight": 180, "married": false },
+        { "name": "Jane", "age": 57, "weight": 140, "married": true },
+        { "name": "Jane", "age": 58, "weight": 140, "married": true },
+        { "name": "Jane", "age": 58, "weight": 141, "married": true },
+        { "name": "Jane", "age": 58, "weight": 141, "married": false}
+    ]
+}
--- a/testbed/person_data/test_data.json
+++ b/testbed/person_data/test_data.json
@@ -2,7 +2,7 @@
     "People": [
         {
             "parents": [
-                {"name": "Henry", "age": 35, "weight": 235, "married": true},
+                {"name": "Ben", "age": 35, "weight": 235, "married": true},
                 {"name": "Martha", "age": 136, "weight": 140, "married": true}
             ],
             "children": [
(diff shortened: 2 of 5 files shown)


tests/benchmark_test.py
```diff
@@ -131,3 +133,131 @@ def test_start_project_from_scratch(mock_collect_user_input):
...
+    hard_prompt = (
+        "Fill out the Person class for this data. It needs load_data and __eq__"
+        " functions. The equality functions should compare all attributes and"
+        " recursively compare the parents. A Person should have individual mother"
+        " and father attributes. The load_data function should be a class method of"
+        " Person. The load function will take the data file's path as an argument."
+        " Note that the father is listed"
+        " first in a parents list."
+    )
+    # makes linter shut up
```
PCSwingle, line 198: You can use # noqa XXX to disable linting warnings on specific lines

tests/benchmark_test.py
```diff
@@ -131,3 +133,131 @@ def test_start_project_from_scratch(mock_collect_user_input):
...
+    module = importlib.util.module_from_spec(spec)
+    sys.modules[module_name] = module
+    spec.loader.exec_module(module)
+    Person = module.Person
+    # from testbed.person_data.person import Person
+
+    people = Person.load_data("../testbed/person_data/data.json")
+
+    # One of the most common mistakes Mentat makes:
+    if len(people) == 14:
```
PCSwingle, line 225: You can just do assert len(people) == 14, Error message here

testbed/person_data/data.json
```diff
@@ -0,0 +1,20 @@
+{
+    "People": [
```
PCSwingle, line 2: This data format feels weird to me? Why are we combining a Family with a Person? I think it would make more sense to have the model work on something more realistic / structured. In most json, whenever you have an array it is almost always an array of one thing; if you want multiple things, you would use key value pairs instead. Not the biggest since obviously this is just a benchmark, but I do think it makes more sense to give it something it has not only seen a lot more but also what it will see.

testbed/person_data/data.json
```diff
@@ -0,0 +1,20 @@
...
+            ],
+            "children": [
+                {"name": "Steve", "age": 15, "weight": 120, "married": false},
+                {"name": "Alex", "age": 12, "weight": 110, "married": false},
+                {"name": "Sarah", "age": 9, "weight": 70, "married": false}
+            ]
+        },
+        { "name": "Fridge", "age": 42, "weight": 180, "married": false },
+        { "name": "Jane", "age": 57, "weight": 140, "married": true },
+        { "name": "Jane", "age": 58, "weight": 140, "married": true },
```
PCSwingle, line 16: Also, why are there 5 Janes?

testbed/person_data/test_data.json
PCSwingle, line 1: Is there a reason we need both data files? They look the exact same

tests/benchmark_test.py
```diff
@@ -131,3 +133,131 @@ def test_start_project_from_scratch(mock_collect_user_input):
...
+        easy_medium_prompt,
+        "y",
+        KeyboardInterrupt,
+    ]
+    run(["person_data/data.json", "person_data/person.py"])
+
+    # This imports the Person class from the file mentat edited just created
+    # import in Python is always relative to the file's original path
+    # so even though we change the cwd, we can't use relative imports to import form the tmp dir
+    file_path = os.path.join(os.getcwd(), "person_data/person.py")
```
PCSwingle, line 213: I can see us wanting to import things in benchmarks a lot; could be good to make this a helper function in some test_utils file or something. Not necessary in this PR though if you don't want to.

biobootloader: In this PR lets put this import stuff in a helper function. We can create a `test_utils` in a future PR when we need to use it somewhere else as well (and better understand what we want to abstract).

biobootloader:
Nice benchmark! Certainly more involved than our older benchmarks. We needed this

I'll take another look after you address Paul's comments

testbed/person_data/test_data.json
swinglejohn: The mother's age is different so we can test the comparison function is recursive (the last assert statement in the test).

tests/benchmark_test.py
swinglejohn: Thanks, that's better

testbed/person_data/data.json
swinglejohn: I updated them to get the order right but it's because we want to compare people that differ in only one attribute for each attribute to ensure that the model checks all attributes for equality. So when the people are loaded and sorted the test just checks the neighboring people for equality.

tests/benchmark_test.py
swinglejohn: Done

testbed/person_data/data.json
swinglejohn: It's true it's a weird data format and that's probably why that's the most challenging part of the test for the model. Realism is definitely good (i.e. lining it up with something similar to what it will see) but we don't want to just give it tests that it has seen before and could have memorized. Since the model can see all of the data it should be able to write code to parse arbitrary formatting.
I'm on the fence about the current data format and would like @biobootloader to chime in once he's read it.

PCSwingle: I think my major issue with the current format is that it requires us to prompt engineer it to get anything correct; the entire point of the benchmarks is so that we can see if a specific prompt does better or worse than the last one. With per benchmark prompt engineering I feel like that throws off the measurements a bit and makes it more difficult to tell if a prompt is actually better or worse

testbed/person_data/data.json
PCSwingle: This is probably also confusing to the model? Maybe we should only have the test for it's code do this and not show it this?

testbed/person_data/data.json
swinglejohn: We could certainly use one of the more difficult prompts with fewer hints. I think that might be better than just giving it an easier data format though. Then we could see if the general prompt improves performance at all as opposed to if the benchmark isn't a challenge at all.

biobootloader: I think the ideal benchmark here would be something where the prompt is fairly straightforward, i.e. just what a person would like to be able to ask it to do, without hints it should be able to gather from looking at the data format. And performance should be something like 25%-50%, as that is both 1) high enough that we can see improvements quickly and 2) low enough that we have something to improve on

If simplifying/standarizing the data format is the way to do that, then lets do it

tests/benchmark_test.py
```diff
@@ -135,6 +135,26 @@ def test_start_project_from_scratch(mock_collect_user_input):
     assert result.stdout.strip() == expected_output.strip()
 
 
+def import_module_from_path(module_name, relative_file_path):
+    """
+    Imports a from the file Mentat just edited.
+    import in Python is always relative to the file's original path so
+    even though we change the cwd, we can't use relative imports to import from the tmp dir.
+
+    I don't really know what module_name is for since it even works as the empty string.
```
PCSwingle, line 144: The module name is the name of the actual module you're importing; i.e. 'collections', 'itertools', so on. We should just make testbed the default module_name (so it's almost like we're writing 'from testbed import relative_file_path'). Otherwise we stand the risk of somehow overwriting a module with an empty string as it's name (unlikely, but still good practice I think).

tests/benchmark_test.py
```diff
@@ -195,46 +215,25 @@ def test_create_person_class_from_data(mock_collect_user_input, temp_testbed):
...
-    module_name = "testbed.person_data.person"
-    spec = importlib.util.spec_from_file_location(module_name, file_path)
-    module = importlib.util.module_from_spec(spec)
-    sys.modules[module_name] = module
-    spec.loader.exec_module(module)
-    Person = module.Person
-    # from testbed.person_data.person import Person
+    relative_file_path = "person_data/person.py"
+    Person = import_module_from_path("", relative_file_path).Person
+    # from testbed.person_data.solution_person import Person
```
PCSwingle, line 227: This is more like writing from testbed import person_data.person; Person = person_data.Person

PCSwingle:
Left a couple more comments about the import function; after resolving those and simplifying the test cases like I mentioned earlier this should be good to go.

PCSwingle:
This is fairly outdated and we have new benchmarks now, so I'm marking this closed.

PR #54, opened 2023-08-07
new backup_manager class for backup mechanism of changed code files

dogukantai:
Hi 👋 This simple MR modifies app.py to add two different argument parsers for the backup mechanism that is used when code files change. We now preserve the original backup according to new additional parameters, ensuring we have an accurate state of the file's initial state. Branch is currently WIP.

### DIFF
--- a/mentat/app.py
+++ b/mentat/app.py
@@ -43,17 +43,14 @@ def run_cli():
     )
     args = parser.parse_args()
 
-    backup_dir = ".mentat_backups"
-
     if args.revert:
-        revert_files(backup_dir)
+        revert_files()
     else:
         paths = args.paths
         exclude_paths = args.exclude
         run(
             expand_paths(paths),
             expand_paths(exclude_paths),
-            backup_dir,
         )
 
 
@@ -79,13 +76,12 @@ def expand_paths(paths: Iterable[str]) -> Iterable[str]:
 def run(
     paths: Iterable[str],
     exclude_paths: Optional[Iterable[str]] = None,
-    backup_dir: Optional[str] = ".mentat_backups",
 ):
     os.makedirs(mentat_dir_path, exist_ok=True)
     setup_logging()
     logging.debug(f"Paths: {paths}")
 
-    backup_manager = CodeBackupManager(backup_dir)
+    backup_manager = CodeBackupManager(".mentat_backups")
 
     cost_tracker = CostTracker()
 
@@ -94,7 +90,7 @@ def run(
 
         cprint("mentat started with automatic backup pipeline...\n", color="light_blue")
 
-        loop(paths, exclude_paths, cost_tracker, backup_dir)
+        loop(paths, exclude_paths, cost_tracker)
     except (
         EOFError,
         KeyboardInterrupt,
@@ -112,7 +108,6 @@ def loop(
     paths: Iterable[str],
     exclude_paths: Optional[Iterable[str]],
     cost_tracker: CostTracker,
-    backup_dir: Optional[str] = ".mentat_backups",
 ) -> None:
     git_root = get_shared_git_root_for_paths(paths)
     config = ConfigManager(git_root)
@@ -142,7 +137,6 @@ def loop(
                 user_input_manager,
                 code_file_manager,
                 code_changes,
-                backup_dir,
             )
         else:
             need_user_request = True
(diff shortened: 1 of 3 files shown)


biobootloader:
Thank you! This is an important feature

A few thoughts, @PCSwingle @swinglejohn let me know what you think too:

- what if backup was always on? and it always went to a directory named something like `.mentat_backup`? This would simplify the interface and implementation (no need for the two command line args `preserve-backup` and `backup-dir`)
- add a command line arg to restore from backup, run simply as `mentat --revert`?
- would it be fine to just have a single backup stored? Keeping multiple adds complexity and the user has to select one. Also, since Mentat only work in git projects, they do have version control.

simugrad:
> - what if backup was always on? and it always went to a directory named something like .mentat_backup? This would simplify the interface and implementation (no need for the two command line args preserve-backup and backup-dir)
> - add a command line arg to restore from backup, run simply as mentat --revert?

In fact, it makes sense to me for the backup to be "on" by default at first, because there is a possibility that the backup parameters may be overlooked by the user. Maybe we can enable it by asking the user in the initialization step of the mentat? Also, reverting control via mentat arguments, would be really nice though.

> - would it be fine to just have a single backup stored? Keeping multiple adds complexity and the user has to select one. Also, since Mentat only work in git projects, they do have version control.

This can make the backup mechanism more consistent and mature in the overall lifecycle of app.

I can move forward with these changes or with any other options you can think of? 👀

mentat/app.py
```diff
@@ -35,10 +36,35 @@ def run_cli():
...
+        "--no-backup",
+        action="store_false",
+        help="No backup mode for the mentat",
+    )
+    parser.add_argument(
+        "--backup-dir",
+        type=str,
+        default=".mentat_backups",
+        help="Directory to store backups, if not provided, backups will be stored in the .mentat_backups directory",
+    )
```
biobootloader, line 49: let's remove these options to keep things simple and nimble for now. We'll just always backup (as I don't see a downside) and always use `.mentat_backups` to store the backup

mentat/app.py
```diff
@@ -171,6 +225,62 @@ def get_user_feedback_on_changes(
...
+    backup_manager = CodeBackupManager(backup_dir)
+    backup_manager.backup_files(code_file_manager)
+
+
+def user_select_files_to_revert(available_backups: list) -> list:
+    cprint(
+        "Enter the numbers of the backup files you wish to revert (e.g., '1 3 4'). Press Enter to revert none.",
+        color="light_blue",
+    )
+    user_response = input()
```
biobootloader, line 240: Let's make "Press Enter" revert all files, and use Ctrl-C to cancel reversion

simugrad:
Thanks for the reviews @biobootloader
Review change modifications have been moved to the MR.

mentat/app.py
```diff
@@ -60,15 +76,25 @@ def expand_paths(paths: Iterable[str]) -> Iterable[str]:
     return globbed_paths
 
 
-def run(paths: Iterable[str], exclude_paths: Optional[Iterable[str]] = None):
+def run(
+    paths: Iterable[str],
+    exclude_paths: Optional[Iterable[str]] = None,
+    backup_dir: Optional[str] = ".mentat_backups",
```
PCSwingle, line 82: I don't think this should be 'Optional'; it should always be a string (never None).

mentat/app.py
```diff
@@ -35,10 +36,25 @@ def run_cli():
...
+
+    if args.revert:
+        revert_files(backup_dir)
+    else:
+        paths = args.paths
+        exclude_paths = args.exclude
+        run(
+            expand_paths(paths),
+            expand_paths(exclude_paths),
+            backup_dir,
```
PCSwingle, line 56: Let's get rid of the backup_dir variable up here and in the run function; it feels a bit weird to have 3 default values for it (one in the run function, one up here, and one in the constructor for backupmanager).

PCSwingle:
Hey @simugrad! Really glad you're making this; definitely an essential feature. Just made a few suggestions for your PR; let me know if you have any questions about anything. I also wouldn't mind working on the suggestions myself; just let me know what you would prefer. Thanks!

simugrad:
> Hey @simugrad! Really glad you're making this; definitely an essential feature. Just made a few suggestions for your PR; let me know if you have any questions about anything. I also wouldn't mind working on the suggestions myself; just let me know what you would prefer. Thanks!

Hey @PCSwingle 👋 Thanks for the quick review! Just resolved your suggestions, because we got rid of the backup_dir argument parser and it will be a constant, I declared it as hard-coded. I agree with you! Also, I'm open to any suggestions!

biobootloader:
@finstatespace 

> Apart from this MR, I would like to discussion some ideas under an issue for the implementation and modularization of other useful features. Shall I bring this?

I'd love to hear your ideas, how about you post on the Discord?

biobootloader:
@simugrad looks like tests are failing, how about you fix these and then either @PCSwingle or I will take another look!

biobootloader:
closing as stale and now we have /undo command

PR #59, opened 2023-08-12
Add commands

PCSwingle:


mentat/commands.py
```diff
@@ -0,0 +1,104 @@
...
+    @classmethod
+    def get_command_completions(cls) -> Iterable[str]:
+        return list(map(lambda name: "/" + name, cls._registered_commands))
+
+    # *args[0] is always the command_name, c style
+    @abstractmethod
+    def apply(self, *args: str) -> None:
+        pass
+
+    # TODO: make more robust way to specify arguments for commands
```
PCSwingle, line 35: Wanted some outside input on this, but other than this this PR should be good to merge. I have a few ideas for how we could do arguments that I think could be better; for one, we could have an 'argumentvalidator' thing that not only would make it easier to print the argument names (i.e., it could put different brackets around required and optional arguments, as well as including defaults, etc.) but it would also let us print a usage of the command if it's misused. I also think it would be a lot better to include the command arguments in the command initialization (rather than the apply function). Let me know your thoughts on it all!

tests/mentat_prompt_session_test.py
```diff
@@ -0,0 +1,55 @@
...
+import pytest
+from prompt_toolkit import PromptSession
+
+from mentat.app import run
+
+# Unfortunately these tests just can't be run on Windows
+pytestmark = pytest.mark.skipif(
+    os.name == "nt",
+    reason="PromptSession throws an error on Github Actions Windows runner",
+)
```
biobootloader, line 14: does it work on Windows though?

mentat/commands.py
```diff
@@ -0,0 +1,104 @@
...
+    def help_message(cls) -> str:
+        return "Displays this message"
+
+
+class CommitCommand(Command, command_name="commit"):
+    default_message = "Automatic commit"
+
+    def apply(self, *args: str) -> None:
+        if len(args) > 1:
+            commit(args[1])
```
biobootloader, line 94: most commit messages will be multiple words separated by a space. On the command line you'd do `git commit -m "multiple words separated by spaces"`, so we should probably also capture everything in quotes

mentat/mentat_prompt_session.py
```diff
@@ -0,0 +1,105 @@
...
+        ):
+            super().append_string(string)
+
+
+class FilteredHistorySuggestions(AutoSuggestFromHistory):
+    def __init__(self):
+        super().__init__()
+
+    def get_suggestion(self, buffer: Buffer, document: Document) -> Suggestion | None:
+        # We want the auto completer to handle commands instead of the suggester
```
biobootloader, line 38: what's the difference between autocompleter and suggester?

mentat/commands.py
biobootloader: lets have a call and maybe you can walk me through this

biobootloader (commented):
Awesome stuff! This is a big step up in functionality we'll be able to have.

I've left some comments, I still need to better understand the prompt_toolkit stuff

tests/mentat_prompt_session_test.py
PCSwingle: Yes, I don't see why it shouldn't work on Windows (although I'll definitely test it myself before merging); the only problem is with Github Actions windows runner (which doesn't like prompttoolkit for whatever reason.

mentat/commands.py
PCSwingle: Hmm, good point; I'm wondering if there's a library that can do argument parsing for us? I'll look into this

mentat/mentat_prompt_session.py
PCSwingle: Autocompleter shows the ghost words to the right of your cursor that you can tab to complete; the autocompleter shows like a dropdown that you can scroll through. If you just type / on this branch you'll see the autocompleter and it will make more sense

PCSwingle: It looks like python has a builtin library shlex.split that will automatically handle this all for us!

mentat/commands.py
PCSwingle: It looks like python has a builtin library shlex.split that will automatically handle this all for us!

tests/mentat_prompt_session_test.py
PCSwingle: Just confirmed; it does work on windows

biobootloader (approved):
🥳

PR #60, opened 2023-08-13
Add AI plugin Flask App

lesaun:
Adds a Flask app that allows for Mentat to be used as a ChatGPT plugin, allowing for ChatGPT Pro users with plugin developer access to avoid paying for GPT-4 API.

To run as a ChatGPT plugin you would add flag `--plugin-api` to Mentat command, and then via ChatGPT go to plugin store > develop your own plugin > enter domain localhost:3333

Let me know if this would be in scope to merge. The prompt and code needs more work, but I've had fun using it with this current state.


biobootloader:
This is very cool but we haven't had the time to merge, and not sure we want to maintain usage this way. It'd be a nice fork project though if people want to use Mentat as a ChatGPT plugin!

PR #61, opened 2023-08-13
Add support for Azure, OpenAI, Llama2, Palm, Cohere, Replicate Models - using litellm

ishaan-jaff:
This PR adds support for models from all the above mentioned providers using https://github.com/BerriAI/litellm/

Here's a sample of how it's used:

```
from litellm import completion, acompletion

## set ENV variables
# ENV variables can be set in .env file, too. Example in .env.example
os.environ["OPENAI_API_KEY"] = "openai key"
os.environ["COHERE_API_KEY"] = "cohere key"

messages = [{ "content": "Hello, how are you?","role": "user"}]

# openai call
response = completion(model="gpt-3.5-turbo", messages=messages)

# llama2 call
model_name = "replicate/llama-2-70b-chat:2c1608e18606fad2812020dc541930f2d0495ce32eee50074220b87300bc16e1"
response = completion(model_name, messages)

# cohere call
response = completion("command-nightly", messages)

# anthropic call
response = completion(model="claude-instant-1", messages=messages)
```

### DIFF
--- a/mentat/llm_api.py
+++ b/mentat/llm_api.py
@@ -5,6 +5,7 @@ from dataclasses import dataclass
 from typing import Generator
 
 import openai
+from litellm import completion
 import tiktoken
 from dotenv import load_dotenv
 from termcolor import cprint
@@ -41,7 +42,7 @@ async def call_llm_api(messages: list[dict[str, str]], model) -> Generator:
         logging.critical("OpenAI call made in non benchmark test environment!")
         sys.exit(1)
 
-    response = await openai.ChatCompletion.acreate(
+    response = await completion(
         model=model,
         messages=messages,
         temperature=0.5,
--- a/requirements.txt
+++ b/requirements.txt
@@ -44,3 +44,4 @@ tqdm==4.65.0
 urllib3==2.0.4
 wcwidth==0.2.6
 yarl==1.9.2
+litellm==0.1.386


ishaan-jaff:
Addresses: https://github.com/biobootloader/mentat/issues/58 

@biobootloader can you please take a look at this when you get the chance?😊

Happy to add more docs/tests if this initial commit looks good

PCSwingle:
Hmm, this looks like a pretty nice library; thanks for finding it and getting it setup! A few things we need to get done before we can merge this: change the completion to acompletion, add in the docs how to put in different api keys, change how we check for maximum context size, and add a way to change which model we use. The codebase is very openai hardcoded right now, so it these might be somewhat bigger changes. I wouldn't mind picking them up myself; up to you if you want to do it. Also, if you have any questions about any of the changes just let me know!

biobootloader:
Thanks @ishaan-jaff ! This would be great. Like @PCSwingle suggested, a couple things we need to add before merging:

- async completion calls (so we can stream results to the user, like we have been doing)
- a way for the user to select the model? I think a command line argument would be best, but open to suggestions
- right now we check that the token count is under the limit for either gpt-4 or gpt-4-32k. We'd have to check for these other models as well. Also, are the tokenization schemes different?
- docs / tests

biobootloader:
Closing this as we now support most of these via #100

krrishdholakia:
Hi @biobootloader
- Async completions: we support those for completion + streaming - https://docs.litellm.ai/docs/completion/stream
- Token counting: we support token counting with model-specific tokenizers (e.g. cohere, anthropic, llama2) where available. We also maintain a list of popular LLMs and their max tokens - https://docs.litellm.ai/docs/token_usage

It wasn't clear to me how #100 enables other models to be supported - my assumption is that it allows proxy usage? 

Your roadmap says you're trying to add new LLMs. If this is something that's still top-of-mind, we'd love to help out. 

cc: @PCSwingle

PR #62, opened 2023-08-13
Refactor: keep git_root private to CodeFileManager

### LINKED ISSUE #57: Refactor: keep `git_root` private to `CodeFileManager`
Right now, `git_root` gets passed around a lot, including being stored in each `CodeChange` object. This is necessary because `CodeFileManager.file_lines` maps full paths to files.

To clean this up, we should refactor to have `CodeFileManager.file_lines` keys just be the relative file paths within the project (from the git root). `CodeFileManager` is the only thing that needs to know the actual path to the git project on the system, since it reads / writes to files.

dheerajck:
Resolves #57

### DIFF
--- a/mentat/code_change.py
+++ b/mentat/code_change.py
@@ -40,7 +40,6 @@ class CodeChange:
         self,
         json_data: dict,
         code_lines: list[str],
-        git_root: str,
         code_file_manager,
     ):
         self.json_data = json_data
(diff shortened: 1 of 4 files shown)


mentat/code_change.py
```diff
@@ -113,13 +111,13 @@ def __init__(
             self.error = "Starting line of change is greater than ending line of change"
 
         if self.action != CodeChangeAction.CreateFile:
-            file_path = os.path.join(self.git_root, self.file)
+            relative_file_path = str(self.file)
```
PCSwingle, line 114: I would prefer this be variable be named rel_path, since that's the norm.

PCSwingle:
Thanks for getting this done, this looks great! Left one minor comment, but I think it should be good to merge after that gets done.

biobootloader (approved):
🚀

PR #66, opened 2023-08-19
Add high-level code map to system prompt

waydegg:
### Summary
This PR adds a high-level code map generated from [ctags](https://github.com/universal-ctags/ctags) into the system prompt which gives the model more context on what other code/classes/methods/etc. it can use that haven't been directly added into the model context by the user. 

This implementation takes inspiration from how [Aider](https://github.com/paul-gauthier/aider) works with large codebases. [Here's a writeup](https://aider.chat/docs/ctags.html) from the author of Aider on how it works.

### Why this is needed
This work is a prerequisite for having Mentat automatically add new source code from files outside the context.

### To-Do
Need to finish these before this PR is ready for review:
- [x] create a 'minified' code map that fits into the model context if the full code map is too large] 
- [ ] ~~prevent mentat from suggesting edits for files which exist in the code map outside the model context]~~
- [x] add a token limit to the code map, filter out less relevant ctags/files as needed
- [x] update test cases 
- [x] handle user not having ctags installed 

### Prior to Merge
- [x] @biobootloader: run benchmarks with and without, to determine if system prompt changes degrade performance on other tasks.

### DIFF
--- a/mentat/code_map.py
+++ b/mentat/code_map.py
@@ -1,13 +1,10 @@
 import json
-import logging
 import os
 import re
 import subprocess
 import tempfile
 from pathlib import Path
 from typing import List
-
-from ipdb import set_trace
 
 from .llm_api import count_tokens
 
@@ -119,7 +116,7 @@ def _get_file_map(file_paths: List[str]):
 
 
 class CodeMap:
-    def __init__(self, git_root: str, token_limit: int | None = None):
+    def __init__(self, git_root: str, token_limit: int | None = 2048):
         self.git_root = git_root
         self.token_limit = token_limit
 
--- a/mentat/code_map.py
+++ b/mentat/code_map.py
@@ -7,6 +7,10 @@ from typing import DefaultDict, Set, Tuple
 
 from ipdb import set_trace
 
+# To-Do
+# 1. prevent mentat editing files which exist in the code map but not in the code message
+# 2. create a mini-code map
+
 
 def get_git_root():
     try:
@@ -109,9 +113,9 @@ def _get_message(ctags: DefaultDict[str, Set[Tuple[str, ...]]]):
         file_maps.append(output)
 
     message = f"""
-        Below is a map of all files tracked by git in this project. Let the user know 
-        if you think they should add the complete source code of any files to this 
-        chat.
+        Below is a read-only code map of all files tracked by git in this project. 
+        If you need to edit any of the files in this code map that aren't in the current context,
+        don't try to edit them and instead give the user the filepaths of what they should add to this context.
     """
     message = message.strip()
     message = re.sub(r"[\n\s]+", " ", message)
--- a/mentat/code_map.py
+++ b/mentat/code_map.py
@@ -32,8 +32,8 @@ def _get_code_map(root: str, file_path: str, exclude_signatures: bool = False):
         try:
             tag = json.loads(output_line)
         except json.decoder.JSONDecodeError as err:
-            cprint(f"Error parsing ctags output: {err}")
-            cprint(repr(output_line))
+            cprint(f"Error parsing ctags output: {err}", color="yellow")
+            cprint(f"{repr(output_line)}\n", color="yellow")
             continue
 
         scope = tag.get("scope")
--- a/mentat/code_map.py
+++ b/mentat/code_map.py
@@ -1,5 +1,4 @@
 import json
-import os
 import re
 import subprocess
 import tempfile
@@ -9,7 +8,7 @@ from typing import List
 from .llm_api import count_tokens
 
 
-def get_git_files(git_root: str, absolute_paths: bool = False) -> List[str]:
+def _get_git_files(git_root: str, absolute_paths: bool = False) -> List[str]:
     git_files_ = (
         subprocess.check_output(["git", "ls-files"], cwd=git_root)
         .decode("utf-8")
@@ -65,6 +64,9 @@ def _get_code_map(root: str, file_path: str, exclude_signatures: bool = False):
 
         ctags.add(tuple(res))
 
+    if len(ctags) == 0:
+        return f"{file_path}\n"
+
     # Build LLM-readable string representation of ctag objects
     sorted_tags = sorted(ctags)
     output = ""
@@ -116,7 +118,7 @@ def _get_file_map(file_paths: List[str]):
 
 
 class CodeMap:
-    def __init__(self, git_root: str, token_limit: int | None = 2048):
+    def __init__(self, git_root: str, token_limit: int | None = None):
         self.git_root = git_root
         self.token_limit = token_limit
 
@@ -144,10 +146,10 @@ class CodeMap:
                 return
 
             with tempfile.TemporaryDirectory() as tempdir:
-                hello_py = os.path.join(tempdir, "hello.py")
+                hello_py = Path(tempdir).joinpath("hello.py")
                 with open(hello_py, "w", encoding="utf-8") as f:
                     f.write("def hello():\n    print('Hello, world!')\n")
-                _get_code_map(tempdir, hello_py)
+                _get_code_map(tempdir, str(hello_py))
         except FileNotFoundError:
             self.ctags_disabled_reason = f"ctags executable not found"
             return
@@ -168,7 +170,10 @@ class CodeMap:
             )
             code_map_token_count = count_tokens(code_map)
 
-            if self.token_limit is not None and code_map_token_count > self.token_limit:
+            if (
+                self.token_limit is not None
+                and code_maps_token_count > self.token_limit
+            ):
                 if exclude_signatures is True:
                     return
                 return self._get_code_map_message(
@@ -214,7 +219,7 @@ class CodeMap:
         return message
 
     def get_message(self):
-        git_file_paths = get_git_files(self.git_root)
+        git_file_paths = _get_git_files(self.git_root)
 
         if not self.ctags_disabled:
             code_map_message = self._get_code_map_message(self.git_root, git_file_paths)
--- a/mentat/code_map.py
+++ b/mentat/code_map.py
@@ -3,13 +3,11 @@ import re
 import subprocess
 from collections import defaultdict
 from pathlib import Path
-from typing import DefaultDict, Set, Tuple
+from typing import DefaultDict, List, Set, Tuple
 
 from ipdb import set_trace
 
-# To-Do
-# 1. prevent mentat editing files which exist in the code map but not in the code message
-# 2. create a mini-code map
+from .llm_api import count_tokens
 
 
 def get_git_root():
@@ -25,23 +23,24 @@ def get_git_root():
     return git_root
 
 
-def get_git_files():
-    git_files = (
-        subprocess.check_output(["git", "ls-files"]).decode("utf-8").splitlines()
+def get_git_files(git_root: str, absolute_paths: bool = False) -> List[str]:
+    git_files_ = (
+        subprocess.check_output(["git", "ls-files"], cwd=git_root)
+        .decode("utf-8")
+        .splitlines()
     )
+    if absolute_paths:
+        git_files = [str(Path(git_root).joinpath(f)) for f in git_files_]
+    else:
+        git_files = git_files_
 
     return git_files
 
 
-def _build_ctags():
+def _build_ctags(root: str, file_paths: List[str]):
     ctags: DefaultDict[str, Set[Tuple[str, ...]]] = defaultdict(set)
 
-    git_root = get_git_root()
-    if git_root is None:
-        return
-    git_file_paths = get_git_files()
-
-    for file_path in git_file_paths:
+    for file_path in file_paths:
         ctags_cmd = [
             "ctags",
             "--fields=+S",
@@ -49,7 +48,7 @@ def _build_ctags():
             "--input-encoding=utf-8",
             "--output-format=json",
             "--output-encoding=utf-8",
-            str(Path(git_root).joinpath(file_path)),
+            str(Path(root).joinpath(file_path)),
         ]
         output = subprocess.check_output(ctags_cmd, stderr=subprocess.PIPE).decode(
             "utf-8"
@@ -83,10 +82,10 @@ def _build_ctags():
     return ctags
 
 
-def _get_message(ctags: DefaultDict[str, Set[Tuple[str, ...]]]):
+def _get_code_map_message(ctags: DefaultDict[str, Set[Tuple[str, ...]]]):
     file_maps = []
-    for ctags in ctags.values():
-        sorted_tags = sorted(ctags)
+    for ctags_ in ctags.values():
+        sorted_tags = sorted(ctags_)
 
         output = ""
         last = [None] * len(sorted_tags[0])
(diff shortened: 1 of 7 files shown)


biobootloader:
Super pumped for this!

waydegg:
For sessions where there aren't a bunch of files/code, the updated system prompt is pretty good at not suggesting edits to files it doesn't have in it's context (but does have in it's Code Map). It does start to break down as LOC and number of files grows however, and at a certain point it just keeps suggesting edits no matter what I try with the system prompt.

I think a solution for this should be handled in a separate PR which does something like:
- Get a list of files to add/edit/delete at the beginning of the response
- As the response/file list is streamed in, check if the files have been added into the context 
- If they haven't been added, stop the response

waydegg:
Here are some partial snippets of different sized code maps:

## Full Code Map 

```
mentat/__main__.py

mentat/app.py
        function
                expand_paths (paths: Iterable[str])
                get_user_feedback_on_changes ( config: ConfigManager, conv: Conversation, user_input_manager: UserInputManager, code_file_manager: CodeFileManager, code_changes: Iterable[CodeChange], )
                loop ( paths: Iterable[str], exclude_paths: Optional[Iterable[str]], cost_tracker: CostTracker, )
                run (paths: Iterable[str], exclude_paths: Optional[Iterable[str]] = None)
                run_cli ()
                user_filter_changes ( user_input_manager: UserInputManager, code_changes: Iterable[CodeChange] )

mentat/change_conflict_resolution.py
        function
                resolve_insertion_conflicts ( changes: list[CodeChange], user_input_manager: UserInputManager, code_file_manager )
                resolve_non_insertion_conflicts ( changes: list[CodeChange], user_input_manager: UserInputManager )

mentat/code_change.py
        CodeChange
                member
                        __init__ ( self, json_data: dict, code_lines: list[str], code_file_manager, )
                        __lt__ (self, other)
                        apply (self, cur_file_lines: list[str])
        CodeChangeAction
                member
                        has_additions (self)
                        has_removals (self)
                        has_surrounding_lines (self)
                variable
                        CreateFile
                        Delete
                        DeleteFile
                        Insert
                        Replace
        class
                CodeChange
                CodeChangeAction
```

## Code Map without signatures

```
mentat/__main__.py

mentat/app.py
        function
                expand_paths
                get_user_feedback_on_changes
                loop
                run
                run_cli
                user_filter_changes

mentat/change_conflict_resolution.py
        function
                resolve_insertion_conflicts
                resolve_non_insertion_conflicts

mentat/code_change.py
        class
                CodeChange
                CodeChangeAction
        member
                __init__
                __lt__
                apply
                has_additions
                has_removals
                has_surrounding_lines
        variable
                CreateFile
                Delete
                DeleteFile
                Insert
                Replace
```

## Code map with file structure only

```
.gitignore
LICENSE
README.md
docs
        configuration.md
mentat
        __init__.py
        __main__.py
        app.py
        change_conflict_resolution.py
        code_change.py
        code_change_display.py
        code_file_manager.py
        code_map.py
        config_manager.py
        conversation.py
        default_config.json
        errors.py
        git_handler.py
        llm_api.py
        logging_config.py
        parsing.py
        prompts.py
        streaming_printer.py
        user_input_manager.py
pyproject.toml
requirements.txt
setup.py
```

biobootloader:
Fantastic! I'll review this tomorrow. Thank you for the detailed code map examples!

mentat/conversation.py
```diff
@@ -52,7 +55,25 @@ def get_model_response(self, config: ConfigManager) -> (str, list[CodeChange]):
         messages = self.messages.copy()
 
         code_message = self.code_file_manager.get_code_message()
-        messages.append({"role": "system", "content": code_message})
+        system_message = code_message
+
+        if self.code_map:
```
PCSwingle, line 60: This looks like an ideal situation for a walrus (:=) operator (instead of 2 nested if statements)

mentat/conversation.py
```diff
@@ -52,7 +55,25 @@ def get_model_response(self, config: ConfigManager) -> (str, list[CodeChange]):
...
+                prompt_token_count = 0
+                prompt_token_count += count_tokens(code_message)
+                for message in messages:
+                    prompt_token_count += count_tokens(message["content"])
+                token_buffer = 1000
+                token_limit = 32768 if self.allow_32k else 8192
+                token_count = (
+                    prompt_token_count + code_map_message_token_count + token_buffer
+                )
+                if token_count < token_limit:
```
PCSwingle, line 73: I think the user should probably know if ctags isn't added because the context isn't big enough; maybe just some yellow text?

mentat/conversation.py
```diff
@@ -52,7 +55,25 @@ def get_model_response(self, config: ConfigManager) -> (str, list[CodeChange]):
...
+                    system_message = "\n".join([code_message, code_map_message])
```
PCSwingle, line 74: Since system_message is set to code_message up there, I think it would be a lot cleaner to do += f"\n{code_map_message}"

mentat/prompts.py
```diff
@@ -1,129 +1,202 @@
...
+    @@start
+    {
+        "file": "core/hello_world.py",
+        "action": "delete-file",
+    }
+    @@end
 """
-)
+
+system_prompt = "\n\n" + "\n".join([system_prompt_prefix, system_prompt_example])
```
PCSwingle, line 202: The example prompt should also be stripped, right? It can also use dedent

mentat/code_map.py
```diff
@@ -0,0 +1,226 @@
...
+import tempfile
+from pathlib import Path
+from typing import List
+
+from .llm_api import count_tokens
+
+
+def _get_git_files(git_root: str, absolute_paths: bool = False) -> List[str]:
+    git_files_ = (
+        subprocess.check_output(["git", "ls-files"], cwd=git_root)
```
PCSwingle, line 12: If you pass in text=True as an argument, you don't need .decode

mentat/code_map.py
```diff
@@ -0,0 +1,226 @@
...
+        .decode("utf-8")
+        .splitlines()
+    )
+    if absolute_paths:
+        git_files = [str(Path(git_root).joinpath(f)) for f in git_files_]
```
PCSwingle, line 17: We don't need 2 separate git_files variables; you can just overwrite git_files in the if statement and drop the else statement

mentat/code_map.py
```diff
@@ -0,0 +1,226 @@
...
+        "--input-encoding=utf-8",
+        "--output-format=json",
+        "--output-encoding=utf-8",
+    ]
+    if exclude_signatures:
+        ctags_cmd_args.append("--fields=-s")
+    else:
+        ctags_cmd_args.append("--fields=+S")
+    ctags_cmd = ["ctags", *ctags_cmd_args, str(Path(root).joinpath(file_path))]
+    output = subprocess.check_output(ctags_cmd, stderr=subprocess.PIPE).decode("utf-8")
```
PCSwingle, line 37: same text=True thing here

mentat/conversation.py
```diff
@@ -52,7 +55,25 @@ def get_model_response(self, config: ConfigManager) -> (str, list[CodeChange]):
...
+        if self.code_map:
+            code_map_message = self.code_map.get_message()
+            if code_map_message is not None:
+                code_map_message_token_count = count_tokens(code_map_message)
+                prompt_token_count = 0
+                prompt_token_count += count_tokens(code_message)
+                for message in messages:
+                    prompt_token_count += count_tokens(message["content"])
+                token_buffer = 1000
+                token_limit = 32768 if self.allow_32k else 8192
```
biobootloader, line 69: we also calculate this token limit in the `__init__`, lets set it as an attribute there and reuse here

mentat/conversation.py
```diff
@@ -52,7 +55,25 @@ def get_model_response(self, config: ConfigManager) -> (str, list[CodeChange]):
...
+        system_message = code_message
+
+        if self.code_map:
+            code_map_message = self.code_map.get_message()
+            if code_map_message is not None:
+                code_map_message_token_count = count_tokens(code_map_message)
+                prompt_token_count = 0
+                prompt_token_count += count_tokens(code_message)
+                for message in messages:
+                    prompt_token_count += count_tokens(message["content"])
```
biobootloader, line 67: perhaps we should make `self.messages` more than just a list of messages, but instead also store token counts for each message and/or a total, so we don't recalculate again and again?

don't have to make that change as part of this PR, we can make a separate issue for that

mentat/conversation.py
biobootloader: I agree, let's display something so the user knows what kind of map was added, and if it was added

mentat/conversation.py
```diff
@@ -52,7 +55,25 @@ def get_model_response(self, config: ConfigManager) -> (str, list[CodeChange]):
         messages = self.messages.copy()
 
         code_message = self.code_file_manager.get_code_message()
-        messages.append({"role": "system", "content": code_message})
+        system_message = code_message
+
+        if self.code_map:
+            code_map_message = self.code_map.get_message()
```
biobootloader, line 61: We aren't using the `token_limit` arg here, but if we did we could get the nice fallback behavior you've implemented (trying to return smaller maps if it doesn't fit in the limit). So shouldn't we count tokens first and tell it how much space it has? (right now if I understand correctly it's always using the limit of 2048 tokens set when `CodeMap` is initialized in `app.py`)

biobootloader (commented):
Great new feature! I've left a few comments, here are some more general points:

Since there's a big change to the system prompt, we should run benchmarks with and without to make sure we don't lose performance. I can do that once we are closer to merging. I'll add this task to the PR description as well.

If it suggests edits to files it can't fully see, it'd be good to suggest the user adds them to the context. We'll soon have "commands" (https://github.com/biobootloader/mentat/pull/59) and a /add command to add a specific file. In this PR we can just warn the user that it tried to edit a file it couldn't see the code for and have them rerun with it added.

Let's make this feature optional as well, since big maps would add to api costs. We can have it be default on and have a command line flag to turn off? (For example if a user is editing an isolated file or two in a big codebase, they may want to turn it off).

biobootloader:
> For sessions where there aren't a bunch of files/code, the updated system prompt is pretty good at not suggesting edits to files it doesn't have in it's context (but does have in it's Code Map). It does start to break down as LOC and number of files grows however, and at a certain point it just keeps suggesting edits no matter what I try with the system prompt.
> 
> I think a solution for this should be handled in a separate PR which does something like:
> 
> * Get a list of files to add/edit/delete at the beginning of the response
> * As the response/file list is streamed in, check if the files have been added into the context
> * If they haven't been added, stop the response

Agreed, let's do this in a follow-up PR

biobootloader:
For testing, lets add a benchmark (https://github.com/biobootloader/mentat/blob/main/tests/benchmark_test.py) that asks it to make change that it could only make successfully with the map. Set up some simple scenario where you only show it one file, and ask it to make a change to that file that requires a call to a function in another file, that it could see in the map.

Since benchmarks don't run on Github actions it'd also be good to add some sanity check tests as well. Especially with things that we wouldn't notice if we accidentally broke, like which map is used with different token limits, etc.

mentat/code_map.py
waydegg: Ooh nice, didn't know about that!

waydegg:
Thanks for the reviews! I agree w/ all of the suggestions/comments, will get to work on addressing everything today.

mentat/conversation.py
waydegg: Good point. I'm going to add a `cprint` at the start of the session too letting the user know if the `CodeMap` is enabled at all, and if it isn't why.

mentat/code_map.py
```diff
@@ -0,0 +1,226 @@
+import json
+import subprocess
+import tempfile
+from pathlib import Path
+from typing import List
+
+from .llm_api import count_tokens
+
+
+def _get_git_files(git_root: str, absolute_paths: bool = False) -> List[str]:
```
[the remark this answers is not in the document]
waydegg: I think including untracked/new files is a good choice, makes for better DX. I'll use the function that's already in `git_handler.py`.

I had `absolute_paths` as an option for developent, but it's not needed anymore as we can just append the root path later on if it's needed.

mentat/conversation.py
waydegg: >We aren't using the token_limit arg here, but if we did we could get the nice fallback behavior you've implemented (trying to return smaller maps if it doesn't fit in the limit)

We're still getting the fallback behavior with or without setting `token_limit` for `self.code_map.get_message()`. Setting `token_limit` in `.get_message()` is just a manual override to whatever `token_limit` the `CodeMap` was instantiated with. We don't *need* it atm, so I can remove it if you'd like. It's useful for development/testing tho.

> So shouldn't we count tokens first and tell it how much space it has?

Could you elaborate on this? I think I'm misinterpreting the question.

waydegg: `token_limit` is just looking at the token count for the generated code map btw

mentat/conversation.py
waydegg: I'm wondering how much of a performance increase this would be? I don't think it's much work to set this up though so it's worth it to see if it helps a good amount.

mentat/prompts.py
```diff
@@ -1,129 +1,202 @@
...
+    other than code changes, such as design ideas, don't return any edit
+    description blocks.
+
+    2. If you're missing code files in "Code Files" that you need to edit: 
+
+    If you need to make changes to files that you don't have the full source code of 
+    (only the high-level code map), tell the user which files you need the source code
+    for so you can make edits correctly.
+"""
+system_prompt_prefix = system_prompt_prefix.strip()
```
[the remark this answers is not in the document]
waydegg: Awesome, never knew about `dedent` before

mentat/conversation.py
biobootloader: Yeah it may not really matter, but if done cleanly would be nice. Doesn't have to be this PR though

mentat/conversation.py
waydegg: Ah gotcha, thanks for clarifying and good call-out :)

waydegg:
@biobootloader I added a benchmark test that forces the model to import from a module in the code map that's outside the code files context. 

Also, I decided to revert the system prompt closer to what it was originally for a couple reasons:
- Since we aren't trying to prevent edits to files not added in CodeFiles for this PR, we shouldn't introduce that logic into the system prompt just yet
- There was some variability in the benchmark tests passing/failing with the new system prompt after I ran the benchmarks several times back-to-back. This is to be expected since we have temperature set to a non-zero value and while the benchmarks on the main branch pass/fail with some variance, there's less of a chance of introducing any regressions by keeping the closer-to-original system prompt

biobootloader:
Thanks for running the benchmarks, and I yeah I agree changing the system prompt is hard. More and better benchmarking would help us change it with more confidence. I like the call to keep it more similar to original.

mentat/conversation.py
```diff
@@ -52,7 +59,40 @@ def get_model_response(self, config: ConfigManager) -> (str, list[CodeChange]):
...
+            )
+            if code_map_message:
+                system_message += f"\n{code_map_message}"
+            else:
+                cprint_message = [
+                    "\nExcluding CodeMap from system message.",
+                    "Reason: not enough tokens available in model context.",
+                ]
+                cprint_message = "\n".join(cprint_message)
+                cprint(cprint_message, color="yellow")
```
biobootloader, line 93: Let's have CodeMap.get_message also return which level map it sent back (full detail with signatures, without signatures, file map only), and inform the user which level is being used.

I'm not sure it'll be something we'll always show the user, but I think it's important for now, because users might notice different quality edits and not know it's due to the map changing (and that removing extra files from context could have helped by giving more space for the map)

biobootloader:
Looks great! This will be ready to merge after addressing my one comment about letting the user know which kind of map (level of detail) is being used.

I've added two follow up issues to add an additional benchmark and test after this is merged: #76 , #77

biobootloader (approved):
🎉  awesome!

PR #67, opened 2023-08-19
Fix git root parsing when using msys2 git on Windows

vickunwu:
Currently, mentat won't work with a msys2-installed Git on Windows because when executing `git rev-parse --show-toplevel`, it will produce a non-compatible path (e.g. /c/Users/xxx/mentat).
According to [Msys2](https://www.msys2.org/docs/git/), they suggest to use `git rev-parse --show-prefix`, which outputs a relative path from the root to the current working directory and it is both a valid Unix and Windows path.
So I modified the related code, and it works fine for my setup.

### DIFF
--- a/mentat/git_handler.py
+++ b/mentat/git_handler.py
@@ -61,6 +61,8 @@ def _get_git_root_for_path(path) -> str:
             .decode("utf-8")
             .strip()
         )
+        # --show-toplevel doesn't work in some windows environment with posix paths,
+        # like msys2, so we have to use --show-prefix instead
         git_root = os.path.abspath(
             os.path.join(dir_path, "../" * len(Path(relative_path).parts))
         )
--- a/mentat/git_handler.py
+++ b/mentat/git_handler.py
@@ -48,12 +48,12 @@ def _get_git_root_for_path(path) -> str:
     else:
         dir_path = os.path.dirname(path)
     try:
-        git_root = (
+        relative_path = (
             subprocess.check_output(
                 [
                     "git",
                     "rev-parse",
-                    "--show-toplevel",
+                    "--show-prefix",
                 ],
                 cwd=os.path.realpath(dir_path),
                 stderr=subprocess.DEVNULL,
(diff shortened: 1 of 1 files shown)


biobootloader:
Thanks!

It looks like this has broken tests (on linux, mac and windows).

Would calling `os.path.normpath` on the path returned by msys2-installed git work instead?

vickunwu:
I am sorry for the primitive mistake I made.
I am not familiar with Python.
It turns out `os.path.split` always returns a tuple of length 2.

vickunwu:
> Would calling `os.path.normpath` on the path returned by msys2-installed git work instead?

The problem is not on the code side. When executing `git rev-parse --show-toplevel`, the output will always append the drive symbol to its front, which will cause a `NotADirectoryError`. Below is the example:

Output of `git rev-parse --show-toplevel`: /c/Users/xxx/mentat
Using `os.path.realpath` to convert: C:\c\Users\xxx\mentat
Using `os.path.normpath` to convert: \c\Users\xxx\mentat

PCSwingle:
Hmm, this feels hacky but it seems like it might be the best way to do this. I'm going to add a comment as to why we do this and then merge it.

mentat/git_handler.py
```diff
@@ -48,19 +48,22 @@ def _get_git_root_for_path(path) -> str:
...
-                    "--show-toplevel",
+                    "--show-prefix",
                 ],
                 cwd=os.path.realpath(dir_path),
                 stderr=subprocess.DEVNULL,
             )
             .decode("utf-8")
             .strip()
         )
+        git_root = os.path.abspath(
```
PCSwingle, line 64:
```suggestion
        # --show-toplevel doesn't work in some windows environment with posix paths,
        # like msys2, so we have to use --show-prefix instead
        git_root = os.path.abspath(
```

biobootloader:
thanks for the change @vickunwu !  @PCSwingle think we should add a test to ensure this still works even if Mentat is called from a subdirectory of the git project? maybe we already have one. Feel free to merge if you think it's fine without!

biobootloader (approved):

PR #69, opened 2023-08-20
Partial files allowed

jakethekoenig:
The user can specify a partial file with : syntax e.g. test.txt:1-2,5 will put only lines 1,2 and 5 in scope. It is checked if that file with a colon already exists. For this a File object is created in the code_file_manager which has the path and segment information bundled.

The expand function is moved from app.py to code_file_manager.py

Test with:
```
mentat README.md:6-12,16
> Replace all instance of the word mentat with test.
```

### DIFF
--- a/mentat/app.py
+++ b/mentat/app.py
@@ -1,4 +1,5 @@
 import argparse
+import glob
 import logging
 from typing import Iterable, Optional
 
@@ -6,6 +7,7 @@ from termcolor import cprint
 
 from .code_change import CodeChange
 from .code_change_display import print_change
+from .code_file import parse_intervals
 from .code_file_manager import CodeFileManager
 from .config_manager import ConfigManager, mentat_dir_path
 from .conversation import Conversation
@@ -36,7 +38,34 @@ def run_cli():
     args = parser.parse_args()
     paths = args.paths
     exclude_paths = args.exclude
-    run(paths, exclude_paths)
+    # Expanding paths as soon as possible because some shells such as zsh automatically
+    # expand globs and we want to avoid differences in functionality between shells
+    run(expand_paths(paths), expand_paths(exclude_paths))
+
+
+def expand_paths(paths: Iterable[str]) -> Iterable[str]:
+    globbed_paths = set()
+    invalid_paths = []
+    for path in paths:
+        new_paths = glob.glob(pathname=path, recursive=True)
+        if new_paths:
+            globbed_paths.update(new_paths)
+        else:
+            split = path.rsplit(":", 1)
+            p = split[0]
+            intervals = parse_intervals(split[1])
+            if Path(p).exists() and intervals is not None:
+                globbed_paths.add(path)
+            else:
+                invalid_paths.append(path)
+    if invalid_paths:
+        cprint(
+            "The following paths do not exist:",
+            "light_yellow",
+        )
+        print("\n".join(invalid_paths))
+        exit()
+    return globbed_paths
 
 
 def run(paths: Iterable[str], exclude_paths: Optional[Iterable[str]] = None):
--- a/mentat/app.py
+++ b/mentat/app.py
@@ -49,7 +49,7 @@ def run_cli():
     no_code_map = args.no_code_map
     # Expanding paths as soon as possible because some shells such as zsh automatically
     # expand globs and we want to avoid differences in functionality between shells
-    run(expand_paths(paths), expand_paths(exclude_paths))
+    run(expand_paths(paths), expand_paths(exclude_paths), no_code_map)
 
 
 def expand_paths(paths: Iterable[str]) -> Iterable[str]:
--- a/mentat/app.py
+++ b/mentat/app.py
@@ -1,5 +1,4 @@
 import argparse
-import glob
 import logging
 from typing import Iterable, Optional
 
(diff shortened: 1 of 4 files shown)


jakethekoenig:
[This is for issue 56.](https://github.com/biobootloader/mentat/issues/56)

Notes on some of the choices I made in this PR:

It's probably a bad idea to name my File class File. I originally wrote it as FileFragment but thought that was too cumbersome. Maybe File partial.

I thought it wasn't necessary to change the prompts at all. The model seems to not have an issue with skipped line numbers.

I didn't add a manual check that the model doesn't add lines outside of specified ranges. I think if someone isn't prompt engineering for this specifically it shouldn't end up happening. And if the user requests `add a print after line 10` after only giving the context 15-20 I think the model should attempt to do it. In my testing I found the model gets confused if you only give it continuous lines from the start and ask for a change after the last line you gave it. But if you give it `README.md:1-10,30-40` and ask it to add or delete something at line 20 it is able to do it.

I added one test but can add more if you like the direction.

I guess you weren't concerned with user interaction in the first step but I though allowing users to explicitly add ranges with :10-20 for example was nice.

I moved `expand` out of `app.py` because it seems best to have all the file wrangling in `code_file_manager.py`.

Sorry if this isn't what you had in mind at all. I maybe should have asked some questions before jumping in.

mentat/app.py
```diff
@@ -37,26 +36,7 @@ def run_cli():
     args = parser.parse_args()
     paths = args.paths
     exclude_paths = args.exclude
-    run(expand_paths(paths), expand_paths(exclude_paths))
-
-
-def expand_paths(paths: Iterable[str]) -> Iterable[str]:
```
PCSwingle, line 43: I want to keep the 'expand_paths' in app.py; the reason why is because some terminals (like zsh) automatically expand globbed paths. This means the later on we call this function, the more likely it is that we get some difference in functionality between terminals. It would definitely be good to add a comment mentioning this to this function though, since this isn't clear at all!

mentat/code_file_manager.py
```diff
@@ -93,7 +94,86 @@ def _abs_file_paths_from_list(paths: Iterable[str], check_for_text: bool = True)
...
+            if len(interval) == 1:
+                intervals += [Interval(int(interval[0]), int(interval[0]))]
+            else:
+                intervals += [Interval(int(interval[0]), int(interval[1]))]
+        return intervals
+    except (ValueError, IndexError):
+        return None
+
+
+class File:
```
PCSwingle, line 136: Hmm, I definitely agree that naming it File is a bit weird; maybe naming it CodeFile would be better? Not sure how much I like that either, but I think it's better.

PCSwingle: I also think we should refactor File and Interval into another file (codefile.py?)

PCSwingle:
Just took a look at your PR and it looks great! Just a few comments that need to be resolved first. One big thing though is the colon and comma syntax; filenames can have colons and commas in them, so I don't think we want to have it be a colon as a separator. I'm not entirely sure what would be best though; I personally don't think we need a way for the user to manually add specific lines. This could just be an internal thing that we'll eventually use when we have auto-context. What do you think @biobootloader ?

mentat/app.py
jakethekoenig: Makes sense. Also thinking about it more one would only want this functionality in the terminal interface anyway. As an editor plugin it would probably be better to build context from opened files.

jakethekoenig:
> Just took a look at your PR and it looks great! Just a few comments that need to be resolved first. One big thing though is the colon and comma syntax; filenames can have colons and commas in them, so I don't think we want to have it be a colon as a separator. I'm not entirely sure what would be best though; I personally don't think we need a way for the user to manually add specific lines. This could just be an internal thing that we'll eventually use when we have auto-context. What do you think @biobootloader ?

I think how it's done now is safe because I check if the file exists before attempting to split on a colon and parse interval information and I always take the last colon and my interval syntax doesn't permit colons. But it's maybe a little strange and a generally bad idea.

Thinking about it more giving the user an affordance like this is possibly a step in the wrong direction and in the long term we should rely on the LLM to request its own context. Perhaps one should ask GPT-3 "the user wants to do X" which lines in the following file do you think will be important to pay attention to in order to make that change.

PCSwingle:
Everything looks great! I'll leave it up to @biobootloader on whether or not we should keep the current colon splitting; my opinion is that the user doesn't need to have access to this functionality, and eventually once we get autocontext the llm will just give us the intervals via json, which would avoid the need for splitting on colons.

mentat/code_file.py
```diff
@@ -0,0 +1,48 @@
...
+                intervals += [Interval(int(interval[0]), int(interval[0]))]
+            else:
+                intervals += [Interval(int(interval[0]), int(interval[1]))]
+        return intervals
+    except (ValueError, IndexError):
+        return None
+
+
+class CodeFile:
+    def __init__(self, path: str | Path):
```
biobootloader, line 33: let's add a docstring here so it's people reading the codebase to understand what this class does

biobootloader (approved):
Nice! Thanks for this!

It is a weird affordance to give users but I think it's fine. It'll be useful for users with single large files that don't fit in context, but the main benefit here is we'll be ready to automatically add partial files automatically.

PR #70, opened 2023-08-22
VSCode Extension (DRAFT)

granawkins:
Implement Mentat as a Visual Studio Code extension. Basic functionality:
- Add a "Mentat" tab in VSCode which opens a new vscode webview
- Page 1: Select files and click a start button
- Page 2: A standard chat interface for the cli workflow, with buttons for interrupt and restart (back to Page 1)

See `/vscode/README.md` for setup instructions.


biobootloader:
@waydegg want to help review this? It's waiting on me to finish #74 , but it'd be good to have you eyes on the VS Code extension implementation

biobootloader:
I wonder if the LSP portion (`vscode/bundled/tool/lsp_server.py`) could be general enough to support other editor extensions at some point, like for NeoVim?

waydegg:
@granawkins do you have any references to VSCode plugins made with Svelte that are somewhat popular/high-quality that you like? Asking since I've never used Svelte before and wanted to see how other plugins do it.

waydegg:
Need to find a way to make the UI work well with light themes:

image

waydegg:
I'm a big fan of what [Cursor](https://www.cursor.so/) is doing for their chat UI/UX. Any thoughts on this?

image

They have some good feedback in their Discord on what people have liked/disliked too.

waydegg:
Is this just blocked on #74? Happy to review but I also don't wanna add a bunch of comments per-maturely if you're still making big additions/edits :) Pretty dope to see this working in VSCode btw!

biobootloader:
> Is this just blocked on https://github.com/biobootloader/mentat/pull/74? Happy to review but I also don't wanna add a bunch of comments per-maturely if you're still making big additions/edits :)

@waydegg yes #74 needs to be merged first. The big open question is the best way to run both Mentat and the Python side of the LSP defined here - should we set up an async loop that runs both? Run on different processes?

> I'm a big fan of what [Cursor](https://www.cursor.so/) is doing for their chat UI/UX. Any thoughts on this?

Looks nice! My hope for this PR is to get a working Mentat experience in VS Code that we can then iterate on. So higher level comments on the best way to set this up are most useful. We should also track interface feature ideas somewhere, feel free to create issues for individual ideas or maybe we can start a list somewhere.

granawkins:
> I wonder if the LSP portion (`vscode/bundled/tool/lsp_server.py`) could be general enough to support other editor extensions at some point, like for NeoVim?

Absolutely, that's what I was thinking too. That's how Rift does it. Since this is from a template I figure we leave it like this for now, and if/when we add NeoVim we can separate it out.

waydegg:
>  The big open question is the best way to run both Mentat and the Python side of the LSP defined here - should we set up an async loop that runs both? Run on different processes?

I think we should def be building towards running everything async (electron already eats up enough processing as is haha). [pygls](https://github.com/openlawlibrary/pygls), the library the VSCode extension template uses that interfaces w/ language servers, has first-class async support. Most of Mentat latency is network io also.

Would love you hear what you guys are thinking.

waydegg:
>I'm not married to Svelte though, it's my first time using it as well. It's really similar to Vue, which I've used a lot, but the bundle is lighter and it builds really fast. Open to suggestions.

I'm not married to any particular framework either (I've been wanting an excuse to learn Svelte anyways haha). I guess my only concern would be ease-of-contributing since most ppl already know React/Vue and there's obv a huge library ecosystem for both frameworks.

waydegg:
At a high level I think this looks good! I'll create a github discussion/issue for feature suggestions we can all collaborate on.

vscode/src/svelte/components/file/File.svelte
```diff
@@ -0,0 +1,102 @@
...
+    <Checkbox 
+      status={status} 
+      renderMixedChildren={renderMixedChildren}
+      handleClick={handleClickCheckbox} 
+    />
+    <span class="file-name">{file.file.name}</span>
+  </div>
+  
+  <div class="children-container {isOpen ? 'open' : ''}">
+    {#if file.children}
```
jakethekoenig, line 59: I notice a ~1 second latency when toggling expand/contract for large directories which can be fixed by checking `isOpen` here.

vscode/src/svelte/components/file/FileTreeElement.ts
```diff
@@ -0,0 +1,75 @@
...
+export class FileTreeElement {
+    public children: FileTreeElement[] | undefined;
+    // A svelte-subscribable store that stores broadcasts changes in status
+    public statusStore = writable(FileInclusionStatus.notIncluded);
+    public status: FileInclusionStatus | undefined; // A locally-accessible cache of status
+
+    constructor(
+        public file: WorkspaceGraphElement, // File metadata
+    ) {
+        if (this.file.children) {
```
jakethekoenig, line 13: For me there's an ~11s time to load after clicking the refresh button on this project (with python and node dependencies installed). Moving this to handleClick and loading files lazily cut it to ~3s.

jakethekoenig:
Interesting: rift already included [mentat](https://github.com/morph-labs/rift/blob/main/rift-engine/rift/agents/mentat_agent.py) as an agent.
Also @waydegg I don't see a cursor discord. Do you mean [their forum](https://forum.cursor.sh/)?

biobootloader:
> Also are we sure we want this as a sub-directory in the mentat repo? I guess if the VSCode and NeoVim plugins end up sharing code it'll be simplest. But it will interfere with the standard way of installing vim plugins: cloning a repo into a plugin directory.

@granawkins and I were talking about this - the benefit of a shared repo is it's easier to coordinate versions and changes between "core" Mentat and the VS Code extension. We do want any code that would be shared by VS Code and NeoVim plugins here (and those bits would probably be useful as well for the community if they make extensions for other editors). @jakethekoenig is the way you see it interfering with vim plugin install just that the cloned repo would be bigger?

Overall my vote is we keep it here for now and see how it goes, but that's not a strong opinion 

> Interesting: rift already included mentat as an agent.

Just announced today! They way they've included it involves inserting callbacks into the code / patching some input / output functions. When we finish getting Mentat setup for multiple interfaces they'll have an easier time connecting to it. 

> Also @waydegg I don't see a cursor discord. Do you mean [their forum](https://forum.cursor.sh/)?

They have a discord but looks like they disabled invites for it and are switching to the forum

vscode/src/svelte/components/file/FileTreeElement.ts
granawkins: Aha - great idea. I've now done something a bit different but same principle.

jakethekoenig: Your approach is really clean. Nice!

jakethekoenig:
> > Also are we sure we want this as a sub-directory in the mentat repo? I guess if the VSCode and NeoVim plugins end up sharing code it'll be simplest. But it will interfere with the standard way of installing vim plugins: cloning a repo into a plugin directory.
> 
> @granawkins and I were talking about this - the benefit of a shared repo is it's easier to coordinate versions and changes between "core" Mentat and the VS Code extension. We do want any code that would be shared by VS Code and NeoVim plugins here (and those bits would probably be useful as well for the community if they make extensions for other editors). @jakethekoenig is the way you see it interfering with vim plugin install just that the cloned repo would be bigger?

I have two concerns:
1. Yeah it would make the cloned repo bigger. I guess that's not actually a big concern because storage/bandwidth is cheap and if becomes one we can make instruct the users to use use `git clone --filter` and `git sparse-checkout` to just get what they need. Though it's complicated because normally if I wanted to instruct a user to do something non trivial I'd put it in a script for them to run but if they have access to the script they've already cloned the repo.
2. Vim plugins expect their files in a specific structure so we may have to litter the root of the project with autoload, doc, plugin and syntax and won't be able to bunch them all up under `vim`. I suppose there's no need to handle vscode and vim in the same way so we could have the vim plugin be a git submodule but the vscode plugin be a subdirectory like it is now.

Sorry if this is getting off topic for this PR.

waydegg:
Closing this b/c of #213

Issue #63, opened 2023-08-15
Separate Core Logic from Terminal Interface


biobootloader:
To prepare for future editor extensions and other interfaces, we need to separate the core logic from the terminal interface (both input and output). The goal is to allow for multiple interfaces: terminal, VSCode extension, NeoVim extension, etc

biobootloader:
A few other considerations:

The `UserInputManager` class as written would only be used for the terminal interface, since it's setup for `prompt_toolkit` completions, etc. It also has an `ask_yes_no` method that re-asks the user until it gets a yes or no: https://github.com/biobootloader/mentat/blob/3a77290355f8797971f83bbeaa172a41d8189bfe/mentat/user_input_manager.py#L84-L92

The implementation will have to watch out for how it handles these errors as well, maybe changing some of them even: https://github.com/biobootloader/mentat/blob/3a77290355f8797971f83bbeaa172a41d8189bfe/mentat/app.py#L72-L76

biobootloader:
The "MentatRunner" will also need a way to report that core mentat has crashed, and why, to the interface

biobootloader:
@granawkins @waydegg what are your thoughts on this?

waydegg:
I like the direction you're going with this.

At a glace I think we'll need base classes/abstractions for:
- The LLM loop (e.g. `MentatRunner`)
- User input (handling shortcuts like ctrl-c and mouse input for IDE extensions)
- Mentat output (handle output for non-terminal interfaces)

Having these 3 abstractions would allow people to start building working plugins/extensions without breaking the core logic.

waydegg:
Added in #119

PR #74, opened 2023-08-25
Separate Core Logic from Terminal Interface

biobootloader:
For #63 , a perquisite to merge #70 

Right now it's just a sketch of how this could work. Before merging we'd finalize design and replace all calls to `print` and `cprint` in the codebase, figure out error handling, and call user input messages from the new interface.

@waydegg @granawkins @PCSwingle let me know what you think, I'll continue to work on this tomorrow

The main idea is that we'll run output through the `interact` function - mostly it'll be output from Mentat to be displayed on the terminal or editor extension, but sometimes interact will be called with `return_user_input=True`, meaning the user should respond. Available user options (yes, no, etc) can be provided through `user_input_options`. I imagine in the future we'll sometimes send things to `interact` that aren't just text information, but raw data (costs, etc), and it'd be up to the interface how to display that.

Feel free to suggest better naming for anything as well

### DIFF
--- a/concept3.py
+++ b/concept3.py
@@ -1,0 +1,108 @@
+import sys
+from prompt_toolkit import PromptSession
+import signal
+import asyncio
+
+
+class MentatInterrupt(Exception):
+    pass
+
+
+class MentatInterface:
+    # should be implemented by subclasses
+    def display(self, content, color=None, end="\n"):
+        pass
+
+    # should be implemented by subclasses
+    async def get_user_input(self, prompt=None, options=None):
+        pass
+
+    # convenience function
+    async def ask_yes_no(self, prompt=None):
+        return (await self.get_user_input(prompt, options=["y", "n"])).lower() == "y"
+
+
+class TerminalMentatInterface(MentatInterface):
+    """Used to communicate with the terminal"""
+
+    def __init__(self):
+        self.session = PromptSession()
+
+    def display(self, content, color=None, end="\n"):
+        print(content, end=end, flush=True)
+        global sigint
+        if sigint:
+            sigint = False
+            raise MentatInterrupt()
+
+    async def get_user_input(self, prompt=None, options=None):
+        while True:
+            if prompt is not None:
+                self.display(f"{prompt} ({'/'.join(options)})")
+
+            user_input =  await self.session.prompt_async()
+
+            if options is None:
+                return user_input
+            elif user_input.lower() in options:
+                return user_input
+            else:
+                self.display("Invalid response")
+
+
+async def run_mentat(interface):
+    for i in range(5):
+        print(i)
+        await asyncio.sleep(1)
+
+    try:
+        interface.display("Welcome to Mentat!\nWhat can I do for you?")
+        while True:
+            prompt = await interface.get_user_input()
+            await get_llm_response(interface, prompt)
+    except MentatInterrupt:
+        print('Closing Mentat')
+
+async def get_llm_response(interface, prompt: str):
+    message = f"Ok, I'll {prompt}.\n"
+    try:
+        for char in message:
+            interface.display(char, end="")
+            await asyncio.sleep(0.05)
+    except MentatInterrupt:
+        print("\nUser interrupted stream\n")
+
+    # Get yes/no confirmation
+    if await interface.ask_yes_no("Apply these changes?"):
+        interface.display("Applying changes")
+    else:
+        interface.display("Not applying changes")
+    interface.display("What else can I do for you?")
+
+sigint = False
+
+def signal_handler():
+    print("### SIGINT ###")
+    global sigint
+    sigint = True
+
+class AsyncController:
+
+    def __init__(self):
+        self.loop = asyncio.get_event_loop()
+        self.loop.add_signal_handler(signal.SIGINT, signal_handler)
+
+    def start_mentat_with_interface(self, interface):
+        print("starting!")
+        self.loop.run_until_complete(run_mentat(interface))
+        print("done!")
+
+
+if __name__ == "__main__":
+    ac = AsyncController()
+
+    interface = TerminalMentatInterface()
+    ac.start_mentat_with_interface(interface)
+
+
+
--- a/mentat/app.py
+++ b/mentat/app.py
@@ -12,6 +12,7 @@ from .config_manager import ConfigManager, mentat_dir_path
 from .conversation import Conversation
 from .errors import MentatError, UserError
 from .git_handler import get_shared_git_root_for_paths
+from .interface import InterfaceType, get_interface, initialize_mentat_interface
 from .llm_api import CostTracker, setup_api_key
 from .logging_config import setup_logging
 from .user_input_manager import UserInputManager, UserQuitInterrupt
@@ -37,6 +38,9 @@ def run_cli():
     args = parser.parse_args()
     paths = args.paths
     exclude_paths = args.exclude
+
+    initialize_mentat_interface(InterfaceType.Terminal)
+
     run(expand_paths(paths), expand_paths(exclude_paths))
 
 
@@ -98,7 +102,9 @@ def loop(
     )
     conv = Conversation(config, cost_tracker, code_file_manager)
 
-    cprint("Type 'q' or use Ctrl-C to quit at any time.\n", color="cyan")
+    get_interface().interact(
+        "Type 'q' or use Ctrl-C to quit at any time.\n", color="cyan"
+    )
     cprint("What can I do for you?", color="light_blue")
     need_user_request = True
     while True:
--- a/mentat/app.py
+++ b/mentat/app.py
@@ -12,7 +12,7 @@ from .config_manager import ConfigManager, mentat_dir_path
 from .conversation import Conversation
 from .errors import MentatError, UserError
 from .git_handler import get_shared_git_root_for_paths
-from .interface import InterfaceType, get_interface, initialize_mentat_interface
+from .interface import InterfaceType, initialize_mentat_interface, output
 from .llm_api import CostTracker, setup_api_key
 from .logging_config import setup_logging
 from .user_input_manager import UserInputManager, UserQuitInterrupt
@@ -102,9 +102,7 @@ def loop(
     )
     conv = Conversation(config, cost_tracker, code_file_manager)
 
-    get_interface().interact(
-        "Type 'q' or use Ctrl-C to quit at any time.\n", color="cyan"
-    )
+    output("Type 'q' or use Ctrl-C to quit at any time.\n", "cyan")
     cprint("What can I do for you?", color="light_blue")
     need_user_request = True
     while True:
--- a/mentat/app.py
+++ b/mentat/app.py
@@ -3,8 +3,6 @@ import glob
 import logging
 from typing import Iterable, Optional
 
-from termcolor import cprint
-
 from .code_change import CodeChange
 from .code_change_display import print_change
 from .code_file_manager import CodeFileManager
@@ -12,10 +10,10 @@ from .config_manager import ConfigManager, mentat_dir_path
 from .conversation import Conversation
 from .errors import MentatError, UserError
 from .git_handler import get_shared_git_root_for_paths
-from .interface import InterfaceType, initialize_mentat_interface, output
+from .interface import InterfaceType, MentatInterface, initialize_mentat_interface
 from .llm_api import CostTracker, setup_api_key
 from .logging_config import setup_logging
-from .user_input_manager import UserInputManager, UserQuitInterrupt
+from .user_input_manager import UserQuitInterrupt
 
 
 def run_cli():
@@ -35,16 +33,31 @@ def run_cli():
         default=[],
         help="List of file paths, directory paths, or glob patterns to exclude",
     )
+    parser.add_argument(
+        "--interface",
+        "-i",
+        choices=[InterfaceType.Terminal, InterfaceType.VSCode],
+        default=InterfaceType.Terminal,
+        help="Interface type",
+    )
     args = parser.parse_args()
     paths = args.paths
     exclude_paths = args.exclude
+    interface_type = args.interface
 
-    initialize_mentat_interface(InterfaceType.Terminal)
+    interface = initialize_mentat_interface(interface_type)
 
-    run(expand_paths(paths), expand_paths(exclude_paths))
+    run(
+        interface, 
+        expand_paths(interface, paths), 
+        expand_paths(interface, exclude_paths)
+    )
 
 
-def expand_paths(paths: Iterable[str]) -> Iterable[str]:
+def expand_paths(
+    interface: MentatInterface, 
+    paths: Iterable[str]
+) -> Iterable[str]:
     globbed_paths = set()
     invalid_paths = []
     for path in paths:
@@ -54,16 +67,20 @@ def expand_paths(paths: Iterable[str]) -> Iterable[str]:
         else:
             invalid_paths.append(path)
     if invalid_paths:
-        cprint(
+        interface.display(
             "The following paths do not exist:",
             "light_yellow",
         )
-        print("\n".join(invalid_paths))
-        exit()
+        interface.display("\n".join(invalid_paths))
+        interface.exit()
     return globbed_paths
 
 
-def run(paths: Iterable[str], exclude_paths: Optional[Iterable[str]] = None):
+def run(
+    interface: MentatInterface,
+    paths: Iterable[str], 
+    exclude_paths: Optional[Iterable[str]] = None
+):
     mentat_dir_path.mkdir(parents=True, exist_ok=True)
     setup_logging()
     logging.debug(f"Paths: {paths}")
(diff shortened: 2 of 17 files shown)


mentat/app.py
```diff
@@ -98,7 +102,9 @@ def loop(
     )
     conv = Conversation(config, cost_tracker, code_file_manager)
 
-    cprint("Type 'q' or use Ctrl-C to quit at any time.\n", color="cyan")
+    get_interface().interact(
+        "Type 'q' or use Ctrl-C to quit at any time.\n", color="cyan"
+    )
```
biobootloader, line 107: we could also add a convenience function to `interface.py` for common interact calls, like ones that just use `content` and `color`. It'd just take those two args and internally call `get_interface`

mentat/interface.py
```diff
@@ -0,0 +1,116 @@
+from abc import ABC, abstractmethod
+from enum import Enum
+from typing import Callable, Dict, Iterable, Optional
+
+from termcolor import cprint
+
+_interface_instance = None
+
+
+def get_interface() -> "MentatInterface":
```
PCSwingle, line 10: We can use from __future__ import annotations to avoid the quotation marks :D

PCSwingle:
Really like this outline; one thing I'm wondering about is that it feels like we're mixing two patterns here: for all of the other singletons, we just pass them around everywhere, and for this one we just use a global instance. I think we should standardize and either pass the singleton around (I prefer this idea) or make every singleton a global. Either way, I really like the look of this! This should make it super easy to make different interfaces

mentat/interface.py
```diff
@@ -0,0 +1,123 @@
...
+    def interact(
+        self,
+        content: Optional[str] = None,
+        color: Optional[str] = None,
+        is_error: Optional[bool] = False,
+        return_user_input: Optional[bool] = False,
+        user_input_options: Optional[Iterable[str]] = None,
+    ) -> str:
+        """Handle both receiving input and sending output using the provided callback."""
+        pass
```
biobootloader, line 60: Actually instead of "interact" we can have a few different functions, input, output, whatever we want. Then the specific `MentatInterface` implementation (like `VSCodeMentatInterface` can decide if it's best to use a single or multiple callbacks to handle them.

mentat/interface.py
```diff
@@ -0,0 +1,123 @@
...
+
+    @abstractmethod
+    def interrupt(self):
+        """Handle user interruptions."""
+        pass
+
+    @abstractmethod
+    def exit(self):
+        """Handle cleanup or exit procedures."""
+        pass
```
biobootloader, line 70: these wouldn't be called by Mentat itself, they'd be called from the interface side when the user presses a button to interrupt or exit. So we'll have to think about how to set up the whole project correctly with asyncio

mentat/interface.py
waydegg: Yup I was thinking the same thing. Left thoughts here: https://github.com/biobootloader/mentat/issues/63#issuecomment-1694540114

waydegg:
> Really like this outline; one thing I'm wondering about is that it feels like we're mixing two patterns here: for all of the other singletons, we just pass them around everywhere, and for this one we just use a global instance. I think we should standardize and either pass the singleton around (I prefer this idea) or make every singleton a global. Either way, I really like the look of this! This should make it super easy to make different interfaces

Also agree w/ this

granawkins:
> Really like this outline; one thing I'm wondering about is that it feels like we're mixing two patterns here: for all of the other singletons, we just pass them around everywhere, and for this one we just use a global instance. I think we should standardize and either pass the singleton around (I prefer this idea) or make every singleton a global. Either way, I really like the look of this! This should make it super easy to make different interfaces

+1

biobootloader:
Sounds like we all agree with @PCSwingle mixing patterns isn't good. So let's figure out which pattern to use (pass around singleton instances vs use global). The singletons we currently have:

- MentatInterface (defined here)
- Conversation
- CodeFileManager
- UserInputManager
- ConfigManager

`UserInputManager` will be subsumed by the `TerminalMentatInterface`. `Conversation` maybe wouldn't always be a singleton? In the future we may have alternative conversation paths.

I expect `CodeFileManager` and `ConfigManager` to always remain singletons though, we probably just want a single instance doing file i/o and we'll just have one config loaded at a time.

granawkins:
Sorry for the slow feedback on this, I wanted to wrap my head around the codebase before jumping in with suggestions -

A couple other thoughts:

1. For VSCode, or any GUI, It'd be nice if one could have multiple MentatRunners at the same time, like different conversations in tabs. Maybe some things would be shared between all instances, and others (`Conversation`) would be instance-specific. Would it make sense to have those be properties of the MentatRunner?

2. WRT loops (this was also brought up in #70). I think the CLI has an extra loop that other interfaces might not need. Something like:
```
class MentatRunner:
    async def get_response(self, prompt):
        # Trigger sending a response to the interface

class TerminalMentatInterface(MentatInterface):
    async def run(self):
        while True:
            user_input = await... # Wait for user input
            await runner.get_response(user_input)
```
So the 'default' pattern is you call `get_response` when you want Mentat to do something, otherwise it's dormant. But for the Terminal, you have a loop to watch for input and feed it to Mentat.

granawkins:
I just added a file, `mentat_runner_concept.py`, which I wrote to illustrate one way of handling all these loops. Scott and I just talked about it and we're going to do some more thinking, and welcome input from anyone else.

The main idea is that MentatRunner itself doesn't have a built-in loop; you just call `get_response` when you have a prompt, and provide an interface for printing the output. Between calls to get_response, the model is dormant. I'm thinking this will make it simpler to (1) support other interfaces like VSCode, which don't use a loop and (2) make it easier to eventually run multiple instances at once.

Some open questions we landed on:
1. How are KeyboardInterrupt and Errors, especially inside async functions, handled?
2. If we have multiple instances in the future, is each 'instance' contained in a MentatRunner (with all state in `self` variables), or is an instance contained in a process (and we continue using singletons)?

Again this is just a concept, a lot of this stuff is redundant, and I'm not married to any of this.
image

biobootloader:
Thank you @granawkins, this proof of concept is really helpful!

biobootloader:
I've added `concept2.py` an iteration/alternative of `mentat_runner_concept.py`.

A few key differences:
- instead of the interface/extension calling "get_response" to send the user prompt to Mentat, Mentat instead calls `get_user_input` to request input from the user. This is used for both user prompts and y/n responses.
- no special mode for streaming, which simplifies things (don't need `_is_streaming`, `_was_streaming`).
- Core Mentat has control of the flow

A few caveats:
- I haven't added interrupt handling yet
- When the user enters input on VSCode, with `mentat_runner_concept.py` design it simply results in a call to `get_response`. With the `concept2.py` design, the user input needs to be returned from `get_user_input`. So when `get_user_input` is called the VSCode implementation would wait to return until user input has been entered.

A caveat on both approaches:
- I'm still not sure how we can handle the current terminal behavior of interrupting the model response / quitting when the user hits Ctrl-C, since I'm not sure yet where to put the `except KeyboardInterrupt` and what happens to tasks in the event loop

granawkins:
I made some small updates to concept2 (seemed clearer than making concept3) and added a concept2_runner:

- It seems a lot of this doesn't need to be async. I believe Ctrl+C will cancel the `asyncio.run` loop, wherever it is. I think we want it to interrupt mentat funcs (e.g. `get_llm_response`) so everything above that call should be non-async. I made this change and KeyboardInterrupt seems to be functioning like we want.
- I made a new `Runner` which starts Mentat as a subprocess, and uses the same blocking pattern as the Terminal interface. Also setup VSCodeInterface to support this.

@biobootloader I tried to just build what you described, let me know if I got it right! And if I'm missing anything about async?

granawkins:
I went ahead and wrote a working implementation of `MentatInterface` and `TerminalMentatInterface` based on our conversations so far. My aim was to (1) follow the structure @biobootloader laid out in `concept2.py`, and (2) change as little as possible in the code. 

A couple thoughts here:
- I made `interface` such that the instance is passed around, rather than it being a global variable, just following the 'consistency' idea from elsewhere.
- Indeed, `self.interface.display` is a bit clunky. What do you think of `self.ui.print` instead?

Anyway the cli seems to run just like it was before, though I will do a bit more testing. I feel pretty confident that this structure can accommodate everything I had in mind for VSCode - I'll take a go at that next.

granawkins:
For the VSCode side, I have an idea that might get the best of everything:

1. A `JSONMentatInterface`, which send everything to `stdout` as a JSON object. e.g. `{'content': 'Accept Changes? (Y/n)', 'color': 'red', 'end': '/n', 'options': ['y', 'n', 'i']}`. It also expects input from `stdin` as JSON, e.g. `{'prompt': }`
2. A `MentatRunner`, which acts kindof like a server: it runs mentat in a subprocess, accepts a `handle_display` function and exposes `get_response(prompt)` and `interrupt` (which sends KeyBoardInterrupt). It handles all conversions to/from JSON and manages stdin/stdout.

Thoughts?

biobootloader:
Thank you @granawkins this is good stuff! Going to comment on your iteration of concept2 first:

> It seems a lot of this doesn't need to be async. I believe Ctrl+C will cancel the asyncio.run loop, wherever it is. I think we want it to interrupt mentat funcs (e.g. get_llm_response) so everything above that call should be non-async.

Yes, currently Mentat is only async when we call `get_llm_response`. In my concept I'd made more parts async because I was thinking we'd be running it async along with the Python side of the LSP for the VS Code extension (i.e. both in the same process / async event loop). Your `concept2_runner` allows for us to keep just the llm call async, which is pretty nice!

biobootloader:
> For the VSCode side, I have an idea that might get the best of everything:
> 
> 1. A `JSONMentatInterface`, which send everything to `stdout` as a JSON object. e.g. `{'content': 'Accept Changes? (Y/n)', 'color': 'red', 'end': '/n', 'options': ['y', 'n', 'i']}`. It also expects input from `stdin` as JSON, e.g. `{'prompt': }`
> 2. A `MentatRunner` class, which acts kindof like a server: it runs mentat in a subprocess, accepts a `handle_display` function and exposes `get_response(prompt)` and `interrupt` (which sends KeyBoardInterrupt). It handles all conversions to/from JSON and manages stdin/stdout.
> 
> Thoughts?

@waydegg what do you think about this? I know on the other PR you commented we should do things async instead of multiprocess (https://github.com/biobootloader/mentat/pull/70#issuecomment-1700256163), but maybe this simplifies things?

https://github.com/biobootloader/mentat/blob/1d232bd5e1fe389d0820b188aa0c8d0623cd41c7/concept2_runner.py#L3-L33

waydegg:
Love the concept/concept runner files! Super helpful and I like a lot of the design choices mentioned so far. Here's some feedback I have on a few points that have been mentioned. I'm probably missing some stuff from this comment thread but these ares some topics that stood out to me:

### Multiple `MentatRunner` instances

I think that we should be designing the core logic around this (i.e. having multiple "chats" running concurrently).

### Processing loops

I think it's totally ok to have a ton of `for` and `while` loops running as long as the code inside them isn't blocking asyncio's event loop. Maybe I'm missing something, but getting user input and sending model output on any of the interfaces (Terminal, VSCode) can and should be async. 

I'm going to take a stab at writing out another pseudo-code concept file tomorrow, but I'm thinking this could be a potential way to structure things:

- There is a single "main loop" that runs on an infinite cycle, managing coroutines/task groups and listening to `SIGINT` and `SIGKILL`. I've seen a lot of async, python-based job processing libraries and webservers are designed like this (see how [Uvicorn](https://github.com/encode/uvicorn/tree/master) does it with [server.py](https://github.com/encode/uvicorn/blob/master/uvicorn/server.py)). 
- Every `MentatRunner` is a coroutine that's run as child `asyncio.Task` managed by the main loop. Any number of language server connections (managed by [pygls](https://github.com/openlawlibrary/pygls)) would run here too. 

A lot of these task will be sharing data with each other, so having them all under the same python process will make this wayyy easier to develop and debug.

### Streaming data
  
There's 2 ways data should be sent *from* OpenAI to the user:

1. In-process (for terminal apps)
2. Over-the-wire (for non-terminal extensions)

I like @granawkins way of describing Mentat almost like it's a server, and extensions are going to interact/interface with this server to function properly. For whatever data transport method we decide on, I think some things to think about are:
- bi-directional communication
- compatibility with most execution environments
- message idempotency
- good python library/ecosystem support

(JSON) RPC (or some variant) comes to mind since it checks all of those points and it's what a lot of Vim/Neovim plugins and VSCode extensions use.


### Handling `SIGINT` and `SIGTERM` 

Going to use Uvicorn again as an example, but see how they setup their signal handlers [here](https://github.com/encode/uvicorn/blob/fc459544a058f59c04700a6c5ece6b6171a1ea74/uvicorn/server.py#L315-L334) (I'm envisioning something like this for Mentat).  You'd have some variable/object every `asyncio.Task` can access that holds some value for determining if they should `.cancel()` themselves or you force-kill everything from the main processing loop.

One thing to be aware of is how 3rd-party libraries handle `KeyboardInterrupt` or any of the signals (prompt toolkit specifically). I actually ran into this issue recently with #82  when I was trying to throw a `KeyboardInterrupt` inside of a prompt toolkit keybind handler.

biobootloader:
I just added `concept3.py`, which attempts to use a single process and add a signal handler to the event loop to manage ctrl-c usage. But it's not quite working...  I'll add a couple comments directly on the file

concept3.py
```diff
@@ -0,0 +1,108 @@
...
+    if await interface.ask_yes_no("Apply these changes?"):
+        interface.display("Applying changes")
+    else:
+        interface.display("Not applying changes")
+    interface.display("What else can I do for you?")
+
+sigint = False
+
+def signal_handler():
+    print("### SIGINT ###")
```
biobootloader, line 85: just for this proof of concept I'm toggling a global switch when ctrl-c is pressed (`signal.SIGINT`)

concept3.py
```diff
@@ -0,0 +1,108 @@
...
+            elif user_input.lower() in options:
+                return user_input
+            else:
+                self.display("Invalid response")
+
+
+async def run_mentat(interface):
+    for i in range(5):
+        print(i)
+        await asyncio.sleep(1)
```
biobootloader, line 56: during this 5 second count, if you hit ctrl-c the signal handler does work - it'll print, and after this the global will be read and Mentat will exit

concept3.py
```diff
@@ -0,0 +1,108 @@
...
+
+    def __init__(self):
+        self.session = PromptSession()
+
+    def display(self, content, color=None, end="\n"):
+        print(content, end=end, flush=True)
+        global sigint
+        if sigint:
+            sigint = False
+            raise MentatInterrupt()
```
biobootloader, line 36: the idea here was we'd check the global whenever Mentat tries to display something, and raise a `MentatInterrupt` if the user had pressed ctrl-c (or, for the VSCode extension, if the user had pressed some button or hotkey to interrupt / quit).

The `MentatInterrupt` gets caught in a try/except wrapped around streaming, to just interrupt the stream. If we aren't in streaming, it gets caught higher up to quit Mentat.

concept3.py
```diff
@@ -0,0 +1,108 @@
...
+        print(i)
+        await asyncio.sleep(1)
+
+    try:
+        interface.display("Welcome to Mentat!\nWhat can I do for you?")
+        while True:
+            prompt = await interface.get_user_input()
+            await get_llm_response(interface, prompt)
+    except MentatInterrupt:
+        print('Closing Mentat')
```
biobootloader, line 64: once we get here, after the 5 second count, the signal handler no longer works. I don't know why, but we get a crash with a `KeyboardInterrupt`. I'll continue looking into why tomorrow

biobootloader:
ok I got concept3 working, aka control-c interrupts work at all stages for the terminal interface. It's ugly though, the reason it wasn't working before was that `prompt-toolkit` was adding it's own signal handler to handle sigints

waydegg:
## "Concept 4"

Just added a "concept4" that implements a basic example + pseudocode based on the comment I left [earlier](https://github.com/biobootloader/mentat/pull/74#issuecomment-1703693866). This "concept4" just showcases how the general flow of data b/w Mentat and the user can look.

There's 2 files:
- [core.py](https://github.com/biobootloader/mentat/blob/separate-core-logic-and-interface/concept4/core.py)
- [terminal_client.py](https://github.com/biobootloader/mentat/blob/separate-core-logic-and-interface/concept4/terminal_client.py)

The idea is having an RPC server (for any non-terminal clients like the VSCode extension) expose a `MentatEngine` which is a single class/instance that exposes all of Mentat's functionality.  The terminal client can use `MentatEngine` directly since they can be run in the same process asynchronously.

### High-level Implementation strategy

To fully build this and get this working with what features Mentat currently supports, I think the work could be broken down like so:

1. **Make existing Mentat entities generic**

We want to make things like `CodeContext`, `Conversation`, `Command` and `MentatCompleter` generic for any client. These have all been designed w/ just a Terminal in mind (since that's all that's supported atm), so they'll need to be updated. Some of these also have parts that are blocking so they'll need to be made async.

2. **Create the RPC server**

[Rift](https://github.com/morph-labs/rift/blob/350a4195077ec2d1ec9b415a19afb6d84c8c69f7/rift-engine/rift/rpc/jsonrpc.py) has a good implementation of an async json RPC server for reference.

3. **Create the client libraries**

At this point the VSCode extension can start building on top of this work and the Terminal client can be updated to use this code.

granawkins:
Lots of good ideas on the table. Feels like with a bit more alignment we put these pieces together and get moving. I could write like 200 lines of my thoughts, or maybe the 3 of us could just hop on a call and discuss?

waydegg:
Added in #119

PR #79, opened 2023-08-26
Add auto completions

waydegg:
# Summary 

This PR adds completions for syntax (classes, variables, functions, etc.) and file names for the prompt session.

Completions kept in sync as files are added/updated/deleted (they are "refreshed" every 5 seconds atm). Whenever a user starts a prompt with "/" we return command completions only. 

Since we need to keep track of what files are in the context for keeping completions up-to-date, I've added a `CodeFileIndex` class that currently just holds a set of file names added into the context. `CodeFileIndex` is shared between `CodeFileManager` and `UserInputManager`. This is done so that when `CodeFileIndex` modifies the filesystem, `UserInputManager can see these modifications and make updates to completions.

# To-Do

- [x] Update test cases for `CodeFileManager`
- [ ] Add test cases for `MentatCompleter`

### DIFF
--- a/mentat/code_file_manager.py
+++ b/mentat/code_file_manager.py
@@ -13,13 +13,10 @@ from .change_conflict_resolution import (
     resolve_non_insertion_conflicts,
 )
 from .code_change import CodeChange, CodeChangeAction
+from .code_file_index import CodeFileIndex
 from .config_manager import ConfigManager
-from .errors import MentatError, UserError
-from .git_handler import (
-    get_git_diff_for_path,
-    get_non_gitignored_files,
-    get_paths_with_git_diffs,
-)
+from .errors import MentatError
+from .git_handler import get_git_diff_for_path, get_paths_with_git_diffs
 from .user_input_manager import UserInputManager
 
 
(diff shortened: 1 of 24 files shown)


PCSwingle:
Just fyi, in my commands PR #59 I change how we use prompt toolkit a lot; you might want to work off of that branch until it gets merged.

waydegg:
@biobootloader would love your thoughts are on `CodeFileIndex` and if you can think of a more elegant way to share file paths w/ `UserInputManager` (i.e. less changes required for `CodeFileManager`)

mentat/mentat_prompt_session.py
```diff
@@ -42,15 +60,142 @@ def get_suggestion(self, buffer: Buffer, document: Document) -> Suggestion | Non
             return super().get_suggestion(buffer, document)
 
 
+### TODO
+# 4. Decide on what words to auto complete (right now it's everything)
+#   a. Use ctags to add types (e.g. function, class) to completions
+# 5. Use ctags if available, fallback to pygments lexer if not
+# 6. Consider fuzzy matching completions
```
biobootloader, line 67: I love fuzzy matching in general. I guess the downside could be too many options popping up for everything you type?

Rift has an `@` syntax for referencing files, maybe something like that could be good

biobootloader:
> @biobootloader would love your thoughts are on `CodeFileIndex` and if you can think of a more elegant way to share file paths w/ `UserInputManager` (i.e. less changes required for `CodeFileManager`)

I think it is going to require changes to `CodeFileManager`. I guess the question is how do we divide responsibilities best between:
- `CodeFileManager`
- `CodeMap`
- `CodeFileIndex`
- `MentatCompleter`

Both `CodeFileManager` and `MentatCompleter` need to know which files / partial files are in the context, and this will grow more complicated as we add features to Mentat that automatically add partial files to context as the conversation continues. So your approach is to make `CodeFileIndex` to do that shared work.

Maybe a better name for `CodeFileIndex` could something like `ContextManager`? Maybe it should also own `get_code_message()` (currently in `CodeFileManager`), and `CodeFileManager` should only be responsible for reading and writing to files?

Another question is what we should auto-complete - just stuff in context or also stuff that's not in context, like from the `CodeMap`? I think including stuff that's outside context will be important because we'll want user to be able add files/classes/functions to context easily, within a conversation. But maybe that'd be too many completions being suggested for big codebases? The completion engine should probably know which completions are coming from within or outside context, so it could prioritize things in context.

I'm not really sure about any of this, but those are some initial thoughts!

biobootloader:
I don't think this branch has the partial files changes from #69, sorry that that complicates things!

mentat/mentat_prompt_session.py
waydegg: You can limit the number of completions shown. Prompt Toolkit has a built-in fuzzy matcher that's pretty fast too.

Yea I like the `@` directive for Rift and Cursor. Happy to make that character show file completions specifically.

waydegg:
> Maybe a better name for CodeFileIndex could something like ContextManager? Maybe it should also own get_code_message() (currently in CodeFileManager), and CodeFileManager should only be responsible for reading and writing to files?

Yup that sounds good. Make `CodeFileManager` just handle CRUD actions for files, and `ContextManager` is where references to files and partial files are stored.

waydegg:
> Another question is what we should auto-complete - just stuff in context or also stuff that's not in context, like from the CodeMap?

Right now it's just stuff inside the context, however I don't think it would hurt to add completions for files/syntax inside git repos? I think we could enable/disable this depending on if the code-map is enabled/disabled.

waydegg:
> I don't think this branch has the partial files changes from https://github.com/biobootloader/mentat/pull/69, sorry that that complicates things!

Good callout! Will fetch from main and update this

pyproject.toml
```diff
@@ -4,3 +4,4 @@ known_first_party = "mentat"
 
 [tool.ruff]
 line-length = 120
+ignore = ["E731"]
```
waydegg, line 7: Added this rule to ignore lambda assignment (https://beta.ruff.rs/docs/rules/lambda-assignment/) as we have these everywhere and imo they're ok as long as you don't have a ton of logic in them.

waydegg:
In an effort to keep the number of changes down in this PR, I think we should add fuzzy matching in a separate feature branch. Same goes for the `@` directive since it's basically another completer we're adding (e.g. another feature).

waydegg:
I don't think it's worth the effort to generate completions from `ctags`. You're only able to get top-level syntax words (can't get anything inside classes/functions), so the `pygments` lexer does a pretty good job at this as is. 

I also decided to not include completions from all files in the git repo (if there is one), only files explicitly added in the code context, since there's too much irrelevant information being suggested. This is the how VSCode does their compeltions anyways which I think is a good choice.

granawkins:
I read through this and tested it out, and it looks good to me. One comment: eventually it'd be nice if this works with different interfaces. Like if I could call `refresh_completions` from the VSCode extension somehow with its current user input. This can def be done later, just wanted to throw it out there.

biobootloader:
> I read through this and tested it out, and it looks good to me. One comment: eventually it'd be nice if this works with different interfaces. Like if I could call `refresh_completions` from the VSCode extension somehow with its current user input. This can def be done later, just wanted to throw it out there.

Agreed, "core" Mentat would provide lists of available completions to all interfaces

biobootloader:
> In an effort to keep the number of changes down in this PR, I think we should add fuzzy matching in a separate feature branch. Same goes for the `@` directive since it's basically another completer we're adding (e.g. another feature).

Agreed, let's make followup issues for these. The `@` approach also relates to https://github.com/biobootloader/mentat/issues/17

waydegg:
>  One comment: eventually it'd be nice if this works with different interfaces

Totally. I'd like to have the completions using pygls to pull from language servers (assuming we can get it working outside of VSCode). Would make the auto-completions tool-agnostic and higher quality.

biobootloader:
not a big deal because it's rare, but both kinds of completitons can be suggested at once (the history one take priority if you hit tab): 
image

I don't think it's worth trying to fix - not sure how often most people would use history completions anyway

waydegg:
> both kinds of completitons can be suggested at once

Do you have any preference on how we should handle this? It's pretty common for me to see both completion sources, so I think it's worth looking into.

I'm thinking we could copy how [ipython](https://github.com/ipython/ipython) does it and have arrow keys finish/cycle command history completions, and have tab/shift+tab or ctrl+n/ctrl+p finish/cycle auto-completions.

mentat/code_file_manager.py
```diff
@@ -60,140 +56,54 @@ def _print_path_tree(tree, changed_files, cur_path, prefix=""):
...
+        cprint(self.config.git_root.name, "blue")
         _print_path_tree(
-            _build_path_tree(self.files, self.git_root),
-            get_paths_with_git_diffs(self.git_root),
-            self.git_root,
+            _build_path_tree(self.code_context.files.values(), self.config.git_root),
+            get_paths_with_git_diffs(self.config.git_root),
+            self.config.git_root,
         )
         print()
```
biobootloader, line 81: I think displaying the project tree / include files should be done by CodeContext. And instead of being in the __init__ should be in something like CodeContext.display_context_tree(), called in app.py after initializing the CodeContext.

biobootloader:
> I'm thinking we could copy how [ipython](https://github.com/ipython/ipython) does it and have arrow keys finish/cycle command history completions, and have tab/shift+tab or ctrl+n/ctrl+p finish/cycle auto-completions.

That sounds great!

biobootloader (approved):
This is awesome!

PR #83, opened 2023-09-01
Clear prompt with ctrl-c before quitting

waydegg:
Instead of exiting immediately on ctrl-c, we clear any text that's been typed. This is pretty common behavior for other cli tools, so I added this since I keep accidentally quitting mentat sessions after trying to clear a bunch of text I just typed.


biobootloader (approved):
Nice!

PR #84, opened 2023-09-02
Add pyright

waydegg:
# Summary

This PR adds [pyright](https://github.com/microsoft/pyright), a static type checker, as a dependency and type checks code with Github Actions.

The goal of this is to make it development and contributing easier. With enforced type checking you get:
- Fewer errors in your code. Static type checking helps catch a lot of errors prior to runtime
- Better readability. Seeing the argument and return types for everything makes understanding someone else's code way easier.
- Better editor/LSP completions. Properly typed python code makes intellisense for VSCode or whatever editor you use work way better

# How to merge this

Right now we have a bunch of type checking errors that will cause the CI/CD build step to fail. If we want to add pyright support, I think we should fix those typing errors in this PR.

### DIFF
--- a/mentat/llm_api.py
+++ b/mentat/llm_api.py
@@ -2,11 +2,12 @@ import logging
 import os
 import sys
 from dataclasses import dataclass
-from typing import Generator
+from typing import Any, AsyncGenerator, cast
 
 import openai
 import tiktoken
 from dotenv import load_dotenv
+from openai.error import AuthenticationError
 from termcolor import cprint
 
 from .config_manager import mentat_dir_path, user_config_path
@@ -14,11 +15,12 @@ from .errors import MentatError, UserError
 
 package_name = __name__.split(".")[0]
 
+# openai doesn't seem to use type hints, so we have to use type: ignore and cast everywhere
+
 
 # Check for .env file or already exported API key
 # If no api key found, raise an error
 def setup_api_key():
-    mentat_dir_path / ".env"
     if not load_dotenv(mentat_dir_path / ".env"):
         load_dotenv()
     key = os.getenv("OPENAI_API_KEY")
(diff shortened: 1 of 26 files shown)


waydegg:
[Here's](https://news.ycombinator.com/item?id=34222407) a recent-ish HN article that has some good discussion on Pyright vs Mypy (another static type checker). Mypy has been around for a lot longer so most python projects that have static type checking will use it over Pyright, however for newer projects people have been opting to use both Pyright and Mypy or just Pyright by itself.

biobootloader:
`pyright` seems great, lets do it!

Let's get https://github.com/biobootloader/mentat/pull/74 merged first and then fix all the typing errors / add typing

PCSwingle:
Super excited for this!!!

jakethekoenig:
Mostly just wanted to test pushing to your branch. I'm down to help out fixing some of the type errors once 74 is merged

PCSwingle:
Just FYI, I rebased on main and force pushed.

PCSwingle:
Just finished adding strict typing for pyright and also replaced a lot of places where we used strings with Paths. Let me if you have any feedback!

PCSwingle:
Also worth mentioning that I purposefully didn't include the tests in pyrights config file; I don't think typing there is too necessary, but if anybody disagrees I'm open to suggestions!

mentat/llm_api.py
```diff
@@ -112,7 +122,7 @@ def choose_model(messages: list[dict[str, str]], allow_32k) -> str:
 
 @dataclass
 class CostTracker:
-    total_cost: int = 0
+    total_cost: float = 0
```
jakethekoenig, line 125: I'd prefer to keep track of money as cents so it can be an int. Not super critical since we're not using the number to bill.

jakethekoenig (approved):
Thanks for doing all this. Did you end up using an LLM for help or did you do it by hand?

mentat/llm_api.py
PCSwingle: The way we calculate it is by dividing the tokens by 1000 and multiplying by the cost, so no matter what we do it will always be a float; definitely in any scenario that wasn't a rough estimate I agree that money should always be an int though!

PCSwingle:
> Thanks for doing all this. Did you end up using an LLM for help or did you do it by hand?

I ended up just doing it by hand haha. I definitely should have tried using Mentat though (although I have a suspicion it wouldn't have been able to do this because there were some obscure errors)!

PCSwingle (approved):

Issue #81, opened 2023-08-31
add an option to run Mentat on a the diff between two commits or branches


biobootloader:
Currently, Mentat is aware of uncommitted changes to files in context (the git diff for these files).

It'd be nice if you could make it aware of changes that have been committed. Like have it look at the changes from the last 3 commit, or the difference between two branches (a PR review assistant?). This could make it more helpful at continuing changes you are working on, or helping you review a PR.

This could also be a way to automatically give context - instead of manually choosing files, just look at all the files that have changes.

granawkins:
I can take a go at this. I'm thinking 3 new (optional) arguments:

1. `--history -h *int`: Include the diffs of last-n commits
2. `--commits -c *hash`: List of specific commits
3. `--branches -b *branch_name`: List of specific branches

- Diffs are split up and appended to the files for the prompt (like the current git diff)
- If you pass a `paths` arg, it includes diffs for only those paths. `--exclude-paths` applies.
- If you don't pass a `paths` arg, by default it includes all files referenced by diffs

Most of this takes place inside `CodeFileManager` and a new class, `CodeDiff` (representing one commit- or branch-diff).

So for your examples above:
- Changes from last 3 commits: `mentat -h 3`
- Difference between 2 branches: `mentat -b other-branch`
- Look at all files that have changed in a commit: `mentat -c commit-hash`
- (bonus) Recent activity on a particular file: `mentat myfile.py -h 5`

granawkins:
Done for now with #86

PR #86, opened 2023-09-06
Diff context

granawkins:
Optionally add the diff of a commit or branch to the system prompt. Goes with #81

### DIFF
--- a/mentat/app.py
+++ b/mentat/app.py
@@ -15,7 +15,7 @@ from .code_file_manager import CodeFileManager
 from .code_map import CodeMap
 from .config_manager import ConfigManager, mentat_dir_path
 from .conversation import Conversation
-from .diff_context import DiffContext
+from .diff_context import get_diff_context
 from .errors import MentatError, UserError
 from .git_handler import get_shared_git_root_for_paths
 from .llm_api import CostTracker, setup_api_key
@@ -151,7 +151,7 @@ def loop(
 ) -> None:
     git_root = get_shared_git_root_for_paths(paths)
     config = ConfigManager(git_root)
-    diff_context = DiffContext(config, history, commit, branch)
+    diff_context = get_diff_context(config, history, commit, branch)
     if not paths:
         paths = diff_context.files
     code_context = CodeContext(config, paths, exclude_paths or [])
--- a/mentat/app.py
+++ b/mentat/app.py
@@ -15,12 +15,12 @@ from .code_file_manager import CodeFileManager
 from .code_map import CodeMap
 from .config_manager import ConfigManager, mentat_dir_path
 from .conversation import Conversation
+from .diff_context import DiffContext
 from .errors import MentatError, UserError
 from .git_handler import get_shared_git_root_for_paths
 from .llm_api import CostTracker, setup_api_key
 from .logging_config import setup_logging
 from .user_input_manager import UserInputManager, UserQuitInterrupt
-from .diff_context import DiffContext
 
 
 def run_cli():
--- a/mentat/app.py
+++ b/mentat/app.py
@@ -55,14 +55,14 @@ def run_cli():
     parser.add_argument(
         "--commit",
         "-c",
-        nargs="*",
+        type=str,
         default=None,
         help="Commit hash to include diff for in system prompt",
     )
     parser.add_argument(
         "--branch",
         "-b",
-        nargs="*",
+        type=str,
         default=None,
         help="Branch to include diff for in system prompt",
     )
@@ -72,13 +72,13 @@ def run_cli():
     exclude_paths = args.exclude
     no_code_map = args.no_code_map
     history = args.history
-    commit = None if not args.commit else args.commit[0]
-    branch = None if not args.branch else args.branch[0]
+    commit = args.commit
+    branch = args.branch
     # Expanding paths as soon as possible because some shells such as zsh automatically
     # expand globs and we want to avoid differences in functionality between shells
     run(
-        expand_paths(paths), 
-        expand_paths(exclude_paths), 
+        expand_paths(paths),
+        expand_paths(exclude_paths),
         no_code_map,
         history,
         commit,
@@ -116,8 +116,8 @@ def run(
     exclude_paths: Optional[Iterable[str]] = None,
     no_code_map: bool = False,
     history: int = 0,
-    commit: Iterable[str] = [],
-    branch: Iterable[str] = [],
+    commit: Optional[str] = None,
+    branch: Optional[str] = None,
 ):
     mentat_dir_path.mkdir(parents=True, exist_ok=True)
     setup_logging()
@@ -146,8 +146,8 @@ def loop(
     cost_tracker: CostTracker,
     no_code_map: bool,
     history: int,
-    commit: Iterable[str],
-    branch: Iterable[str],
+    commit: Optional[str],
+    branch: Optional[str],
 ) -> None:
     git_root = get_shared_git_root_for_paths(paths)
     config = ConfigManager(git_root)
@@ -158,7 +158,9 @@ def loop(
     code_context.display_context()
     diff_context.display_context()
     user_input_manager = UserInputManager(config, code_context)
-    code_file_manager = CodeFileManager(user_input_manager, config, code_context, diff_context)
+    code_file_manager = CodeFileManager(
+        user_input_manager, config, code_context, diff_context
+    )
     code_map = CodeMap(git_root, token_limit=2048) if not no_code_map else None
     if code_map is not None and code_map.ctags_disabled:
         ctags_disabled_message = f"""
--- a/mentat/app.py
+++ b/mentat/app.py
@@ -72,8 +72,8 @@ def run_cli():
     exclude_paths = args.exclude
     no_code_map = args.no_code_map
     history = args.history
-    commit = args.commit
-    branch = args.branch
+    commit = None if not args.commit else args.commit[0]
+    branch = None if not args.branch else args.branch[0]
     # Expanding paths as soon as possible because some shells such as zsh automatically
     # expand globs and we want to avoid differences in functionality between shells
     run(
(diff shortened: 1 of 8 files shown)


granawkins:
The plan is to add 3 new flags:
  - `--history/-H` (integer): show the diff going back N commits
  - `--commit/-c` (str): show the diff w/r/t some commit
  - `--branch/-b` (str): show the diff w/r/t some branch

I've had a couple realizations in the process:
1. Adding the diff can result in lots of duplicated lines. If we just append the diff, anything new in the current version is included in the system prompt twice. For this reason I took a new approach: Annotate changed lines with a '-' or '+' to represent the diff. I still need to benchmark performance of this vs just appending.
2. For the 3 cases of these new flags, I compare w/r/t the *active* code. I suggest the default case (none of those flags) would compare with the latest git commit (like it does now, but with the new annotations). 

Still needs some cleaning up and testing, but I think the approach is solid, and it is functional.

biobootloader:
> Adding the diff can result in lots of duplicated lines. If we just append the diff, anything new in the current version is included in the system prompt twice. So how, I annotate changed lines with a '-' or '+' to represent the diff. I still need to benchmark performance of this vs just appending.

This makes a lot of sense, my only concern is it could make it harder for the model to understand the current code (especially if there are lots of removed lines). But it does reduce tokens and put diffs in context, so clear benefits there. 

Benchmarking will be important, but might mostly be us trying it out. We only have one benchmark that has it use the appended diffs: `test_calculator_continue_change`, and that just adds a couple lines.

If it could be done cleanly, it might be nice to be able to swap diff formats with a flag. I.e. swap between the old format, "appended diffs" and your new one, "inline diffs". That'd be useful for benchmarking it now, but also might be useful in the future when we have more benchmarks and can do more rigorous testing, and may find that both formats are useful in different scenarios or better/worse for different models. I don't want to add more code to maintain though if you don't think it's worth it.

> For the 3 cases of these new flags, I compare w/r/t the active code. I suggest the default case (none of those flags) would compare with the latest git commit (like it does now, but with the new annotations).

Agreed

granawkins:
I ran some tests to see how the new (more compact) diff format affects performance. 

TL;DR - The shortened format beats the more verbose in 2/3 subjective tests, ties on the third. Will optimize for this (winning) approach next.

# The Comparison
I compare 4 different cases:
A. Annotate diff additions with a simple '+' at start of line, and insert removed with '-'. Default case.
B. Like [A], but duplicate the additions (add lines with '+' *after* code lines, not instead).  In diff_context.py, change line 173 to `_annotate_file_message_duplicate`
C. Add vanilla diff after code file, without unchanged lines. In diff_context.py set line 168 `append = True`.
D. Add vanilla diff after code file, with some unchanged lines. Changes in [C] plus in git_handler.py line 110 remove `”-U0”

I created 3 tasks, easy medium and hard, for an earlier diff on this commit. For each task I made a scoring schema, and ran the test 4 times. I compare the total prompt length and the sum of scores from all 4 runs.

```
git checkout 9c9e5785  #Check out latest branch
```

Question 1:
- Run with `mentat mentat/git_handler.py -H 1`
- **What was done in the previous commit?**
- Scoring (1 point for each)
  1. Removed `get_history_diff` function from git_handler.py
  2. Removed `get_commit_diff` function from git_handler.py
  3. Removed `get_branch_diff` function from git_handler.py
  4. Added a new function `get_diff_for_file` in. git_handler.py
  5. Updated error logging message in the `get_diff_for_file` function
  6. Removed unused code related to the removed functions

Question 2:
- Run with `mentat mentat/git_handler.py -H 1`
- **The get_diff_for_file function currently returns the diff with respect to the active code. Let’s add another argument to that function so you can choose which commit or branch to diff against. Allow a default case (if not specified) that compares against current, like it has been.**
- Scoring (1 point for each):
  1. Add a new arg to get_diff_for_file with reasonable name, instead of overwriting
  2. Conditionally Add it to the subprocess command
  3. Add it correctly and don’t change anything else not needed

Question 3:
- Run with `mentat mentat/git_handler.py mentat/diff_context.py -H 1`
- **Add a new function in git_handler.py, get_diff_files. It should return a list of relative paths to files affected by the specified diff - a git_root, target and (optionally) against. Then, in diff_context.py, replace the local version of this with the new one - don't forget to import it.**
- Scoring (1 point for each):
  1. Add the new func to git_handler properly
  2. Include try block
  3. Import
  4. Import without duplicating
  5. Use the new func on line 145 instead of making _get.. a wrapper.
  6. Insert it correctly

# Results
| Case | Q1 Tokens | Q1 Score | Q2 Tokens | Q2 Score | Q3 Tokens | Q3 Score|
|------|-------------|-------|--|--|--|--|
|A|3343|20|3391|4|5645|14
|B|3348|20|3396|3|6259|9
|C|3549|15|3597|2|7145|8
|D|3542|20|3590|3|7138|3

biobootloader:
Really nice work! I feel a lot more comfortable changing the format with these thorough tests you've done. Looks like your intuition was correct on the format!

Thank you for the clear writeup. It's amazing what a big difference the format makes in the scores, especially tasks 2 and 3.

It'd be great to add a benchmark as part of this PR that has a detailed diff for Mentat to deal with. Perhaps you can adapt one of your tasks? The tricky part will be automatically scoring it. A pass/fail is fine if you just try to run the function it creates, like we do with other benchmarks. Try to get something that currently passes 25-75% of the time.

biobootloader:
A big use case I think this will be awesome for is understanding PRs. But what's the best way to run Mentat on a PR? Like say you checkout a PR branch - if you run `mentat --branch main` you'd get the diff against main, but that's going to include changes from commits added to main, not just changes from this branch.

What we want is the changes just from this branch, relative to the last commit on main that is an ancestor of this branch. We can get that commit with `git merge-base main HEAD` - should we add another option to run this that does that automatically? So it'd also take a `branch` as user input, but then it would run `git merge-base {branch} HEAD` to find the common ancestor with that branch, then use that commit as if the user had run `mentat --commit abc`

It's a lot of user options to have, but this seems like a core use case - we can reduce options later when we find out what users actually use

granawkins:
I see - basically as shortcut for `mentat --commit $(git merge-base main HEAD)`, right? How bout we just call the command --merge-base so you can run `mentat --merge-base`.

Another option is collapsing everything into a single arg called `--diff`:
- `HEAD` (active changes) by default
- If a string, match to commit or branch
- If int, same as 'history': `HEAD~{i}`
- Maybe we have a special case, 'merge-base', which does the above?

This also enforces only passing a single diff argument. Thoughts?

biobootloader:
> basically as shortcut for mentat --commit $(git merge-base main HEAD), right?

yes, except should we let people choose the branch, so it's not just main?

I really like the idea of collapsing everything to a single `--diff` arg. But I'm not sure how to support `merge-base` along with specifying a branch though...

biobootloader:
Maybe the two args would be `--diff` (taking branch, commit, or int) and `--pr-diff` (taking a branch, and running merge-base on it?)

biobootloader:
If there a way to let git process the input for us? So we don't have to actually handle cases of branch / commit / int? Maybe instead of an int they would use `HEAD~3`, just like with git. Like we'd just pass the user input straight to `git diff`...

biobootloader:
Ok so the technical term is "tree-ish": https://git-scm.com/docs/gitglossary#Documentation/gitglossary.txt-aiddeftree-ishatree-ishalsotreeish

So my last comment was saying maybe Mentat should take "tree-ish" arguments and let git itself deference them. That would give users flexibility, being able to do things like `HEAD~3` or `main^2` or whatever

granawkins:
Ya I like that a lot - pass-through straight to git diff, and forget about taking integers.

How about this interface:
```
mentat --diff <branch/commit>
mentat --diff --base <branch/commit>
```

biobootloader:
How about

```
mentat --diff <branch/commit>
mentat --diff-base <branch/commit>

or

mentat --pr-diff <branch/commit>

```

biobootloader:
Another thought - though this should probably be in a follow up PR:

For the "PR review" use case, one unfortunate thing that can happen is: a PR touches 10 files, and those files can't all be added to context. So the user can't ask Mentat questions about what the PR changed.

The solution would be a fallback system: if not all changed files fit, then we switch modes and don't include the entire files, we just include the actual diff and some surrounding context / the code map.

biobootloader (approved):
🚀 

Sorry for the delay reviewing!

jakethekoenig:
Ah I'm too slow. Great work @granawkins! Mentat likes it:
```
(.venv) Development/mentat % mentat --pr-diff main mentat/diff_context.py mentat/git_handler.py
Files included in context:
mentat
└── mentat
    ├── diff_context.py
    └── git_handler.py

Diff annotations:
 ─•─ Merge-base Branch main: c2fb2dc3: Merge pull request #84 from waydegg/add-pyright | 8 files | 422 lines

File and prompt token count: 4767 / 8192
Type 'q' or use Ctrl-C to quit at any time.

What can I do for you?
>>> I'd like some help reviewing this pull request. Can you describe its function and point
out anything you think could be improved?

Including CodeMap (filepaths only)

Total token count: 5147

streaming...  use control-c to interrupt the model at any point

The pull request adds a new feature for handling and displaying the context of diffs in a git repository. It introduces a new module `diff_context.py` and makes changes to `git_handler.py`.

Key functionalities added:
1. A new class `DiffAnnotation` to represent the diff annotations.
2. Functions to parse the diff and generate a list of annotations.
3. Functions to annotate the file message with the diff annotations.
4. A new class `DiffContext` that holds the configuration and target for the diff context.
5. Functions to display the diff context and annotate the file message.
6. Functions to get the type of treeish (commit, branch, relative) and to get the diff context based on provided arguments.
7. New functions in `git_handler.py` to get the diff for a file, get the metadata for a treeish, get the files in a diff, check if the HEAD exists, and get the default branch.

Possible improvements:
1. Add more comments and docstrings to explain the code and its purpose.
2. Add tests for the new functions and classes to ensure they work correctly.
3. Make sure the code follows the project's coding style and conventions.
```

PR #88, opened 2023-09-11
add /add and /remove commands

waydegg:
This PR adds commands for adding and removing files/partial files from the code context.

What isn't covered here that I'm including in additional PRs:
- update code tree w/ out-of-repo files
- add commands inputs to history completions
- fix completions for /add command to show files
- add command to print code map
- add command to list context files
- handle going over context with added files
- add (fuzzy) completions for /remove command
- merge/remove partial files

### DIFF
--- a/mentat/commands.py
+++ b/mentat/commands.py
@@ -3,6 +3,7 @@ from __future__ import annotations
 from abc import ABC, abstractmethod
 from typing import List
 
+from ipdb import set_trace
 from termcolor import colored, cprint
 
 from .code_context import CodeContext
@@ -120,6 +121,9 @@ class CommitCommand(Command, command_name="commit"):
 
 class AddCommand(Command, command_name="add"):
     def apply(self, *args: str, code_context: CodeContext) -> None:
+        if len(args) == 0:
+            cprint("No files specified\n", "yellow")
+            return
         for file_path in args:
             code_file = CodeFile(file_path)
             code_context.add_file(code_file)
@@ -135,6 +139,9 @@ class AddCommand(Command, command_name="add"):
 
 class RemoveCommand(Command, command_name="remove"):
     def apply(self, *args: str, code_context: CodeContext) -> None:
+        if len(args) == 0:
+            cprint("No files specified\n", "yellow")
+            return
         for file_path in args:
             code_file = CodeFile(file_path)
             code_context.remove_file(code_file)
--- a/mentat/commands.py
+++ b/mentat/commands.py
@@ -1,7 +1,7 @@
 from __future__ import annotations
 
 from abc import ABC, abstractmethod
-from typing import List
+from typing import List, Optional
 
 from termcolor import colored, cprint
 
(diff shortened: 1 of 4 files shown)


mentat/mentat_prompt_session.py
```diff
@@ -200,7 +202,10 @@ def prompt(self, *args, **kwargs):
         while (user_input := super().prompt(*args, **kwargs)).startswith("/"):
             arguments = shlex.split(user_input[1:])
             command = Command.create_command(arguments[0])
-            command.apply(*arguments[1:])
+            if isinstance(command, (AddCommand, RemoveCommand)):
+                command.apply(*arguments[1:], code_context=self.code_context)
```
PCSwingle, line 206: i think this is a perfect spot to use **kwargs; we can always pass in code_context, no matter what the command type, and add and remove can just access it with kwargs["code_context"]; let me know what you think, but I think this is cleaner and makes it a lot easier in the future to add other possible singletons that are needed.

mentat/commands.py
```diff
@@ -114,3 +116,33 @@ def argument_names(cls) -> list[str]:
...
+    def argument_names(cls) -> list[str]:
+        return ["file1", "file2", "..."]
+
+    @classmethod
+    def help_message(cls) -> str:
+        return "Add files to the code context"
+
+
+class RemoveCommand(Command, command_name="remove"):
+    def apply(self, *args: str, code_context: CodeContext) -> None:
```
PCSwingle, line 137: That would mean adding **kwargs here and in every apply function

PCSwingle:
Looks great! Thanks for adding this! Only 2 suggestions: the kwargs suggestion in the comments, and adding a test for each command. :rocket:

mentat/mentat_prompt_session.py
waydegg: I actually had done this initially, but opted to have `code_context` specified directly for commands that need it. Purely a personal preference, but my thinking is it's clearer what's going on when we explicitly define method arguments.

waydegg:
> adding a test for each command

Will do! There's not much happening atm so the tests won't be super useful atm.

mentat/mentat_prompt_session.py
PCSwingle: Hmm... I like **kwargs a lot more since it means we don't need this isinstance here (I think the actual implementation of the commands should be only in commands file, and it shouldn't have to bleed out into other files like this). @biobootloader, what do you think?

waydegg: Ah that makes sense. In that case I think either choice is good as long as we're consistent :)

biobootloader: I agree that `**kwargs` obscures what's going on, so I usually don't like it, but also it's a bit awkward to have to check. I think I have an idea to simplify let me try it out and commit here

biobootloader: Ok I changed it so we pass `code_context` to `create_command`. This way we don't need `**kwargs` and don't need to check command type outside of the commands file. Commit here: https://github.com/biobootloader/mentat/pull/88/commits/61c10536dd996c7d0398e3216961d811c51545d6

mentat/commands.py
```diff
@@ -19,11 +19,21 @@ def __init_subclass__(cls, command_name: str | None) -> None:
...
     @classmethod
-    def create_command(cls, command_name: str) -> Command:
+    def create_command(
+        cls, command_name: str, code_context: Optional[CodeContext] = None
+    ) -> Command:
         if command_name not in cls._registered_commands:
             return InvalidCommand(command_name)
+
+        command_cls = cls._registered_commands[command_name]
+        if command_cls in [AddCommand, RemoveCommand]:
```
PCSwingle, line 30: Hmm, at this point why not just have code_context be an instance variable of the Command class so every command has access to it?

biobootloader: Eventually we'll come up with a cleaner way of sharing global singletons like CodeContext. First we need to see how the interface refactor goes which will inform whether we might eventually have multiple Mentat conversations running on the same process or not.

Until then I think this is fine, it makes it easier to read `AddCommand` and `RemoveCommand` and understand what they have access to.

biobootloader (approved):
Great stuff @waydegg! Feel free to merge if you're happy with my change.

Also if you could add this stuff to an issue somewhere (maybe you already did?):

> What isn't covered here that I'm including in additional PRs:
>update code tree w/ out-of-repo files
>add commands inputs to history completions
>fix completions for /add command to show files
>add command to print code map
>add command to list context files
>handle going over context with added files
>add (fuzzy) completions for /remove command
>merge/remove partial files

PR #90, opened 2023-09-12
refactor benchmarks

PCSwingle:
Refactors the benchmarks to make it easier to make new benchmarks; I ran every benchmark a few times to confirm that they all can still pass.


biobootloader (approved):
Nice!

PR #92, opened 2023-09-12
Exercism Benchmark

jakethekoenig:
Run:
```
pytest -s testbed/benchmarks/exercism_pratice_python.py
```

Edit num_to_run to test on a subset. There are a few TODOs I need to clean up before it's done.

You can see the generated code [here](https://github.com/exercism/python/compare/main...jakethekoenig:exercism-python:Mentat-solutions-first-try). You can run those tests with the following command:
```
pytest --continue-on-collection-errors exercises/practice/
```

Results: 81/134 pass, 2 hung, 14 couldn't be ran and 27 failed for an overall pass rate of 60%


testbed/benchmarks/exercism_pratice_python.py
```diff
@@ -0,0 +1,78 @@
...
+            if results == 0:
+                passed += 1
+            else:
+                failed += 1
+        else:
+            print(f"Could not find test for {problem_dir}")
+        num_to_run -= 1
+
+    # TODO: I'm not sure it makes sense to run with pytest because it's not so much
+    # a pass fail script as something to collect a variety of statistics. I may refactor into a scipt.
```
biobootloader, line 76: yeah, I'd say just use pytest if it's going to use shared fixtures useful for other tests / benchmarks, otherwise makes sense as a script

jakethekoenig:
Sorry for the flurry of commits. I should have been adding `--preview` to my local linting.

jakethekoenig:
I'm running it with retries now and will report its performance when it's done. I think I'm mostly done though I'll probably add some documentation. Also it's still not correctly excluding the tests I think because I'm not running from git root. So I'll fix that too.

tests/benchmarks/README.md
```diff
@@ -0,0 +1,28 @@
...
+
+## Writing Benchmarks
+
+The following is a list of some best practices for writing benchmarks
+
+- You should not automatically tear down github repos you clone and modify so people running the test can inspect code output in detail.
+
+In [the exercism benchmark](./exercism_practice_python.py) you can see examples of the following things you might want for a benchmark:
+- How to simulate an interaction with Mentat in the function `mock_user_input_manager`.
+- How to clone a github repo in `clone_exercism_python_repo`.
```
biobootloader, line 28: This is great, thanks for taking the time to write it up clearly!

conftest.py
```diff
@@ -0,0 +1,26 @@
...
+        "--max_workers",
+        action="store",
+        default="1",
+        help="Number of workers to use for multiprocessing",
+    )
+    parser.addoption(
+        "--start_at",
+        action="store",
+        default="0",
+        help="Number of workers to use for multiprocessing",
```
biobootloader, line 25: which exercise to start at? maybe a more general way to say since this would be used for other benchmark sets as well

tests/benchmarks/exercism_practice_python.py
```diff
@@ -0,0 +1,152 @@
...
+
+def test_practice_directory_performance(
+    mock_user_input_manager,
+    clone_exercism_python_repo,
+    num_exercises,
+    max_iterations,
+    max_workers,
+    start_at,
+):
+    exercises = os.listdir("exercises/practice")[start_at:num_exercises]
```
biobootloader, line 133: should it be `[start_at:start_at + num_exercises]`?

tests/benchmarks/exercism_practice_python.py
```diff
@@ -0,0 +1,152 @@
...
+    threadLocal.test_output_file = f"{threadLocal.exercise}/test_output.txt"
+    threadLocal.iterations = 0
+    problem_file = problem_dir.replace("-", "_")
+    run(
+        [f"{threadLocal.exercise}/{problem_file}.py"],
+        no_code_map=True,
+    )
+    passed = exercise_passed()
+    sys.__stdout__.write(
+        f"\nFinished {problem_dir} in {threadLocal.iterations} iterations {passed}"
```
biobootloader, line 116: It would also be good to display the cost here if we can get it from Mentat cleanly

```suggestion
        f"\nFinished {problem_dir} in {threadLocal.iterations} iterations. Passed={passed}"
```

tests/benchmarks/exercism_practice_python.py
```diff
@@ -0,0 +1,152 @@
...
+def test_practice_directory_performance(
+    mock_user_input_manager,
+    clone_exercism_python_repo,
+    num_exercises,
+    max_iterations,
+    max_workers,
+    start_at,
+):
+    exercises = os.listdir("exercises/practice")[start_at:num_exercises]
+    num_exercises = len(exercises)
```
biobootloader, line 134: would be better to not overwrite the input `num_exercises`. And lets print how many exercises are actually being run (since due to start_at or the actual number of exercises available it might not match the input `num_exercises`

tests/benchmarks/exercism_practice_python.py
```diff
@@ -0,0 +1,152 @@
...
+        UserInputManager, "collect_user_input", new=mocked_collect_user_input
+    )
+
+
+@pytest.fixture
+def clone_exercism_python_repo(start_at):
+    exercism_url = "https://github.com/exercism/python.git"
+
+    local_dir = f"{os.path.dirname(__file__)}/../../../exercism-python"
+    if start_at != 0:
```
biobootloader, line 74: a bug here? if `start_at=0` the repo won't get cloned / reset

biobootloader (commented):
Really nice! 🚀 

Did you ever get any OpenAI rate limit errors?

It'd be nice if while running this it kept you informed on the tokens/minute and requests/minute being made to OpenAI, as well as the cost. But Mentat doesn't expose those cleanly, maybe we can add that as a follow up issue.

conftest.py
jakethekoenig: I added this flag because a few times the benchmark would fail half way through with a spurious error and I needed an easy way to restart. Since then I added a try catch block around run so I haven't run into that problem as much and I think maybe I'll replace this flag with a `--continue` flag which when set the test won't git reset and directories that already have test_output.txt will be skipped.

jakethekoenig:
> Really nice! 🚀
> 
> Did you ever get any OpenAI rate limit errors?
> 
> It'd be nice if while running this it kept you informed on the tokens/minute and requests/minute being made to OpenAI, as well as the cost. But Mentat doesn't expose those cleanly, maybe we can add that as a follow up issue.

Yeah I did a bunch. I think it can run to completion with 2 workers but all the actual complete tests I ran with 1 worker because I didn't want to have to start over on a failure.

I agree it would be good to have running cost/token information. I wonder if instead of calling `run` it'd be better to call `get_model_response` directly like [here](https://github.com/biobootloader/mentat/blob/ffbdc52ba59eca1797530f04a3cf86f3661db143/mentat/app.py#L141) and then we could consult the `cost_tracker`.

jakethekoenig:
Thinking about your comments I changed up the flags a bit in a way that I think will be more intuitive easier for other people to use. I quickly added start_at when a test failed half way through and I wanted to restart it. But actually it makes more sense to have the default behavior not tear down the git repo state and just skip tests that already run. Also I surrounded the mentat run with a try catch so hopefully this kind of utility won't be necessary.

I agree improving the mid test logging to show a current costs/tokens/passes/time elapsed would be really nice but I'm not sure it's the most valuable thing to work on. I guess it's nice to have it in place so future benchmarks can copy it.

biobootloader:
> I agree improving the mid test logging to show a current costs/tokens/passes/time elapsed would be really nice but I'm not sure it's the most valuable thing to work on.

agreed, we don't need it for now, this is great already! After the interface refactor we'll figure out the right way to access the cost tracker / count tokens. Then we could also auto-adjust workers to stay under rate limits too

tests/benchmarks/exercism_practice_python.py
```diff
@@ -0,0 +1,174 @@
...
+            lines = f.readlines()
+            return "failed" not in lines[-1] and "passed" in lines[-1]
+    except FileNotFoundError:
+        return False
+
+
+def get_error_message():
+    with open(threadLocal.test_output_file, "r") as f:
+        lines = f.readlines()
+        lines = lines[-30:]
```
biobootloader, line 29: I saw Aider sends "the first 50 lines of test error". It makes more sense to me to send the last lines like you're doing here. Is 30 the most that fits in context sometimes?

biobootloader (approved):
🚀

Issue #91, opened 2023-09-12
Refactor CodeChange code


PCSwingle:
Refactoring CodeChange into a concrete and abstract format; the concrete format would manage parsing the model's output and would be able to convert into an abstract change. The abstract change would use an internal format and would handle actually changing files. This would allow us to easily create new formats and test them out, without having to change anything outside of the parsing (and maybe streaming?) logic. After talking with @biobootloader here is the current plan:

### Overview

Concrete Change: Contains logic for parsing different model formats; can be converted to an abstract change. Can also be converted from an abstract change so that we can automatically generate examples for different models (and possibly tests) from a single example set.

Additions/Deletions: Make up an Abstract Change. These will be the simplest possible changes, either adding a block of code at a certain line or deleting a certain block of code. Ideally any conflicts will be resolved by the concrete change parser, but even if they aren't the Additions and Deletions will be processed in order and so conflicts shouldn't be an issue.

Abstract Change: Will be stored per file; each file will contain a list of Additions and Deletions; additionally, each Addition and Deletion will be marked with a specific Concrete Change; this will allow the user to use interactive mode to accept or reject (and eventually maybe ask the model to change?) specific changes that it outputs.

### Plan

1. Create Abstract Changes, Additions/Deletions, and modify CodeFileManager to use them when applying changes
2. Create tests for Abstract Changes
3. Create Concrete Change basic interface
4. Look into seeing if possible to have unified streaming logic for different Concrete Change formats; if so, create it
5. Migrate current format parsing and streaming over to Concrete Change
6. Test current format Concrete Change
7. Create two-way conversion between Abstract Change and Concrete Change
8. Add testing for conversion
9. Add un-parsing for current format Concrete Change -- this will be used for tests and prompt examples
10. Create Abstract Change examples and tests, and use automatic conversion to Concrete Change all the way to pre-parsed text for both prompt examples and tests

### Possible alternative?

Another design choice we could have made is to have Abstract Changes represent an entire file, and simply be the text of the file after applying all changes rather than having Additions and Deletions; this would resolve some problems with conflicts and renaming, but wouldn't let us be able to reference specific Concrete Changes from the Abstract Changes, meaning we would mostly likely end up simply using hunks (groups of changed lines) in interactive mode. For now, @biobootloader and I have decided to go with the previous plan to allow interactive mode to be more powerful.

jakethekoenig:
[...]

PR #94, opened 2023-09-13
Refactor changes

PCSwingle:
Refactor CodeChange into AbstractChange and ConcreteChange as outlined in #91 

Edit: After talking with @biobootloader, we've decided to change the structure quite a bit. This is now the general layout:

Model output -> Parser (this parser can be injected in, so any parser will work) -> new FileEdit format that groups by file and only has 'Replacement' subedits -> User filter (make these changes, keep these ones, don't use these ones, etc.) -> Conflict Resolution (handled by FileEdit; no conflict resoution needed on a per-parser basis). The final list of FileEdits can be converted into a final file's code lines, which is written to the file by the CodeFileManager.

Instead of completely changing the Parser and CodeChange files, I've extracted them into OriginalFormatParser and OriginalFormatChange classes; while we don't necessarily need intermediate CodeChanges for every parser, in this case it helps us use our previous parsing code instead of re-writing it all.

TODO:
- [x] Make Parser interface and make it easily injectable
- [x] Add prompts connected to Parser
- [x] Add tests for FileEdit, especially conflict resolution
- [x] Fix all TODO's scattered around the code
- [x] Restructure tests between format specific tests and non-format specific tests


mentat/code_changes/abstract/abstract_change.py
```diff
@@ -0,0 +1,225 @@
...
+    # Exclusive, 0 indexed
+    ending_line: int = attr.field()
+    owner: int = attr.field()
+
+
+@attr.s
+class Rename:
+    # None represents deleting this file
+    name: Path | None = attr.field()
+    owner: int = attr.field()
```
biobootloader, line 36: I don't love using `None` for deletion inside `Rename`. I might prefer explicit `DeleteFile`, `CreateFile`, `RenameFile` classes. Or maybe the name could be `FileUpdate` or something like that instead of `Rename`

PCSwingle: Hmm, looking back at it I think I agree that having DeleteFile CreateFile and RenameFile classes would be nicer and simpler; I'll switch over to that.

PCSwingle: Hmm, but thinking about it more I do like having just one class

mentat/parsers/block_parser.py
```diff
@@ -0,0 +1,28 @@
...
+from typing_extensions import override
+
+from mentat.code_file_manager import CodeFileManager
+from mentat.config_manager import ConfigManager
+from mentat.parsers.file_edit import FileEdit
+from mentat.parsers.original_format.original_format_parsing import (
+    stream_and_parse_llm_response,
+)
+from mentat.parsers.parser import Parser
+from mentat.prompts import block_parser_prompt
```
jakethekoenig, line 12: I'd define the string here instead of importing. Or do we plan to keep all the prompts in the same place? I guess I could see that being advantageous if some of us become more specialized "prompt engineers".

Similarly I don't see the value of importing `original_format_parsing` code instead of deleting that file and moving its code here.

mentat/parsers/block_parser.py
```diff
@@ -0,0 +1,28 @@
...
+
+
+class BlockParser(Parser):
```
jakethekoenig, line 15: It's called block because the changes are given in "blocks" like this?
```
@@start
{
    "file": "core/script.py",
    "action": "insert",
    "insert-after-line": 3,
    "insert-before-line": 4
}
@@code
    if name == "Bob":
        print("Nice to see you again!")
@@end
```

jakethekoenig (commented):
I really like the direction here. Thanks!

mentat/parsers/block_parser.py
PCSwingle: Hmm, I personally like having the prompt in a separate file; in fact, I would prefer to eventually just have it be a text file (instead of messing around with all of this python dedent stuff and strip for the prompt), and there would be a different text file for each prompt. Definitely open to suggestions on that too!

As for the original_format_parsing (and original_format_change) files, we could definitely move their code to this file; my reasoning for leaving it in those 2 files was that it's more like 'legacy' code; ideally, we don't have to touch that streaming/parsing code, and future streaming parsers we build around the current system will be cleaner (since they were built for this system). I definitely don't have anything against moving both files code to this file though!

mentat/parsers/block_parser.py
PCSwingle: Haha that is the reasoning; I don't like the name but I spent too long trying to think of a better name! Definitely let me know if you have any better ideas!

jakethekoenig: The name is fine. I was worried future formats could also be reasonably called "block" but thinking about it it seems like a lot of future formats we will try will be easier to name e.g. GitDiffParser, FullFileParser

mentat/conversation.py
```diff
@@ -1,31 +1,45 @@
...
+        parser: Parser,
         config: ConfigManager,
         cost_tracker: CostTracker,
         code_file_manager: CodeFileManager,
         code_map: CodeMap | None = None,
     ):
+        prompt = parser.get_system_prompt()
         self.messages = list[dict[str, str]]()
-        self.add_system_message(system_prompt)
+        self.add_system_message(prompt)
```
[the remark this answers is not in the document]
PCSwingle: :+1:

PCSwingle:
This PR is almost ready to merge; I'm going to wait on #101 to merge and then I'll resolve conflicts.

mentat/parsers/block_parser.py
jakethekoenig: I could definitely see putting prompts in plain text. Though the ergonomics of importing a string are maybe a little better than reading a file.

I'm fine with leaving it separated and not cleaning it up to fit more organically in the current system. I imagine after a period of experimentation we'll settle on just one parser that we like best and delete the others from the code base. I could be wrong though. Possibly different models/languages or types of requests will work better with certain formats.

PCSwingle:
> > This PR is almost ready to merge; I'm going to wait on #101 to merge and then I'll resolve conflicts.
> 
> Are the todos still in `file_edit` for a future PR?

The Replacement 'owner' TODO yes; the current parsing format has a one to one 'change' to Replacement ratio, but eventually we might have a format that creates 2 replacements that we want to represent as one 'change' that the user can filter. It isn't a big addition though and just complicates a few things, so I figured we might as well add that when we need it. As for the conflict resolution TODO's, I personally don't think user input for resolving them is super important, but if you or @biobootloader think we should get that done before merging this PR I'm 100% ok with doing it now.

pyrightconfig.json
```diff
@@ -1 +1 @@
-{"include": ["mentat"], "typeCheckingMode": "strict"}
+{"include": ["mentat"], "ignore": ["testbed", "tests"], "typeCheckingMode": "strict"}
```
granawkins, line 1: Nice. was wondering about this.

mentat/parsers/block_parser.py
waydegg, line 1: What's your thinking on which classes should be wrapped with `attrs`?

waydegg (changes requested):

waydegg (commented):
Left some comments!

mentat/parsers/block_parser.py
biobootloader: I like having the prompts in other files as well. Right now the imported `block_parser_prompt` is our entire original system prompt. Before adding a second output format we'll have to split that into parts: 1) a generic part that explains that the model is part of a code editing system, that it might help the user write code or answer questions, etc. and 2) the output format specific part that gives a description of the response format and examples.

When we make that split lets think about the file format and explore the options @waydegg suggested

biobootloader: > my reasoning for leaving it in those 2 files was that it's more like 'legacy' code; ideally, we don't have to touch that streaming/parsing code, and future streaming parsers we build around the current system will be cleaner

I'm ok leaving that code there since our intention is we find better output formats and then delete those files. I wouldn't want it hanging around as legacy code for very long though - if we find we still want this output format for something, we'd want to clean it up / integrate it more.

biobootloader:
Thanks @PCSwingle for this important and complicated refactor! And thanks everyone for the quick reviews 🚀 

I'll take a closer look later today after you've had a chance to address comments, let me know when you're ready for that!

I'm excited to try some new output formats!

mentat/code_file_manager.py
```diff
@@ -126,106 +121,66 @@ def _add_file(self, abs_path: Path):
...
+                        f"File '{rel_path}' changed while generating; current"
+                        " file changes will be erased. Continue?",
+                        color="light_yellow",
+                    )
+                    if not self.user_input_manager.ask_yes_no(default_yes=False):
+                        cprint(f"Not applying changes to file {rel_path}")
+            else:
+                stored_lines = []
+
+            if file_edit.rename_file_path is not None:
```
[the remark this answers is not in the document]
PCSwingle: Hmm, I'm not entirely sure about this because that stages the rename; I don't think we want mentat to stage anything unless the user specifically asks us to; and of course when the user eventually stages the changes git will recognize the rename either way. It is nice having git recognize it immediately though. What do you think?

mentat/conversation.py
```diff
@@ -1,28 +1,36 @@
...
+from timeit import default_timer
+
+from openai.error import InvalidRequestError, RateLimitError
 from termcolor import cprint
 
-from .code_change import CodeChange
-from .code_file_manager import CodeFileManager
-from .config_manager import ConfigManager, user_config_path
-from .llm_api import (
+from mentat.config_manager import ConfigManager, user_config_path
```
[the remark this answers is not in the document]
PCSwingle: Ah, good catch! We can talk about changing our lint settings to automatically do whatever we decide next time we meet.

mentat/parsers/file_edit.py
```diff
@@ -0,0 +1,169 @@
...
+from mentat.parsers.change_display_helper import (
+    DisplayInformation,
+    FileActionType,
+    get_full_change,
+)
+from mentat.user_input_manager import UserInputManager
+
+if TYPE_CHECKING:
+    # This normally will cause a circular import
+    from mentat.code_file_manager import CodeFileManager
```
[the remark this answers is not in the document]
PCSwingle: I agree, it would be good to clean up our code so it isn't so interconnected; we might be able to get it done after refactoring the interface and our singletons.

mentat/parsers/file_edit.py
```diff
@@ -0,0 +1,169 @@
...
+
+
+# TODO: Add 'owner' to Replacement so that interactive mode can accept/reject multiple replacements at once
```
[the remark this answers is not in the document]
PCSwingle: When we enter interactive mode we want to be able to veto (and in the future ask for fixes to) specific changes that the model made; right now, we just veto individual Replacements, but if the Replacements had an 'owner' which represents that multiple replacements were made by one vetoable change, we would group them together in interactive mode.

mentat/parsers/block_parser.py
PCSwingle: If it's a dataclass; i.e., meant to pretty much be a struct (although we can add functions them) so that attrs automatically gives us slots, eq, comparisons, and other stuff.

mentat/parsers/block_parser.py
PCSwingle: Hmm, I'm not sure I understand what the File means in this context? Most likely all of our formats will need the model to specify the file.

mentat/code_file_manager.py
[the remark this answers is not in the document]
biobootloader: hmm yeah we don't want to stage. I believe `git mv` forces git to recognize it as a rename (which it sometimes fails, if similarity isn't high enough). Let's leave as is for now so we don't stage. Perhaps in the future we could have Mentat "remember" a rename and so when it commits for the user it'll inform git

mentat/code_file_manager.py
```diff
@@ -126,106 +121,66 @@ def _add_file(self, abs_path: Path):
...
-                    self._add_file(abs_new_path)
-                    code_lines = self.file_lines[rel_path]
-                    with open(abs_new_path, "w") as f:
-                        f.write("\n".join(code_lines))
-                    self._delete_file(abs_path)
-                    file_changes[code_change.name] += file_changes[rel_path]
-                    file_changes[rel_path] = []
-                    self.file_lines[code_change.name] = self._read_file(
-                        code_change.name
+    # Mainly does checks on if file is in context, file exists, file is unchanged, etc.
```
biobootloader, line 133: We used to do these checks earlier, right? like when the change was created by parsing?

On the one hand, some these should be rare errors (creating a file that already exists, editing non-existent file). But editing a file not in context might be common, since we are starting to include parts of files in context. It's not ideal if Mentat crashes *after* the user has seen the changes and accepted them / spent time doing interactive selection or providing feedback.

PCSwingle: We still do do them in the current parser; but of course, we don't want to copy that code to every parser we make. I agree we should always do it in the parser though; what if we make it so that FileEdit does these checks on creation? That way it would be done at parsing for every parser. This also made me realize that we don't do any rename file checks here anymore; I'll fix that.

biobootloader: Yes doing it on FileEdit creation sounds great!

mentat/parsers/block_parser.py
biobootloader: `OriginalFormatParser` would be unambiguous, but `BlockParser` seems fine for now.

The name really just matters if we end up keeping this format around, which we'd only do if it is the best for some models / scenarios. I don't expect that but if it does happen we'll find a name that distinguishes it well from the other formats

mentat/parsers/file_edit.py
```diff
@@ -0,0 +1,196 @@
...
+        for index, replacement in enumerate(self.replacements):
+            for other in self.replacements[index + 1 :]:
+                # TODO: another type of conflict (not caught here) would be both replacements being inserts on same line
+                if (
+                    other.ending_line > replacement.starting_line
+                    and other.starting_line < replacement.ending_line
+                ):
+                    # TODO: Ask user for conflict resolution
+                    other.ending_line = replacement.starting_line
+                    other.starting_line = min(other.starting_line, other.ending_line)
```
biobootloader, line 179: Let's print something here so the user knows what happened. Perhaps just say "these two changes conflicted, they've been auto-merged back to back, double check: {the changes back to back}". From the user's perspective they may have seen each change individually and thought everything looked good.

mentat/parsers/file_edit.py
```diff
@@ -0,0 +1,196 @@
...
+            or self.is_deletion
+            or (self.rename_file_path is not None)
+            or len(self.replacements) > 0
+        )
+
+    def resolve_conflicts(self, user_input_manager: UserInputManager):
+        self.replacements.sort(reverse=True)
+        for index, replacement in enumerate(self.replacements):
+            for other in self.replacements[index + 1 :]:
+                # TODO: another type of conflict (not caught here) would be both replacements being inserts on same line
```
biobootloader, line 172: so what happens in that scenario now? does Mentat crash? or does it just randomly put one first?

PCSwingle: Right now it just puts the insertion first in the list first and the next insertion last

mentat/parsers/file_edit.py
```diff
@@ -0,0 +1,169 @@
...
+                if (
+                    other.ending_line > replacement.starting_line
+                    and other.starting_line < replacement.ending_line
+                ):
+                    # TODO: Ask user for conflict resolution
+                    other.ending_line = replacement.starting_line
+                    other.starting_line = min(other.starting_line, other.ending_line)
+
+    def get_file_lines(self, file_lines: list[str]):
```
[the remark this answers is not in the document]
biobootloader: `get_updated_file_lines` ?

mentat/parsers/file_edit.py
biobootloader: how about we treat that scenario the same as the other conflict scenario then, and alert / display to the user what's happening

mentat/parsers/block_parser.py
waydegg: > Hmm, I'm not sure I understand what the File means in this context?

"File" would mean a literal file on the file system. Technically GitDiffParser would be pulling from a file too :/

biobootloader (approved):
Nice stuff! Feel free to merge after addressing comments

Issue #93, opened 2023-09-13
Add git subdirectory tests


PCSwingle:
Add tests for running mentat from a subdirectory of the git repository (we might have them already but I don't think we do; would be good to check first!).

granawkins:
I can do this quick but just to clarify, what's the expected behavior exactly?

e.g. in Mentat, I navigate to `mentat/mentat`, then run `mentat code_file.py`, and it runs with the top-level `mentat` as the git root. (this needs one fix to work, `code_file_manager.py` line 50)

Are you meant to also specify paths in sibling directories? e.g. `mentat code_change.py ../tests/code_change_test.py`?

PCSwingle:
Sorry I wasn't clear! I'm pretty sure it works as expected right now (I could be wrong); if you run it from within a subdirectory of a git repo, it correctly picks up the git root, and when you specify files when starting mentat it correctly picks the files relative to your cwd; all this issue is for is adding tests to make sure these things work correctly when the cwd is a subdirectory of the git repo. Hopefully that makes more sense, let me know if you have any other questions!

granawkins:
Resolved by #96

PR #96, opened 2023-09-14
Test run in subdirectory

granawkins:
Addresses #93, test that you can run mentat from a subdirectory of your `git_root`, and pass in paths relative to your `pwd`:
- One small fix in `code_file_manager` to make it work as described
- One new test in `code_file_manager_test` to confirm it works

### DIFF
--- a/tests/code_file_manager_test.py
+++ b/tests/code_file_manager_test.py
@@ -36,12 +36,16 @@ def test_partial_files(mock_config):
     file_path = os.path.join(dir_name, file_name)
     os.makedirs(dir_name, exist_ok=True)
     with open(file_path, "w") as file_file:
-        file_file.write(dedent("""\
+        file_file.write(
+            dedent(
+                """\
              I am a file
              with 5 lines
              third
              fourth
-             fifth"""))
+             fifth"""
+            )
+        )
     file_path_partial = file_path + ":1,3-5"
 
     code_context = CodeContext(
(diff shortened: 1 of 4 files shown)


tests/code_file_manager_test.py
```diff
@@ -66,3 +66,26 @@ def test_partial_files(mock_config):
              4:fourth
              5:fifth
               """)
+
+
+def test_run_from_subdirectory(mock_config, temp_testbed):
+    """Run mentat from a subdirectory of the git root"""
+    # Change to the subdirectory
+    run_from_dir = mock_config.git_root / "multifile_calculator"
```
PCSwingle, line 74: This can just be temp_testbed / "multifile_calculator"; temp_testbed just returns the path to the current temporary file. Additionally, I bet we could just run os.chdir("multifile_calculator") and avoid having to reference the testbed at all; feel free to do whichever you prefer!

PCSwingle:
Thanks for adding this test and finding this bug, Grant! I didn't even realize this was bugged right now. Looking into it more, I'm seeing a couple issues with our existing code that led to this bug in the first place; the CodeFile paths should always be relative to the git root, and a couple other things I noticed in the _read_file function; I think I'll go ahead and fix those right now, and when you review my changes feel free to edit them or merge everything

PCSwingle:
I changed CodeFile to always be created with an absolute path and always updated the test you created to make changes to files; this also brought to mind though that in a separate PR we should probably migrate CodeFileManager's file_lines to first of all not use paths relative to the git root and also maybe move the files lines to the CodeFile class? CodeFileManager would then be able to access the file lines through CodeContext. What do you think about that idea?

PCSwingle (approved):

PR #100, opened 2023-09-19
Model configuration

jakethekoenig:
Users can set which model to use. If it's not a GPT-4 model warnings are given. If it is a non-deprecated OpenAI model costs and tokens are computed. How to use an alternative API endpoint is documented.

### DIFF
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -43,12 +43,13 @@ A list of key-value pairs defining a custom [Pygment Style](https://pygments.org
 
 ### Maximum Context
 
-To save costs on larger models you can set a maximum context.
+If you're using a model other than gpt-3.5 or gpt-4 we won't be able to infer the model's context size so you need to manually set the maximum context like so. 
 ```json
 {
     "maximum-context": "16000"
 }
 ```
+This can also be used to save costs for instance if you want to use a maximum of 16k tokens when using gpt-4-32k.
 
 ## 🦙 Alternative Models
 
@@ -60,6 +61,7 @@ and then setting your model in `~/.mentat/.mentat_config.json`:
 ```json
 {
     "model": "/absolute/path/to/7B/llama-model.gguf"
+    "maximum-context": 2048
 }
 ```
-For non OpenAI models we may not be able to infer a maximum context size so to avoid overrun errors you should set that too.
+For models other than gpt-3.5 and gpt-4 we may not be able to infer a maximum context size so you'll also have to set the maximum-context.
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -4,11 +4,11 @@ To change Mentat's configuration, create `.mentat_config.json` in the git root o
 
 ## Options
 
-### Allow 32k
+### Use 32k
 Allow Mentat to use OpenAI's gpt-4 32k context window model. Your API key must already have access to the model.
-```
+```json
 {
-    "allow-32k": true
+    "model": "gpt-4-32k-0314"
 }
 ```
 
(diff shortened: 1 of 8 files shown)


jakethekoenig:
I know you said you didn't want users to use models that aren't good enough and have a bad experience but I think we should be able to easily set it to use gpt-3.5 especially as we experiment with edit formats just to experiment. I added warnings so hopefully people won't get too frustrated.

Some thoughts:
* I don't like having to pass the ConfigManager around. I wonder if we can import it as a singleton where we need it? I know @PCSwingle mentioned getting the git_root out of it which I think would be a good idea and help.
* It's unfortunate that OpenAi doesn't put `context_size` and `token_cost` in openai.Model.list(). I feel like it shouldn't be necessary for us to hard code them. Feels like all those numbers could change frequently.
* I linked the 2 issues that have to do with other models. Actually one could already use the other models by setting `OPENAI_API_BASE` but it now works with the .env file which it didn't before and is now documented.
* I removed some of the 32k specific logic and just have general warnings for getting close to context length.

biobootloader:
Great! I agree, let's let people use whatever models they want, as long as we warn them they are untested that's fine. And we can even start publishing benchmarks / performance of various models so people know what to expect. Some models might be good enough for asking questions about codebases but not editing them.

> I don't like having to pass the ConfigManager around. I wonder if we can import it as a singleton

agreed, once @waydegg's refactor is done we'll have more clarity on this

docs/configuration.md
```diff
@@ -41,3 +41,25 @@ A list of key-value pairs defining a custom [Pygment Style](https://pygments.org
 }
 ```
 
+### Maximum Context
+
+To save costs on larger models you can set a maximum context.
+```json
+{
+    "maximum-context": "16000"
```
biobootloader, line 49: Could it make sense to default to None, or 32k? I don't want someone to change the model to gpt-4-32k but then not realize or know why it's not using the full context. If someone runs a local model or something we can't infer context length for and this isn't set we can tell them it's required

jakethekoenig: Yeah, this is just an example for the doc. The default is `null`. There is code to set the `context_size` for gpt-3 and gpt-4 models. If the user runs a model for which we can't infer the context length and they don't set one then mentat fails with a pretty descriptive error (from openai not us) when they run out of context.

biobootloader: oops haha perfect!

biobootloader: > If the user runs a model for which we can't infer the context length and they don't set one then mentat fails with a pretty descriptive error (from openai not us) when they run out of context.

might be good to just require them to set it when we can't infer it, since our logic for filling up context (with code maps, embeddings, etc) will probably be based on a max. And it will save them a wasted call / get better warnings as they get close

biobootloader (approved):
Awesome! This has been a highly requested feature.

mentat/conversation.py
```diff
@@ -104,16 +145,16 @@ def get_model_response(self, config: ConfigManager) -> tuple[str, list[CodeChang
...
         state = run_async_stream_and_parse_llm_response(
-            messages, model, self.code_file_manager
+            messages, self.model, self.code_file_manager
         )
 
         self.cost_tracker.display_api_call_stats(
             num_prompt_tokens,
-            count_tokens(state.message),
-            model,
+            count_tokens(state.message, self.model),
```
PCSwingle, line 156: I could be wrong, but I think using any model other than GPT 3.5 or 4 will crash here; inside the cost tracker, we just access a dictionary with the model as the key. Since we don't know how much it costs for models other than GPT, I think we should disable the cost tracker altogether for other models.

PCSwingle: Wait, somehow I didn't notice you changed the CostTracker! Sorry, you can ignore this!

PCSwingle (approved):
Looks good to me! The whole model and cost code was a complete mess, so thanks for cleaning all of that up!

PR #102, opened 2023-09-20
Script to turn a log into a training example

jakethekoenig:
Logging is added to the conversation class so that the complete transcript from the LLMs perspective is in the log. A script that turns it into the format OpenAI expects for a training example is added.

### DIFF
--- a/mentat/conversation.py
+++ b/mentat/conversation.py
@@ -1,3 +1,5 @@
+import logging
+
 from termcolor import cprint
 
 from .code_change import CodeChange
@@ -85,12 +87,15 @@ class Conversation:
             )
 
     def add_system_message(self, message: str):
+        logging.debug(f"System Message:\n{message}")
         self.messages.append({"role": "system", "content": message})
 
     def add_user_message(self, message: str):
+        logging.debug(f"User Message:\n{message}")
         self.messages.append({"role": "user", "content": message})
 
     def add_assistant_message(self, message: str):
+        logging.debug(f"Assistant Message:\n{message}")
         self.messages.append({"role": "assistant", "content": message})
 
     def get_model_response(self, config: ConfigManager) -> tuple[str, list[CodeChange]]:
--- a/mentat/conversation.py
+++ b/mentat/conversation.py
@@ -148,6 +148,7 @@ class Conversation:
                 cprint_message = "\n".join(cprint_message)
                 cprint(cprint_message, color="yellow")
 
+        logging.debug(f"Code Message:\n{system_message}")
         messages.append({"role": "system", "content": system_message})
 
         num_prompt_tokens = get_prompt_token_count(messages, self.model)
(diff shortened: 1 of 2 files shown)


jakethekoenig:
This is a tool to make it easy to make fine tuning training data. You can run it with no arguments to turn your last mentat run into a training example or pass in the log file you want to turn into a training example.

Notes:
I think in some ways it'd be more principled to log the entire conversation sent to the LLM at each step but I didn't want our logs to grow quadratically. For now it's fine but I could imagine future systems where we let the LLM "explore" by running different commands seeing different files etc. and then truncating some of that conversation intelligently.

There's also a little bit of redundancy: User Input and User Message are both logged and mostly the same but different for the LLM.
LLM response and Assistant Message are both logged and the same. I like the parallelism of logging everything in conversation so I'd prefer to delete LLM reponse's log. I didn't want to create a merge conflict with the parsing refactor though so haven't yet.

PCSwingle (approved):
This is super nice!

biobootloader:
This is clever!

One tricky thing is it's not logging the "code message" where we give the model the context (code files, diffs, code maps, etc). The reason is that that message isn't added to the conversations `messages` - instead it's always appended temporarily right before making an API call:

https://github.com/biobootloader/mentat/blob/14ec691dba3943e2a69231441b513357bc97f77d/mentat/conversation.py#L93-L97

We only append it temporarily since we don't want the context being sent in the conversation history multiple times.

jakethekoenig:
> This is clever!
> 
> One tricky thing is it's not logging the "code message" where we give the model the context (code files, diffs, code maps, etc). The reason is that that message isn't added to the conversations `messages` - instead it's always appended temporarily right before making an API call:
> 
> https://github.com/biobootloader/mentat/blob/14ec691dba3943e2a69231441b513357bc97f77d/mentat/conversation.py#L93-L97
> 
> We only append it temporarily since we don't want the context being sent in the conversation history multiple times.

Ah thanks. I'd assumed it would have been set once in the beginning but I suppose it can change during a mentat run if people add, remove or edit files. I log it every time now and in my script only add the first one to the training run.

biobootloader:
> I suppose it can change during a mentat run if people add, remove or edit files

Yeah, if Mentat edits files too. And it could change more if we build context differently for new queries, with embedding lookups or not having as much space for the code map.

> I log it every time now and in my script only add the first one to the training run.

I guess each conversation should really create multiple training examples, one for each time the model responds with edits. What do you think?

jakethekoenig:
> I guess each conversation should really create multiple training examples, one for each time the model responds with edits. What do you think?

Yeah that's probably best. I'm a little unsure how OpenAI is chopping up training examples as it is. If a conversation goes User, Assistant, User, Assistant I guess that's already two training examples?

biobootloader:
> I'm a little unsure how OpenAI is chopping up training examples as it is. If a conversation goes User, Assistant, User, Assistant I guess that's already two training examples?

That's a good question. My assumption would be it's a single example with just the last message but I don't know. That would give more flexibility, like showing it how to recover from a poor first response, without actually training for the poor first response

PR #103, opened 2023-09-21
Tree-Sitter for parsing codefiles

granawkins:
We currently use Universal ctags to extract function data from code files, but @waydegg's also suggested using Tree-Sitter as it gives you richer data. I have come to agree, and put together this demo:
- For each language you want to support, you have to first download the "grammar" for it from their GitHub (~100 MB)
- Then you build a Language binary, which is a `ts-lang.so` file (~1.5MB). I've automated this and the above step so you can just run `python3 generate_tree_sitter.py` and to download grammars and compile.
- Load the binary into a `tree_sitter.Parser` to parse a file into a `tree_sitter.Node`, the root of your syntax tree.
- The syntax tree depends on the grammar, and somewhat by language, so we may need language-specific helper funcs to parse it into our usable format.

This is here for discussion for now, and I'll connect it up after I finish the code_context refactoring.


biobootloader:
> The syntax tree depends on the grammar, and somewhat by language, so we may need language-specific helper funcs to parse it into our usable format.

How similar are these syntax trees? I could imagine they are pretty different at a low level (within lines of code) but hopefully only have minor differences at the higher levels that we care more about?

bitnom:
> > The syntax tree depends on the grammar, and somewhat by language, so we may need language-specific helper funcs to parse it into our usable format.
> 
> How similar are these syntax trees? I could imagine they are pretty different at a low level (within lines of code) but hopefully only have minor differences at the higher levels that we care more about?

from what I remember tree sitter has like 99999 different parameters it extracts, and most the work is figuring out which ones you want. ctags is just easier to get started with.

granawkins:
This is built on in several other places and not likely to be useful anymore.

PR #104, opened 2023-09-21
remove logging from test environment

PCSwingle:
Removes logs from test environment, but not from benchmarks. We could also put benchmark logs in a subdirectory "benchmark_logs" instead of having them all in one directory; what do you think?

### DIFF
--- a/mentat/llm_api.py
+++ b/mentat/llm_api.py
@@ -39,14 +39,19 @@ def setup_api_key():
         raise UserError(f"OpenAI gave an Authentication Error:\n{e}")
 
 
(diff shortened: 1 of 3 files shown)


jakethekoenig (approved):
I realized I didn't mark my benchmark as such so with this change it wouldn't get run if the `--benchmark` flag wasn't set so I pushed a commit for that. It still won't get collected automatically because it doesn't start with `test_`. Thinking more about it it's possible it should be collected automatically just with a default of only running on one exercism test instead of all of them. But in any case that's not relevant for this PR.

PR #105, opened 2023-09-22
refactor prompts to be in separate txt files, add replacement parser

PCSwingle:
Create a replacement parser; this will a. give us a (hopefully better!) format to try, and b. let us see what logic between parsers/streamers can be abstracted.

TODO:

- [x] Abstract most of streaming between blockparser and replacementparser to make it much easier to create new parsers
- [x] Create tests for replacementparser
- [x] Create model error handling and error tests for replacementparser
- [x] Resolve added TODO's


PCSwingle:
Almost completely done! Only things left to do are add error handling for the new replacement parser, add tests for the replacement parser, and thoroughly QA everything. Below I've listed both the changes in actual functionality and a couple of things I don't like about my refactoring that maybe somebody else has some thoughts on.

Changes to functionality: 
If the model messes up a change in the middle of it's response (messing up json or not giving line numbers for example), instead of only throwing out that one and keeping the ones both before and after it, we now only keep the changes before that one. I've never seen the model make a mistake like that before, so I think it's all ok, and would be too much work to keep this in.

We no longer print the entire contents of the file when the model deletes it; I can easily undo this, but I think this looks a lot cleaner. Let me know what you think!

Things about my code refactoring that I don't like:
The way I pass information to _add_code_block; since the subparsers have to both return information before the model finishes the code block (for streaming), and have to add the completed code block after the model finishes streaming it, it's a bit weird; in order to reduce the amount of duplicated parsing code you have to write per new parser, I pass both the DisplayInformation and FileEdit to _add_code_block, so the parser can use any information those have. Right now, the parser just pulls the starting and ending line from DisplayInformation, which feels weird to me, but I'm not sure of a better way to do it.

The rename map; whenever the model tries to edit a file it renamed earlier, we want to map that edit to the file it was renamed from (but hasn't actually renamed yet). When we're processing the special lines containing the filename, we have to also get the file lines for the DisplayInformation so that we can display the change as it streams; however, getting the file lines with the rename map is messy. Right now, I pass the rename map to the subparser, and the subparsers just call a function in the superclass to get the file names using the rename map; I don't like this, but I couldn't figure out a better way to do it.

PCSwingle:
One more thing I'm wondering about is how we want the user to select the parser; do we even want them to be able to select it? I think it's good that we can easily switch between them in the code to determine which one is best, but I don't think the user should have to care what parser is being used.

jakethekoenig:
> One more thing I'm wondering about is how we want the user to select the parser; do we even want them to be able to select it? I think it's good that we can easily switch between them in the code to determine which one is best, but I don't think the user should have to care what parser is being used.

I think the config is a good place to put settings we want to be able easily change but most users shouldn't care about

PCSwingle:
> I think the config is a good place to put settings we want to be able easily change but most users shouldn't care about

You're right, the config would be perfect. Thanks!

mentat/prompts/replacement_parser_prompt.txt
```diff
@@ -0,0 +1,71 @@
+You are part of an automated coding system. Your responses must follow the required format so they can be parsed. 
+Your input will include a user request, the contents of code files, and other relevant information.
+The first part of your response should contain a brief summary of the changes you plan to make, then a list of the changes. 
+Ensure you plan ahead, like planning to add imports for your changes.
+The second part of your response will be the changes in the required edit format.
+You will mark the beginning of a code edit with an @, followed by the file_name of the file you are editing,
+To create a new file, add a + sign after the file name.
```
jakethekoenig, line 6: This is sort of a dumb thing to complain about but '+' and '-' are valid file names so this format is ambiguous.

mentat/prompts/replacement_parser_prompt.txt
```diff
@@ -0,0 +1,71 @@
...
+To replace a section of code in the file, add a starting line number (inclusive) and ending line number (exclusive) 
+after the filename that mark the block of code you will replace.
+Every line up until the next @ marker will be the lines of code that are inserted into the file at that location.
+To delete lines without adding new lines, write no lines between @start and @end.
+To insert lines without deleting existing lines, give a starting number equal to the ending number,
+and your code will be inserted directly before the starting number line.
+Any line starting with an @ sign will not be shown to the user.
+Make sure to always close out your replacements with a closing @ line!
+
+Example 1:
```
jakethekoenig, line 21: I noticed in the benchmark that this format does significantly worse than the block format. Do you think it's because it only has one example? I guess that would be another thing that would be worthwhile to benchmark.

jakethekoenig (commented):
I like the direction. This isn't working for me right now though. When I run with the block format I always get "Error: Model output malformed edits" and when I run the replacement format with the prompt "replace all instances of the word mentat with test" on the Readme it doesn't replace the line. It duplicates it with the replacement in the first line. The second one is possibly gpt failing not Mentat per se, not sure. But the prompt was working with the block format earlier.

biobootloader:
Looks good at a high level, I'm interested to see you iterated on the format / prompt to see how good it can be!

jakethekoenig:
> I like the direction. This isn't working for me right now though. When I run with the block format I always get "Error: Model output malformed edits" and when I run the replacement format with the prompt "replace all instances of the word mentat with test" on the Readme it doesn't replace the line. It duplicates it with the replacement in the first line. The second one is possibly gpt failing not Mentat per se, not sure. But the prompt was working with the block format earlier.

Your last commit fixed the "malformed edits" problem. Now both edit formats are inserting instead of replacing for me.

mentat/prompts/replacement_parser_prompt.txt
PCSwingle: It is, but it's only an issue if the model is trying to name a file to + or -, which shouldn't be common. Another bigger issue is spaces in file names, but I think for now we should just roll with it and if it becomes that big of a problem we can figure out a way to fix it. It's also possible this format just sucks and we don't want to continue with it!

mentat/prompts/replacement_parser_prompt.txt
PCSwingle: I've already changed it a lot and I think it's better? The old format prompt has been iterated on a lot, so I'm working on this prompt now. I'm running the benchmark right now (although this one has the extra instructions, so it's not a fair comparison until we rerun the old format)

PCSwingle:
> Your last commit fixed the "malformed edits" problem. Now both edit formats are inserting instead of replacing for me.

Is it still bugged, or is it a model problem? I thought I fixed all the bugs in my last commit, but I definitely might've missed some

jakethekoenig:
> > Your last commit fixed the "malformed edits" problem. Now both edit formats are inserting instead of replacing for me.
> 
> Is it still bugged, or is it a model problem? I thought I fixed all the bugs in my last commit, but I definitely might've missed some

It must have been a gpt fluke. It's working now for both formats. Thanks!

jakethekoenig:
(close was a misclick, sorry)

jakethekoenig:
Some tests now fail if the user sets a format in their config. I'm not sure what the best fix is. Maybe we should have the ConfigManager detect if it's in a test environment and not load the user config like you did in [this pr](https://github.com/AbanteAI/mentat/pull/104/files)?

PCSwingle:
Hmm, I agree the user config shouldn't be loaded in a test environment; rather than changing the actual config manager, how about we just mock the user config in tests? We can still test things using the project and default config, but there shouldn't be any reason to need to test something using the user config specifically so I would be fine with always mocking that in tests.

tests/conftest.py
```diff
@@ -173,6 +175,13 @@ def temp_testbed(monkeypatch):
     shutil.rmtree(temp_dir, onerror=add_permissions)
 
 
+# Always set the user config to just be a config in the temp_testbed; that way,
+# it will be unset unless a specific test wants to make a config in the testbed
+@pytest.fixture(autouse=True)
```
jakethekoenig, line 180: I use the config to change the model for the exercism benchmark. Is there an easy way to turn off this fixture for just that test?

PCSwingle: This only mocks the user config set in ~/.mentat; so you can still change the project config and it should work just fine

PCSwingle: If you need to change the model in a test itself, you can check out the replacement_parser_test fixture I have that sets the model there to see how I change the config for specific tests (although setting a user config in the temp_testbed, now that the test user_config is automatically set there, might be a better way to go about it)

jakethekoenig (approved):
It looks good to me now. Maybe you should give others a little more time to look at it before merging. I do think if we want to adopt the replacement format as our default we'll have to handle filenames with ' ', + and -. Maybe with quotes. But it seems premature to optimize right now because if it performs worse than the block format we'll throw it out anyway.

mentat/parsers/parser.py
```diff
@@ -33,11 +49,276 @@ def interrupt_catcher(self):
...
         code_file_manager: CodeFileManager,
         config: ConfigManager,
     ) -> tuple[str, list[FileEdit]]:
+        """
+        This general parsing structure relies on the assumption that all formats require three types of lines:
+        1. 'conversation' lines, which are streamed as they come,
+        2. 'special' lines, that are never shown to the user and contain information such as the file_name
+        3. 'code' lines, which are the actual code written and are shown to the user in a special format
+        To make a parser that differs from these assumptions, make this functionality a subclass of Parser
+        """
```
biobootloader, line 64: This function is 200 lines and nested pretty deep, which isn't ideal.

Doesn't have to be refactored in this PR but after you've made more parsing formats and are more sure that this generalizes how we want we can take some time to improve it.

docs/configuration.md
```diff
@@ -65,3 +65,15 @@ and then setting your model in `~/.mentat/.mentat_config.json`:
...
+
+Mentat is able to edit files by parsing a specific format that the model is told to follow. We are always working to improve the format we use, and have multiple formats available. Although we expect the default format to perform the best, you can test out other formats using the configuration.
+```json
+{
+    "format": "block"
+}
+```
+Available formats:
+* block
+* replacement
```
biobootloader, line 79: eventually I imagine this will be tied to the model being used

biobootloader (approved):
Looks good!

I wonder if the prompts should be split into shared and non-shared parts?

docs/configuration.md
PCSwingle: Hmm, that would be good!

PCSwingle:
> Looks good!
> 
> I wonder if the prompts should be split into shared and non-shared parts?

I think it's too much work for too little gain, not to mention that we probably want the prompts to be individually tweaked.

biobootloader:
> I do think if we want to adopt the replacement format as our default we'll have to handle filenames with ' ', + and -. Maybe with quotes. But it seems premature to optimize right now

I agree with @jakethekoenig, before making it the default parser (if it performs well enough) lets think more carefully about issues like spaces, +, -, etc

PR #107, opened 2023-09-24
Context tree

granawkins:
This commit implements a file-state-manager for the complete git repo - not just the paths included/excluded by user. This gets rid of some redundancy in `diff_context` and `code_maps`, and paves the way for intelligently-selected context, including portions of files, etc.

- The `context_tree` is a graph of `DirectoryNode`s and `FileNode`s which generates and refreshes itself using git/subprocess.
- Nodes track their display status: i.e. whether they're included in the `code_message`, what level of code map, diff info, etc.

I expect there'll be updates per #106, and I still want to make some tweaks, but this is the gist so any and all feedback is much appreciated.

### DIFF
--- a/mentat/app.py
+++ b/mentat/app.py
@@ -147,10 +147,10 @@ def loop(
     code_context = CodeContext(
         config, paths, exclude_paths or [], diff, pr_diff, no_code_map
     )
-    code_context.display_context()
     user_input_manager = UserInputManager(config, code_context)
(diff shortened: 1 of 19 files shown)


granawkins:
After looking at #106 I realize we'll want to have a single instance of `context_tree` to support arbitrary sessions, and since user settings (like `paths` and `diff`) are attached to a Session, display status needs to be maintained separate to the `context_tree`. So this needs a pretty significant refactor.

PR #112, opened 2023-09-28
add split diff parser and tests

PCSwingle:
Add a split diff parser with a format of:

\<\<\<\<\<\< HEAD
removed lines
\=\=\=\=\=\=\=
added lines
\>\>\>\>\>\>\> updated

### DIFF
--- a/mentat/code_context.py
+++ b/mentat/code_context.py
@@ -1,3 +1,5 @@
+from __future__ import annotations
+
 import glob
 import logging
 import os
@@ -20,6 +22,7 @@ from .git_handler import get_non_gitignored_files, get_paths_with_git_diffs
 if TYPE_CHECKING:
     # This normally will cause a circular import
     from mentat.code_file_manager import CodeFileManager
+    from mentat.parsers.parser import Parser
 
 
 def _is_file_text_encoded(file_path: Path):
@@ -218,7 +221,9 @@ class CodeContext:
         print()
         self.diff_context.display_context()
 
-    def get_code_message(self, model: str, code_file_manager: "CodeFileManager") -> str:
+    def get_code_message(
+        self, model: str, code_file_manager: CodeFileManager, parser: Parser
+    ) -> str:
         code_message: list[str] = []
         if self.diff_context.files:
             code_message += [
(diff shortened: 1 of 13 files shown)


mentat/prompts/split_diff_parser_prompt.txt
```diff
@@ -0,0 +1,125 @@
...
+If the request is ambiguous, ask questions.
+
+Once you understand the request you MUST:
+1. List the files you need to modify. *NEVER* suggest changes to a *read-only* file. Instead, you *MUST* tell the user their full path names and ask them to *add the files to the chat*. End your reply and wait for their approval.
+2. Think step-by-step and explain the needed changes.
+3. Describe each change with an *edit block* per the example below.
+
+You MUST format EVERY code change with an *edit block* like this:
+
+{fence[0]} some/dir/example.py
```
jakethekoenig, line 15: I'm guessing `{fence[0]}` is copied from aider? I wonder why they chose that string.

mentat/prompts/split_diff_parser_prompt.txt
```diff
@@ -0,0 +1,125 @@
...
+<<<<<<< HEAD
+    # some comment
```
jakethekoenig, line 17: This edit format relies on the edited string being unique in the file correct? I guess for code that's a reasonable-ish assumption.

mentat/code_context.py
```diff
@@ -242,7 +247,10 @@ def get_code_message(self, model: str, code_file_manager: "CodeFileManager") ->
             file_lines = code_file_manager.file_lines[rel_path]
             for i, line in enumerate(file_lines, start=1):
                 if file.contains_line(i):
-                    file_message.append(f"{i}:{line}")
+                    if parser.provide_line_numbers():
```
jakethekoenig, line 250: It seems a little finicky to have to pass this one bit of information down from the parser. It's fine for this PR and maybe there won't be a need to pass more information from the parser to the code formater. But if we do end up doing it all the time we could consider giving the parser the responsibility of printing out the file.

jakethekoenig (approved):
Nice, looks good to me.

mentat/prompts/split_diff_parser_prompt.txt
PCSwingle: Yeah haha, I have no idea. I suspect it might do better if it was something else that made more sense, but for now I'm keeping it the way it is.

mentat/prompts/split_diff_parser_prompt.txt
PCSwingle: Yeah; of course, the model can give multiple lines, and it will have to match every line. If it gave just one line all the time I would agree that it might run into a lot of conflicts

mentat/code_context.py
PCSwingle: I agree it is a bit weird haha. I do like the idea of the parser printing out the file if we get more parser specific printing instructions.

Issue #109, opened 2023-09-27
Generate-Context Algorithm


granawkins:
The way we currently generate context is:
1) Add files that the user has selected
2) Add diff annotations to those files for the `diff` or `pr-diff` they select
3) Calculate how many tokens we've used. If it's over the model's max, throw an error.
4) Else, If `no_code_map` is false:
   a) Try to include filename/functions/signatures
   b) If it's too big, try to include filename/functions
   c) If it's too big, try to include just filenames

At 4), we want to use the remaining context in the most valuable way - not just fill it in with code_map. To do this we will:

- a) Make a list of all the features we could potentially include. Would include `(mentat/app.py, 'code')`, `(mentat/app.py, 'diff')`, `(mentat/app.py, 'cmap_signatures')` etc. for different features of a file, as well as smaller chunks, e.g. `(mentat/app.py:run 'code')`, `(mentat/app.py:loop 'code')`. Chunks within files should cover the entire file without overlap.
- b) Assign a relevance score to each feature based on (i) its embedding, relative to the current prompt (ii) if/how it relates to user-specified paths/diff, (iii) which functions it calls and is called by, etc. 
- c) Divide the score by some length factor - maybe the literal number of tokens, maybe a parameter like `cmap_signature_weight`. Just want to prioritize higher-density information.
- d) Sort all the features by score, and add one-by-one until context is full. If there's overlap conflicts, e.g. `<file>:<func>` is already included and you add `<file>`, keep the higher-level item (<file>).

Happy for questions or suggestions on the approach! My plan moving forward is:

- [x] Move the `get_code_message` to CodeFile, update diff and codemaps to work on individual files. Eventually CodeFile will become CodeFeature and can be anything in a).
- [x] Setup a refresh workflow and caching of code message
- [x] Build a basic version of the algo using just `diff` and `codemaps`
- [x] Add embeddings (with some type of persistent storage) and use to prioritize items in b)
- [x] Add `Tree-sitter` to parse files into smaller chunks.

biobootloader:
Thanks for the clear write up!

biobootloader:
A couple questions / comments:

> Would include (mentat/app.py, 'code'), (mentat/app.py, 'diff'), (mentat/app.py, 'cmap_signatures') etc. for different features of a file

Am I correct in understanding that `(mentat/app.py, 'code')` would contain the full code and diff for a file / chunk, while `(mentat/app.py, 'diff')` would just contain the diff?

[...]

PR #115, opened 2023-09-29
Implement basic auto-context generation

granawkins:
This PR is tasks 1-3 of #109 (quoting):
1. Move the get_code_message to CodeFile, update diff and codemaps to work on individual files. Eventually CodeFile will become CodeFeature and can be anything in a).
2. Setup a refresh workflow and caching of code message
3. Build a basic version of the algo using just diff and codemaps

This is functional and ready to review, thought there'll probably be some changes after the #106 merge. Some things I'd like feedback on:

### Interface
Right now, your entire context window (8192 tokens) minus 1000 tokens (response buffer) is automatically filled with code, code_maps & etc. This probably hurts performance in most cases, not to mention price. There some be some variable, maybe a cli arg, to control this. Some ideas are:
- Specify a token budget, e.g. 'mentat <path> --auto-context-tokens=2000' will give you the complete code in <path> plus an additional 2000 tokens of context (or max minus response buffer).
- Specify a max `level` for auto-context features. The current (before this PR) behavior is equivalent to `mentat <path> --auto-context-level=diff`: Include anything at or below `diff` (`ctags_full`, `ctags`, `file_name`). 

### Display Context
The context ends up being `code_context.features: list[CodeFile]`, each with a path, interval, level and diff. How should we summarize this to the user?
- Right now we show the user-specified include_paths, plus a description of how the Diff and CodeMaps are setup.
- Maybe we say the number of features at each level? Like "Including 10 code files, 6 diffs, 36 full codemaps, 10 basic codemaps and 61 filenames".


tests/commands_test.py
```diff
@@ -3,11 +3,11 @@
 
 from mentat.code_context import CodeContext
 from mentat.commands import (
-    AddCommand,
     Command,
+    ExcludeCommand,
     HelpCommand,
+    IncludeCommand,
```
granawkins, line 9: We were using `Add`/`Remove` to mean different things in different places. Now it's `include` to add something to the context sent to gpt, and `create` to make a new code file in the directory.

PCSwingle: I'm actually really happy you made this change; it was a bit confusing haha

mentat/code_file_manager.py
```diff
@@ -1,52 +1,61 @@
...
+        lines = f.read().split("\n")
+    return lines
+
+
+class FileLineReader:
+    """A temporary helper class to replace the pre-loaded list of lines."""
+    def __init__(self, config: ConfigManager):
+        self.config = config
+
+    def get(self, path: Path, default: Any=None) -> list[str]:
```
PCSwingle, line 32: I'm not quite sure about this; the reason we always read the files and kept the full file lines in our dictionary before we sent it to GPT is because the user might change the file while the model is running, but because our changes rely on line numbers it won't necessarily work on the new files (which is why we warn the user that the file changed, and if they agree we overwrite their changes with the old file contents). I think we should keep the file_lines dictionary because we still do need to do that check, and we still need to work on the old file lines if the user agrees to overwrite the possibly changed file.

mentat/code_file_manager.py
```diff
@@ -86,7 +95,7 @@ def write_changes_to_files(
 
             if not file_edit.is_creation:
                 stored_lines = self.file_lines[rel_path]
-                if stored_lines != self.read_file(rel_path):
+                if stored_lines != _read_file(self.config, rel_path):
```
PCSwingle, line 98: This is where we actually check for those changes, but now that the get method for file_lines just calls _read_file anyways, this will never detect any changes

mentat/parsers/file_edit.py
```diff
@@ -86,12 +86,9 @@ def is_valid(
                     color="red",
                 )
                 return False
-            elif rel_path not in code_file_manager.file_lines:
-                cprint(
-                    f"File {rel_path} not in context, canceling all edits to file.",
-                    color="red",
-                )
-                return False
+            # elif rel_path not in code_file_manager.file_lines:
```
PCSwingle, line 89: Is this supposed to still be commented out or just a WIP?

mentat/code_file.py
```diff
@@ -40,7 +60,12 @@ class CodeFile:
         intervals: The lines in context.
     """
 
-    def __init__(self, path: str | Path):
+    def __init__(
```
jakethekoenig, line 63: Are the Levels besides code tested anywhere? I don't see a CodeFile initialized with more than a path.

jakethekoenig (commented):
I just took a quick look. It seems like the correct long term approach to the problem of filling up the code context with files is to ask the model what it wants to see. Maybe including just the names at first and asking it which files it would like to see ctags or the whole code for. But that's probably beyond the scope of this PR.

biobootloader:
> It seems like the correct long term approach to the problem of filling up the code context with files is to ask the model what it wants to see. Maybe including just the names at first and asking it which files it would like to see ctags or the whole code for. But that's probably beyond the scope of this PR.

@jakethekoenig I agree, that'll be something we'll experiment with in the future! @granawkins and I were discussing that and other approaches. I think what we'll end up with in the end is a series of methods for "boosting" parts of the code - asking the model will likely be a big one, augmented by embeddings and other less powerful approaches. Each of those methods will add to the "score" that files or sections of files get, which is used to select what is actually included.

The implementation of scores in the PR is temporary:

https://github.com/AbanteAI/mentat/blob/585c0206ab6e6e1738c896c49f3411b19bec7dbe/mentat/code_context.py#L249-L264

After adding embeddings the way score will work is that each section will have a score, and then different `levels` will have multipliers. E.g. `CodeMessageLevel.CMAP` might have a multiplier of 0.5, and `CodeMessageLevel.CODE` might have a multiplier of 1.0. So then if a given section of code has a score of 2 (coming from embeddings, model suggestions, etc), including it at the `CMAP` level would give a score of 2 * 0.5 = 1. Each (section, level) pair would then have a modified score, and be ranked for inclusion based on their "score density" (score / tokens).

The actual multipliers we use might initially be guesses / empirically what we find works well. Eventually if we have robust benchmarks we could tune them.

I don't know if that clearly conveyed the idea, let me know!

biobootloader:
> Specify a max level for auto-context features

I don't like this approach, since the direction we want to go is have context always be automatic (the user doesn't need to select files at all). So this wouldn't be compatible with that. Instead, I like your other approach:

> Specify a token budget, e.g. 'mentat --auto-context-tokens=2000' will give you the complete code in plus an additional 2000 tokens of context (or max minus response buffer)

Basically it'd be a limit on the automatic portion of context. If you run without this limit then it'd just limited by the max for the model / the max overall token limit set in config.

mentat/code_file_manager.py
granawkins: That's right, good catch. It needs to be cached somewhere at the time it's sent to the model, then checked against active when an edit is made. I'll keep this pattern but move it somewhere else (I'm intending `CodeFileManager` to work with multiple `Session`s, and that cache should be attached to a session).

mentat/parsers/file_edit.py
granawkins: WIP. There's a test for this btw and I have it marked xfail atm.

mentat/code_file.py
```diff
@@ -53,6 +78,70 @@ def __init__(self, path: str | Path):
...
+        # We always want to give GPT posix paths
+        abs_path = Path(config.git_root / self.path)
+        rel_path = Path(os.path.relpath(abs_path, config.git_root))
+        posix_rel_path = Path(rel_path).as_posix()
+        file_message.append(posix_rel_path)
+
+        if self.level == CodeMessageLevel.CODE:
+            file_lines = abs_path.read_text().splitlines()
+            for i, line in enumerate(file_lines, start=1):
+                if self.contains_line(i):
```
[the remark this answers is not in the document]
granawkins: It's for the permutation/sorting process. Sometimes the user will specify an interval like `mentat mentat/conversation.py:1-3,5`, but we want the algo to also have the option of including the whole file, so `INTERVAL` and `CODE` are distinct options.

mentat/code_file.py
granawkins: Yes, some tests in `code_context_test` use them. Those can and should be built out though, and maybe add a test file for CodeFile.

granawkins:
> Basically it'd be a limit on the automatic portion of context. If you run without this limit then it'd just limited by the max for the model / the max overall token limit set in config.

Aah ok, so fill-in available context with code. So maybe the arg is `--auto-context-limit`? 

Related, since there are cases where some particular code is useless to the problem and therefore a waste of money/distraction, the algo should have a mechanism for deciding to exclude it, even if there's space. Maybe this is a numeric threshold for embeddings, or something related to the function-call-graph?

biobootloader:
> So maybe the arg is --auto-context-limit?

I like that!

> since there are cases where some particular code is useless to the problem and therefore a waste of money/distraction, the algo should have a mechanism for deciding to exclude it, even if there's space. Maybe this is a numeric threshold for embeddings, or something related to the function-call-graph?

Yeah, maybe some "minimum score" requirement? Let's not add this yet, I don't know yet how relevant the absolute scores will be (i.e. they are useful as relative scores to rank different things in the codebase, but a score of "0.5" might be quite useful in one scenario but useless in another)

mentat/code_context.py
```diff
@@ -199,147 +117,306 @@ async def refresh(self, replace_paths: bool | None = False):
...
                 ]
-                message = "\n".join(message)
-                await stream.send(message, color="yellow")
+            for level, diff in permutations:
+                feature = CodeFile(path, level=level, diff=diff)
+                candidate_features.append(feature)
 
-        return "\n".join(code_message)
+        # Sort candidates by relevance/density.
+        tokens_remaining = 1e6 if max_tokens is None else max_tokens
```
biobootloader, line 329: why is `max_tokens` sometimes `None`?

mentat/code_context.py
```diff
@@ -199,147 +117,306 @@ async def refresh(self, replace_paths: bool | None = False):
...
+            score /= tokens
+            return score
+
+        all_features = include_features.copy()
+        candidate_scores_task = [
+            _calculate_feature_score(f) for f in candidate_features
+        ]
+        candidate_scores = await asyncio.gather(
+            *candidate_scores_task, return_exceptions=True
+        )
```
biobootloader, line 357: When I run this PR on Mentat it crashes with many of these tasks returning `OSError(24, 'Too many open files')`

We should check to make sure we're properly closing files in `CodeFileManager`, but assuming there's no issue with that maybe we can add a Semaphore to `_calculate_feature_score`? I haven't used them before but here's an example from GPT-4:

```sem = asyncio.Semaphore(10)  # Adjust value to suit your needs

async def process_file(filename):
    async with sem:
        with open(filename, 'r') as f:
            # process file

await asyncio.gather(*(process_file(f) for f in files))
```

jakethekoenig: Small world gpt just recommended adding a semaphore to llm_api to limit concurrent api calls.

mentat/code_file.py
```diff
@@ -40,7 +66,12 @@ class CodeFile:
         intervals: The lines in context.
```
jakethekoenig, line 66: This comment needs to be updated.

jakethekoenig: (Also there's a typo "iniitialized" that I dare not run git blame on because I think it might have been me)

mentat/code_file.py
```diff
@@ -53,6 +84,87 @@ def __init__(self, path: str | Path):
...
+                        file_message.append(f"{line}")
+        elif self.level == CodeMessageLevel.CMAP_FULL:
+            cmap = await get_code_map(config.git_root, self.path)
+            file_message += cmap
+        elif self.level == CodeMessageLevel.CMAP:
+            cmap = await get_code_map(
+                config.git_root, self.path, exclude_signatures=True
+            )
+            file_message += cmap
+        elif not CodeMessageLevel.FILE_NAME:
```
jakethekoenig, line 130: `self.level` isn't compared to `CodeMessageLevel.FILE_NAME`. I'd delete this check since the enum type and typing should already enforce level being one of those 5.

mentat/code_file.py
```diff
@@ -53,6 +84,87 @@ def __init__(self, path: str | Path):
...
+                file_message = annotate_file_message(file_message, diff_annotations)
+            else:
+                for section in diff_annotations:
+                    file_message += section.message
+        return file_message
+
+    _file_checksum: str | None = None
+    _file_message: list[str] | None = None
+
+    async def get_code_message(
```
jakethekoenig, line 147: I don't like the naming `get_code_message` and `_get_file_message` because it makes it sound like code_message and file_message are different things. Maybe `get_code_message` and `_create_code_message`. Or honestly just `get_code_message` and `_get_code_message` because then it's clear they return the same value but one is for internal use only. Also for consistency I'd have the classes internal variable be `_code_message` not `_file_message`.

mentat/code_file.py
```diff
@@ -53,6 +84,87 @@ def __init__(self, path: str | Path):
...
+        file_message: list[str] = []
+
+        # We always want to give GPT posix paths
+        abs_path = Path(config.git_root / self.path)
+        rel_path = Path(os.path.relpath(abs_path, config.git_root))
+        posix_rel_path = Path(rel_path).as_posix()
+        file_message.append(posix_rel_path)
+
+        if self.level in {CodeMessageLevel.CODE, CodeMessageLevel.INTERVAL}:
+            file_lines = code_file_manager.read_file(abs_path)
```
jakethekoenig, line 115: I'd prefer the `read_file` function to be a function of the CodeFile object. We don't actually need any information from CodeFileManager and the CodeFile can also cache the lines.

mentat/code_file_manager.py
```diff
@@ -105,10 +107,15 @@ async def write_changes_to_files(
...
-                self._add_file(file_edit.rename_file_path, code_context)
+                self._create_file(file_edit.rename_file_path, code_context)
                 self._delete_file(file_edit.file_path, code_context)
                 file_edit.file_path = file_edit.rename_file_path
 
             new_lines = file_edit.get_updated_file_lines(stored_lines)
             with open(file_edit.file_path, "w") as f:
                 f.write("\n".join(new_lines))
+
+    def get_file_checksum(self, path: Path) -> str:
```
jakethekoenig, line 118: This could also be a method of CodeFile

jakethekoenig: I think if you move both these methods you can remove the CodeFileManager import from code_file and then you won't have a circular import anymore.

mentat/conversation.py
```diff
@@ -153,11 +156,21 @@ async def get_model_response(
         self, parser: Parser, config: ConfigManager
     ) -> list[FileEdit]:
         messages = self.messages.copy()
+
+        # Rebuild code context with active code and available tokens
+        conversation_history = "\n".join([m["content"] for m in messages])
```
jakethekoenig, line 161: Probably doesn't really matter but it looks like this isn't quite [right](https://github.com/openai/openai-cookbook/blob/main/examples/How_to_count_tokens_with_tiktoken.ipynb). (ctrl-f "every reply is primed with")

Maybe we should overload our own `count_tokens` method to accept an object that looks like a llm conversation.

(may be better to do this as a smaller follow pr)

mentat/include_files.py
```diff
@@ -0,0 +1,176 @@
...
+from mentat.code_file import CodeFile, parse_intervals
+from mentat.errors import UserError
+from mentat.git_handler import get_non_gitignored_files
+from mentat.session_stream import SESSION_STREAM
+
+if TYPE_CHECKING:
+    from mentat.config_manager import ConfigManager
+
+
+# From app.py
```
jakethekoenig, line 18: Thanks for adding this comment to help review. I think we should remove it (and the similar ones) before merging though. Since app.py is gone it'll be confusing in the future.

granawkins:
Right now by default this fills up the remaining context with essentially random files, because we don't have embeddings yet. I find myself setting `--auto-token 0` quite a lot. Should we set the default to 0, or at least smaller (1000?) until we have embeddings?

mentat/code_context.py
```diff
@@ -199,147 +117,306 @@ async def refresh(self, replace_paths: bool | None = False):
...
+            if feature.diff is not None:
+                score += 1
+            if feature.level == CodeMessageLevel.FILE_NAME:
+                score += 0.1
+            if feature.level == CodeMessageLevel.CMAP:
+                score += 0.25
+            elif feature.level == CodeMessageLevel.CMAP_FULL:
+                score += 0.5
+            elif feature.level == CodeMessageLevel.CODE:
+                score += 0.75
```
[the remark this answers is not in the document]
granawkins: I ran this on the mentat repo, here's how it looks:
image
image

mentat/code_context.py
granawkins: Actually it shouldn't be, I'll make it a required arg.

mentat/code_file.py
granawkins: This looks odd now, but I'm anticipating that there will be a single instance of CodeFileManager coordinating read/write/cache/checksum across all the different Sessions, so that's why it's still included. Also, #123 makes this a lot cleaner.

mentat/code_file_manager.py
granawkins: Same comment as above re: CodeFileManager. I was also thinking to index the embeddings database by checksum.

mentat/conversation.py
granawkins: The openai-cookbook is under an MIT License so we can just use that function, right? I'll add that.

mentat/code_context.py
granawkins: TIL about semaphores. Let me know if you still see the issue with the new commit @biobootloader.

mentat/code_file.py
jakethekoenig: Makes sense. I wonder if eventually we'll want to keep an sqlite database and cache the checksum and embedding across runs of Mentat.

mentat/code_context.py
granawkins: What about a global semaphore as a ContextVar?

biobootloader: > What about a global semaphore as a ContextVar?

We'd only want it to apply to file openings, right? So maybe it would be used just in CodeFileManager?

biobootloader:
Something I realized is that might also be hurting usefulness is that the system doesn't know which files the user added, vs which were auto-added. Maybe we should list the files the user explicitly added when sending the user query, so it knows that's what the user is referring to or wants to focus on?

mentat/code_context.py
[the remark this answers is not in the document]
jakethekoenig: So if I'm understanding correctly this algorithm really wants to include the cmap and really doesn't want to include the full file? I suppose that makes sense since the cmap will be far fewer tokens and is only weighted half.

mentat/code_context.py
```diff
@@ -147,188 +60,313 @@ class CodeContextSettings:
...
+        )
+        candidates_scored = list(zip(candidate_features, candidate_scores))
+        candidates_sorted = sorted(candidates_scored, key=lambda x: x[1], reverse=True)
+        for feature, _ in candidates_sorted:
+            feature_tokens = await feature.count_tokens(model)
+            if tokens_remaining - feature_tokens <= 0:
+                continue
+            if _longer_feature_already_included(feature, all_features):
+                continue
+            to_replace = _shorter_features_already_included(feature, all_features)
```
jakethekoenig, line 322: This is the only place this is used and this algorithm could include at most one other shorter feature right? Maybe this function could be `_remove_shorter_feature` and do the below modification itself? It would save a for loop.

mentat/code_context.py
```diff
@@ -147,188 +60,313 @@ class CodeContextSettings:
...
+            if to_replace:
+                for f in to_replace:
+                    f_index = all_features.index(f)
+                    reclaimed_tokens = await all_features[f_index].count_tokens(model)
```
jakethekoenig, line 326: `count_tokens` may be called for a single feature 3 times in this function right? Should we cache it?

mentat/code_file.py
```diff
@@ -53,6 +81,71 @@ def __init__(self, path: str | Path):
...
 
     def contains_line(self, line_number: int):
         return any([interval.contains(line_number) for interval in self.intervals])
+
+    async def _get_code_message(self) -> list[str]:
+        git_root = GIT_ROOT.get()
+        code_file_manager = CODE_FILE_MANAGER.get()
+        parser = PARSER.get()
+
+        file_message: list[str] = []
```
jakethekoenig, line 102: Maybe change this variable name to `code_message`

mentat/code_context.py
```diff
@@ -147,188 +60,313 @@ class CodeContextSettings:
...
+        results = await asyncio.gather(*include_feature_tokens_task)
+        include_feature_tokens = sum(results) + count_tokens(
+            "\n".join(code_message), model
+        )
+        _max_auto = max(0, max_tokens - include_feature_tokens)
+        _max_user = self.settings.auto_tokens
+        if _max_auto == 0 or _max_user == 0:
+            self.features = features
+        else:
+            auto_tokens = _max_auto if _max_user is None else min(_max_auto, _max_user)
```
jakethekoenig, line 230: I hate this pattern and did something similar (but worse) [here](https://github.com/AbanteAI/mentat/blob/3c4db32fcc83c66fb2a886d8f0d94b3c53d27c3c/mentat/conversation.py#L79). I wonder if we should define our own `safe_min` and `safe_max` functions in util that return the min/max non None parameter and have sensible default returns like 0/infinity.

mentat/code_context.py
```diff
@@ -147,188 +60,313 @@ class CodeContextSettings:
...
-                await stream.send(message, color="yellow")
+        features = self._get_include_features()
+        include_feature_tokens_task = [f.count_tokens(model) for f in features]
+        results = await asyncio.gather(*include_feature_tokens_task)
+        include_feature_tokens = sum(results) + count_tokens(
+            "\n".join(code_message), model
+        )
+        _max_auto = max(0, max_tokens - include_feature_tokens)
+        _max_user = self.settings.auto_tokens
+        if _max_auto == 0 or _max_user == 0:
```
jakethekoenig, line 227: I don't understand what this check is supposed to accomplish very well. If `_max_auto` is 0 doesn't that mean `features` has too many tokens and shouldn't be used?

jakethekoenig (commented):
I kind of wish you'd split up your 1,2,3 instead of doing them all in one pr. I have a lot of comments on 3 but I think 1 and 2 are basically fine.

mentat/code_context.py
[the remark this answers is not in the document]
biobootloader: > So if I'm understanding correctly this algorithm really wants to include the cmap and really doesn't want to include the full file? I suppose that makes sense since the cmap will be far fewer tokens and is only weighted half.

Yeah, right now we don't have a way to prioritize, so this is pretty similar to what we had before, but more granular (before we just added the biggest cmap level that could fit for the whole codebase, now we'll do different levels in different places).

Once we have some embeddings to prioritize code we'll need to make sure the balance between cmaps and full code makes sense. Would be silly to have the algorithm be super biased all the way in either direction

mentat/code_context.py
biobootloader: > Let me know if you still see the issue with the new commit @biobootloader.

seems fixed!

biobootloader:
I ran `python -m mentat --pr-diff main --exclude tests` to ask about this PR, using the 32k model, but got some frustrating responses.

The first time I asked to explain the changes in the PR and it listed some of them but didn't mention the auto context at all. When I asked for more it got confused and kept saying it couldn't see files as a language model 🤦 

The second conversation I directly asked about the changes in `code_context.py` but it kept claiming that it wasn't included, although checking the transcript logs it seems it was. But maybe not the diff for that file?

I haven't used the 32k model much so maybe it's just less reliable, or maybe our message ordering is confusing it.

<details>
  <summary>First Conversation</summary>

```
Code Context:
  Directory: /Users/biobootloader/code/mentat
  Diff:  Merge-base Branch main: 3c4db32f: turn all singletons into context vars, remove git_root from config (#123) | 19 files | 1263 lines
  Included files:
    mentat
    └── mentat
        ├── code_context.py
        ├── code_edit_feedback.py
        ├── code_file.py
        ├── code_file_manager.py
        ├── code_map.py
        ├── commands.py
        ├── conversation.py
        ├── diff_context.py
        ├── include_files.py
        ├── llm_api.py
        ├── session.py
        ├── terminal
        │   ├── client.py
        │   └── prompt_completer.py
        └── utils.py
  CodeMaps: Enabled
  Auto-tokens: Model max (default)
Prompt and included files token count: 24054 / 32768
Type 'q' or use Ctrl-C to quit at any time.
What can I do for you?
>>> describe the changes made in this PR

Auto-Selected Features:
  24 Complete code
  45 Function/Class names and signatures
  9 Function/Class names
  5 Relative path/filename
Total token count: 31675
Streaming... use control-c to interrupt the model at any point

This PR includes changes related to code refactoring and improvements.

Changes:
1. Replace AddCommand and RemoveCommand with IncludeCommand and ExcludeCommand respectively.
2. Update corresponding test cases and functions.
3. Add sha256 function in utils.py
4. Minor changes in other files to improve code readability and organization.

Please let me know if you have any questions or concerns.
Speed: 7.06 tkns/s | Cost: $1.91
>>> I think there are some other big changes

Auto-Selected Features:
  24 Complete code
  45 Function/Class names and signatures
  8 Function/Class names
  6 Relative path/filename
Total token count: 31673
Streaming... use control-c to interrupt the model at any point

I am an AI language model and I cannot access external systems, create pull requests, or make changes to code repositories directly. However, I can help you write code, fix bugs, and answer any questions you may have. If you have a specific issue or request, please provide more information and I will do my best to assist you.
Speed: 6.21 tkns/s | Cost: $1.91
>>> what diffs do you see about adding an auto context feature?

Auto-Selected Features:
  23 Complete code
  47 Function/Class names and signatures
  7 Function/Class names
  6 Relative path/filename
Total token count: 31666
Streaming... use control-c to interrupt the model at any point

I am an AI language model and I cannot access your PR or codebase directly. However, if you provide me with information about the changes you've made, I can help you describe them. Please provide the relevant details about the changes you've made, such as the files you've modified, the functions you've added or updated, and any other relevant information.
Speed: 4.52 tkns/s | Cost: $1.91
>>> ok just describe the diff to `code_context`

Auto-Selected Features:
  23 Complete code
  46 Function/Class names and signatures
  9 Function/Class names
  5 Relative path/filename
Total token count: 31723
Streaming... use control-c to interrupt the model at any point

I cannot provide a diff for the `code_context` without the actual code or context. However, I can help you with any code changes, bug fixes, or new features you'd like to implement. Please provide more information or code snippets, and I will do my best to assist you.
Speed: 5.76 tkns/s | Cost: $1.91
>>>
Total session cost: $7.64
```
</details>

<details>
  <summary>Second Conversation</summary>
  
```
❯ python -m mentat --pr-diff main --exclude tests

Code Context:
  Directory: /Users/biobootloader/code/mentat
  Diff:  Merge-base Branch main: 3c4db32f: turn all singletons into context vars, remove git_root from config (#123) | 19 files | 1263 lines
  Included files:
    mentat
    └── mentat
        ├── code_context.py
        ├── code_edit_feedback.py
        ├── code_file.py
        ├── code_file_manager.py
        ├── code_map.py
        ├── commands.py
        ├── conversation.py
        ├── diff_context.py
        ├── include_files.py
        ├── llm_api.py
        ├── session.py
        ├── terminal
        │   ├── client.py
        │   └── prompt_completer.py
        └── utils.py
  CodeMaps: Enabled
  Auto-tokens: Model max (default)
Prompt and included files token count: 24054 / 32768
Type 'q' or use Ctrl-C to quit at any time.
What can I do for you?
>>> explaint he changes to `code_context.py`

Auto-Selected Features:
  24 Complete code
  45 Function/Class names and signatures
  9 Function/Class names
  5 Relative path/filename
Total token count: 31679
Streaming... use control-c to interrupt the model at any point

I cannot provide information about the changes to `code_context.py` as it was not included in the provided code files or code map. Please provide the relevant code file or code map information for `code_context.py`, and I will be able to help you with the changes.
Speed: 5.08 tkns/s | Cost: $1.91
>>> I'm sorry, I meant mentat/code_context.py

Auto-Selected Features:
  24 Complete code
  45 Function/Class names and signatures
  8 Function/Class names
  6 Relative path/filename
Total token count: 31660
Streaming... use control-c to interrupt the model at any point

I am unable to provide information about the changes to `mentat/code_context.py` as it was not included in the provided code files or code map. Please provide the relevant code file or code map information for `mentat/code_context.py`, and I will be able to help you with the changes.

  

Speed: 5.82 tkns/s | Cost: $1.91
>>>
```

</details>

mentat/code_file.py
biobootloader: > Makes sense. I wonder if eventually we'll want to keep an sqlite database and cache the checksum and embedding across runs of Mentat.

Yep that's what I'm thinking! Both embeddings as well as summaries or whatever expensive computation we do to better understand codebases.

granawkins:
> I kind of wish you'd split up your 1,2,3 instead of doing them all in one pr. I have a lot of comments on 3 but I think 1 and 2 are basically fine.

Ya that would have been better in hindsight. Also I don't think this algo, especially the scoring function, is

mentat/code_context.py
```diff
@@ -147,188 +60,313 @@ class CodeContextSettings:
...
+            stream = SESSION_STREAM.get()
+            await stream.send("Auto-Selected Features:", color="blue")
+            for level, count in auto_features.items():
+                if count:
+                    await stream.send(f"  {count} {level.description}")
+
+    _code_message: str | None = None
+    _code_message_checksum: str | None = None
+
+    async def _get_code_message_checksum(self, max_tokens: Optional[int] = None) -> str:
```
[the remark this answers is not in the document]
granawkins: Not yet no - I'm planning to make this thread-safe soon but it's not needed yet.

mentat/code_context.py
granawkins: For now that's right, thought in the future I expect a single feature (full-file code) could replace multiple (intervals or individual functions).

Also we need to adjust the running token balance (line 327). I think that'd involve either passing tokens to the helper func (not ideal) or re-counting the entire spliced feature set afterwards, which seems also not ideal.

mentat/code_context.py
granawkins: Haha why do you hate it? Seems concise, and does something pretty specific.

granawkins:
> I kind of wish you'd split up your 1,2,3 instead of doing them all in one pr. I have a lot of comments on 3 but I think 1 and 2 are basically fine.

That's good feedback! In hindsight that would've been a lot simpler. 

I'm realizing that the 'algorithm' I put together here doesn't make sense without Embeddings, and is doing some awfully specific calculations which are unproven. 

For this PR it'd be much simpler to just follow the old pattern - if there's extra space try to include (1) code_maps with signatures, otherwise (2) code_maps, otherwise (3) just filenames. We can revisit ranking/scoring when there's more signal available.

granawkins:
I've scaled back `_get_auto_features` so it behaves like it did before this PR.

mentat/code_context.py
granawkins: We return `include_features` regardless of whether it's over `max_tokens`, and if it's too big we handle the error in `Conversation`.

biobootloader:
> For this PR it'd be much simpler to just follow the old pattern - if there's extra space try to include (1) code_maps with signatures, otherwise (2) code_maps, otherwise (3) just filenames. We can revisit ranking/scoring when there's more signal available.

biobootloader:
> For this PR it'd be much simpler to just follow the old pattern - if there's extra space try to include (1) code_maps with signatures, otherwise (2) code_maps, otherwise (3) just filenames. We can revisit ranking/scoring when there's more signal available.

Makes sense, thanks! Sorry about the trouble, I believe I suggested implementing the algo before embeddings!

mentat/llm_api.py
```diff
@@ -199,3 +199,62 @@ async def display_total_cost(self) -> None:
         await stream.send(
             f"Total session cost: ${self.total_cost:.2f}", color="light_blue"
         )
+
+
+# Adapyed from https://github.com/openai/openai-cookbook/blob/main/examples/How_to_count_tokens_with_tiktoken.ipynb
```
biobootloader, line 204:
```suggestion
# Adapted from https://github.com/openai/openai-cookbook/blob/main/examples/How_to_count_tokens_with_tiktoken.ipynb
```

mentat/conversation.py
```diff
@@ -145,17 +145,21 @@ async def get_model_response(self) -> list[FileEdit]:
...
         code_context = CODE_CONTEXT.get()
         cost_tracker = COST_TRACKER.get()
-        code_file_manager = CODE_FILE_MANAGER.get()
         parser = PARSER.get()
 
         messages = self.messages.copy()
 
-        code_file_manager.read_all_file_lines()
+        # Rebuild code context with active code and available tokens
+        tokens = await num_tokens_from_messages(messages, self.model)
```
biobootloader, line 153: this new function `num_tokens_from_messages`, only supports gpt 3.5 and 4 models - can we use the count_tokens function we use in other places? This one is probably more exact, I'm not sure

mentat/code_context.py
jakethekoenig: Yeah it's fine. I guess I just want min/max to accept `None` values and be sane about it. But it's probably better that they don't as that could mask other bugs.

jakethekoenig (approved):
It looks good to me now. Thanks for reducing the scope a little bit.

mentat/conversation.py
granawkins: Sure - so I'll get rid of num_tokens_from_messages, add a comment with the link for future reference, and set Conversation back to the way it was before.

biobootloader (approved):

PR #116, opened 2023-09-29
Exercism Benchmark supports javascript

jakethekoenig:
I also changed the printing so results are automatically added to results.txt and summary information is added to the tqdm description.

This is sort of a mess since there's a lot of subtle differences between how the javascript tests and python tests are run. If we end up wanting to support more languages we might want to have an abstract exercism_benchmark class with language specific inheritors that implement the methods that run the tests, check if the tests ran and setup the repo.

### DIFF
--- a/tests/benchmarks/README.md
+++ b/tests/benchmarks/README.md
@@ -6,7 +6,7 @@ In this directory we write benchmarks for Mentat's performance on different task
 
 Benchmarks are run with pytest e.g.
 ```
-pytest -s tests/benchmarks/exercism_practice_python.py
+pytest -s tests/benchmarks/exercism_practice.py
 ```
 Note you need the `-s` to see the printed results.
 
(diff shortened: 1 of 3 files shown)


PCSwingle (approved):
Looks good to me! I definitely agree that it's a bit messy right now; an idea I had for cleaning it up would be to make a BenchmarkRunner class and have subclasses for python and javascript that get selected by the --language option, and the subclasses would handle all of the language specific things. Feel free to merge now if you want to or work on that, up to you.

PCSwingle:
Also, I love the improved progress bar! I didn't even know you could do that!

jakethekoenig:
> Looks good to me! I definitely agree that it's a bit messy right now; an idea I had for cleaning it up would be to make a BenchmarkRunner class and have subclasses for python and javascript that get selected by the --language option, and the subclasses would handle all of the language specific things. Feel free to merge now if you want to or work on that, up to you.

I'll do that when we want to support one more language

jakethekoenig:
> Also, I love the improved progress bar! I didn't even know you could do that!

I didn't even know about tqdm until you added it, so thanks for that. gpt told me how to add a description.

PR #119, opened 2023-10-01
Generic interface

waydegg:
### Summary

This PR covers:
- updating all blocking code to be async
- creating a standard message format between the core library and client libraries
- moving all terminal-specific logic and dependencies under `mentat/terminal`

Things that will be covered in a later PR:
- A JSON RPC server with an LSP-compantible API that any editors with language server support can connect to 
- A `SessionManager` to coordinate multiple, concurrent chats

### To-do

- [x] merge with main and resolve merge conflicts
- [ ] ~~validate `cprint` colors received from core~~
- [x] check/test for race conditions b/w core and terminal client (specifically w/ inputs and interrupts)
- [x] fix test cases 
- [x] create existing test cases
  - [x] core library
  - [x] terminal client
- [x] fix misc typing errors


biobootloader:
Great work! I haven't gotten through it all yet, but one initial issue: if you run with too many files to fit in context, it crashes in an unclear way, with an exception printed above the file tree, and then at the end a "task is already stopped"

```
❯ python -m mentat --diff main
Main task for Session(6f05a5ca-2dfb-40f2-bd39-1dea11e50e51) threw an exception
  File "/Users/biobootloader/code/mentat/mentat/session.py", line 142, in run_main
    await self._main()
  File "/Users/biobootloader/code/mentat/mentat/session.py", line 81, in _main
    conv = await Conversation.create(
           ^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/biobootloader/code/mentat/mentat/conversation.py", line 93, in create
    raise Exception(
Exception: Included files already exceed token limit (43877 / 32768). Please try running again with a reduced number of files.
Files included in context:
mentat
├── docs
│   └── configuration.md
├── mentat
│   ├── __main__.py
│   ├── broadcast.py
│   ├── code_context.py
│   ├── code_edit_feedback.py
│   ├── code_file_manager.py
│   ├── code_map.py
│   ├── commands.py
│   ├── config_manager.py
│   ├── conversation.py
│   ├── diff_context.py
│   ├── errors.py
│   ├── llm_api.py
│   ├── logging_config.py
│   ├── parsers
│   │   ├── change_display_helper.py
│   │   ├── file_edit.py
│   │   └── parser.py
│   ├── session.py
│   ├── session_input.py
│   ├── session_stream.py
│   └── terminal
│       ├── __init__.py
│       ├── client.py
│       ├── output.py
│       ├── prompt_completer.py
│       └── prompt_session.py
├── requirements.txt
├── setup.py
└── tests
    ├── benchmarks
    │   └── exercism_practice_python.py
    ├── broadcast_test.py
    ├── code_file_manager_test.py
    └── conftest.py
Diff annotations:
 ─•─ Branch main: b5785106: Exercism Benchmark supports javascript (#116) | 34 files | 1858 lines


Excluding CodeMap from system message.
Reason: not enough tokens available in model context.
Task is already stopped

```

biobootloader:
`CostTracker.display_total_cost` is never called (it's supposed to be called whenever the session ends)

https://github.com/AbanteAI/mentat/blob/50388cee93245a90a37b879b0af2940c06d05fdd/mentat/llm_api.py#L190-L194

mentat/code_map.py
```diff
@@ -25,17 +32,32 @@ def _get_code_map(root: Path, file_path: Path, exclude_signatures: bool = False)
...
+        preexec_fn=os.setpgrp,
+    )
+    stdout, stderr = await process.communicate()
+
+    if process.returncode == 0:
+        output = stdout.decode("utf-8").strip() if stdout else ""
+        output_lines = output.splitlines()
+    else:
+        error_output = stderr.decode("utf-8").strip() if stderr else ""
+        raise Exception(f"Command failed with error: {error_output}")
```
biobootloader, line 49: should we make a helper function that runs async subprocess calls?

waydegg:
>if you run with too many files to fit in context, it crashes in an unclear way, with an exception printed above the file tree, and then at the end a "task is already stopped"

Just pushed a fix for this

waydegg:
Just reverted all the changes to `StreamingPrinter` and most of the code in `display_change_helper.py`. We can process the escape codes on the client directly (meaning we can leave termcolor `colored` calls in) which is less code changes. Eventually we should change the way we send messages with `StreamingPrinter` since it's not very optimized as it's written for terminal clients, but I think it's ok to leave as-is for now.

pyrightconfig.json
```diff
@@ -1 +1,4 @@
-{"include": ["mentat"], "ignore": ["testbed", "tests"], "typeCheckingMode": "strict"}
+{
+  "include": ["mentat", "tests", "testbed"],
+  "strict": ["mentat"]
```
waydegg, line 3: I don't think we should disable all type checking for tests/testbed modules, so I re-enabled it

mentat/streaming_printer.py
waydegg, line 1: There's a race condition where an extra character/message may be sent to the `SessionStream` just after an interrupt message is received.

mentat/session_stream.py
```diff
@@ -0,0 +1,99 @@
...
+
+    Stores message history for a Session and holds an in-memory message bus.
+
+    Terminal and extension clients can read these messages and render them accordingly.
+    For the terminal, they would be rendered with `cprint`.
+    """
+
+    def __init__(self):
+        self.messages: List[StreamMessage] = []
+        self.interrupt_lock = asyncio.Lock()
```
waydegg, line 53: This feels janky, will come up with a better way to handle interrupts from clients later.

mentat/session_stream.py
```diff
@@ -0,0 +1,99 @@
...
+
+_SESSION_STREAM: ContextVar[SessionStream] = ContextVar("mentat:session_stream")
+
+
+def get_session_stream():
+    return _SESSION_STREAM.get()
+
+
+def set_session_stream(session_stream: SessionStream):
+    return _SESSION_STREAM.set(session_stream)
```
waydegg, line 24: Probably can remove these get/set methods and just make `_SESSION_STREAM` a public variable

mentat/session_input.py
```diff
@@ -0,0 +1,80 @@
...
+        # TODO: combine this into a single message (include content)
+        await stream.send("(Y/n)" if default_yes else "(y/N)")
+        response = await collect_user_input()
+        content = response.data
+        if content in ["y", "n", ""]:
+            break
+    return content == "y" or (content != "n" and default_yes)
+
+
+async def listen_for_interrupt(
```
waydegg, line 39: This isn't really used right now (it will be used once we drop `StreamingPrinter`)

mentat/terminal/output.py
```diff
@@ -0,0 +1,41 @@
...
+    color: str | None = None,
+    flush: bool = False,
+):
+    if color:
+        f_color = color.replace("_", "").replace("light", "bright")
+        if f_color != "" and not f_color.startswith("ansi"):
+            f_color = "ansi" + f_color
+        print_formatted_text(FormattedText([(f_color, content)]), end=end, flush=flush)
+    else:
+        print(content, end=end, flush=True)
```
waydegg, line 21: Builtin `print` properly renders the color codes added by termcolor's `colored`, wasn't able to get the rendering working for `print_formatted_text`. Once we move away from using termcolor on the core library this code will be updated.

mentat/code_map.py
waydegg: yea that's a good idea

requirements.txt
```diff
@@ -14,6 +14,7 @@ gitdb==4.0.10
 GitPython==3.1.35
 idna==3.4
 iniconfig==2.0.0
+ipdb==0.13.13
```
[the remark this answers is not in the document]
jakethekoenig: I added it because it was used in some files, but I guess that was a temporary development thing and it probably shouldn't be in the requirements. I was thinking we should probably split the requirements into `requirements.txt` and `requirements-test.txt` as there are a lot of requirements that are only relevant for development: pytest*, ruff, pyright, black, isort and maybe others.

pyrightconfig.json
PCSwingle: I think we should keep it disabled; for the testbed, we want to be able to make tests that don't have typing in them, and for the regular tests, it will just add the hassle of specifying the type of fixtures as argument functions.

tests/conftest.py
```diff
@@ -88,39 +83,29 @@ async def async_generator():
...
-    return mock_method
+    return mock
 
 
-@pytest.fixture
+@pytest.fixture(scope="function")
 def mock_setup_api_key(mocker):
-    mocker.patch("mentat.app.setup_api_key")
-    mocker.patch("mentat.conversation.is_model_available")
+    mocker.patch("mentat.llm_api.setup_api_key")
```
PCSwingle, line 92: This is wrong; mocking works by mocking the IMPORT of the function, not the actual function itself (which is extremely counterintuitive, but that's how it is), so you need to mock it in the file that it's imported/used in; in this case, mentat.session.setup_api_key, and mentat.conversation.is_model_available

mentat/session.py
```diff
@@ -0,0 +1,215 @@
...
+        diff: Optional[str] = None,
+        pr_diff: Optional[str] = None,
+    ):
+        stream = SessionStream()
+        await stream.start()
+        set_session_stream(stream)
+
+        git_root = get_shared_git_root_for_paths([Path(path) for path in paths])
+        # TODO: Config should be created in the client (i.e., to get vscode settings) and passed to session
+        config = await ConfigManager.create(git_root)
```
biobootloader, line 72:
```suggestion
        git_root = get_shared_git_root_for_paths([Path(path) for path in paths])
        # TODO: Config should be created in the client (i.e., to get vscode settings) and passed to session
        # TODO: calculate git_root in Config creation?
        config = await ConfigManager.create(git_root)
```

PCSwingle: I don't think it makes sense for git root to be in the config in the first place; ideally we move it out sometime soon after this PR (when we change how we pass around all of the other singletons)

jakethekoenig:
Sorry I picked the wrong day to say I'd stream. The exercism benchmark is very close to done. It works occasionally but I get an error like this a lot:
```
ERROR:root:Main task for Session(bf2942b1-1daf-4289-b469-bc6a90559bd9) threw an exception
  File "/home/jake/Development/mentat/mentat/session.py", line 159, in run_main
    await self._main()
  File "/home/jake/Development/mentat/mentat/session.py", line 123, in _main
    file_edits = await conv.get_model_response(self.parser, self.config)
  File "/home/jake/Development/mentat/mentat/conversation.py", line 162, in get_model_response
    message, file_edits, time_elapsed = await self._stream_model_response(
  File "/home/jake/Development/mentat/mentat/conversation.py", line 136, in _stream_model_response
    async with parser.interrupt_catcher():
  File "/usr/lib/python3.10/contextlib.py", line 206, in __aexit__
    await anext(self.gen)
  File "/home/jake/Development/mentat/mentat/parsers/parser.py", line 48, in interrupt_catcher
    self._interrupt_task.cancel()
AttributeError: 'NoneType' object has no attribute 'cancel'
```
Also I need to figure out how to turn off logging.

jakethekoenig:
> Sorry I picked the wrong day to say I'd stream. The exercism benchmark is very close to done. It works occasionally but I get an error like this a lot:
> 
> ```
> ERROR:root:Main task for Session(bf2942b1-1daf-4289-b469-bc6a90559bd9) threw an exception
> Traceback (most recent call last):
>   File "/home/jake/Development/mentat/mentat/session.py", line 159, in run_main
>     await self._main()
>   File "/home/jake/Development/mentat/mentat/session.py", line 123, in _main
>     file_edits = await conv.get_model_response(self.parser, self.config)
>   File "/home/jake/Development/mentat/mentat/conversation.py", line 162, in get_model_response
>     message, file_edits, time_elapsed = await self._stream_model_response(
>   File "/home/jake/Development/mentat/mentat/conversation.py", line 136, in _stream_model_response
>     async with parser.interrupt_catcher():
>   File "/usr/lib/python3.10/contextlib.py", line 206, in __aexit__
>     await anext(self.gen)
>   File "/home/jake/Development/mentat/mentat/parsers/parser.py", line 48, in interrupt_catcher
>     self._interrupt_task.cancel()
> AttributeError: 'NoneType' object has no attribute 'cancel'
> ```
> 
> Also I need to figure out how to turn off logging.

Maybe someone else could look at it too. I'm not sure whether I'm using session wrong or this is uncovering a bug in session

PCSwingle:
Just finished looking through all of the comments on this PR and these are the things I see us needing to discuss/do (don't necessarily have to do all of them before merging):

- [ ] Separate requirements.txt into requirements.txt and dev-requirements.txt
- [ ] Delete extra exercism_practice_python.py file? @jakethekoenig is it confirmed a duplicate file we should get rid of?
- [ ] Windows doesn't have os.setpgrp (@waydegg is looking into this rn)
- [ ] @waydegg, do you still think we should remove the try/except in client.py _run? I think we should but I'll let y'all decide
- [ ] Make _SESSION_STREAM variable public
- [ ] Add tests for the terminal client

We should also probably all be doing testing searching for bugs and race conditions; I think I cleared out all of the big bugs already, so hopefully there aren't many left!

mentat/code_context.py
```diff
@@ -186,15 +195,15 @@ def refresh(self, replace_paths: bool | None = False):
             if replace_paths and not self.settings.paths:
                 self.settings.paths = self.diff_context.files
         except UserError as e:
-            cprint(str(e), "light_yellow")
-            exit()
+            await stream.send(str(e), color="light_yellow")
+            exit()  # NOTE: should this be an exit()?
```
granawkins: I was copying the pattern from `expand_paths` - which I see still exists but is now in `mentat/terminal/client.py`. Anyway it's not functioning correctly - if you give it an invalid diff, right now it just closes. 

I replaced the call to `exit()` this with a MentatError, and move the initialization of `CodeContext` to `Session._main`, right before Conversation. Now it gives you an informative message and then closes.

jakethekoenig:
> * [ ]  Delete extra exercism_practice_python.py file? @jakethekoenig is it confirmed a duplicate file we should get rid of?

Yep, just deleted it.

jakethekoenig:
The exercism benchmark is good to go. I think it looks a lot cleaner interacting with mentat through sessions. Because things aren't done in mocks there's no longer any need to store things in threadLocal and all the interactions are just in the one function. I fixed the exception I was running into by just checking if `_Interrupt_task` is `None` before canceling it.

PCSwingle:
I marked the tests for terminalclient as complete; there are only 2 tests, but I couldn't think of anything else that's really specific to the terminal client; feel free to add something if you have any ideas! I was trying to find a way to test ctrl-c/ctrl-d quitting, but couldn't figure out how (since it's a keybind for PromptToolkit)

waydegg:
> validate cprint colors received from core

I'll add types for colors later. We have an `extra` field in `StreamMessage` that can have "reserved" keywords (like `color`) which we can add types to.

PCSwingle (approved):
:rocket:

jakethekoenig (approved):

biobootloader (approved):
🍾 

Great work @wayde and everyone who helped the last couple days!

I've created an issue to collect the various followups mentioned in comments here, let's move things there so they don't get forgotten or lost here: https://github.com/AbanteAI/mentat/issues/121

granawkins (approved):
Great work!

PR #120, opened 2023-10-01
Full gpt transcripts logged

jakethekoenig:
This is a little redundant because the user interactions and LLM responses are logged elsewhere but:
1. Going forward I don't think we should assume we can reconstruct the conversational state from user interactions and LLM responses. We may add features like summarization of code edits to save tokens or not send user requests verbatim (this already happens to some extent with special tokens like q.
2. It seems like a nice QoL feature to log the transcripts in the exact format needed for interaction with OpenAI apis.

### DIFF
--- a/mentat/logging_config.py
+++ b/mentat/logging_config.py
@@ -43,7 +43,7 @@ def setup_logging():
 
     transcripts_logger = logging.getLogger("transcript")
     transcripts_formatter = logging.Formatter("%(message)s")
-    transcripts_handler = logging.FileHandler(logs_path / "transcripts.log")
+    transcripts_handler = logging.FileHandler(logs_path / f"transcript_{timestamp}.log")
     transcripts_handler.setFormatter(transcripts_formatter)
     transcripts_logger.addHandler(transcripts_handler)
     transcripts_logger.setLevel(logging.INFO)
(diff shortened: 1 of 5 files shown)


PCSwingle (approved):
Thanks for adding this! This is definitely something I think we'll find super useful

mentat/logging_config.py
```diff
@@ -43,6 +43,14 @@ def setup_logging():
     costs_logger.setLevel(logging.INFO)
     costs_logger.propagate = False
 
+    transcripts_logger = logging.getLogger("transcript")
+    transcripts_formatter = logging.Formatter("%(message)s")
+    transcripts_handler = logging.FileHandler(logs_path / "transcripts.log")
```
granawkins, line 48:
```suggestion
    transcripts_handler = logging.FileHandler(logs_path / f"transcript_{timestamp}.log")
```

My one suggestion - makes it easier to review specific logs for debugging.

granawkins (changes requested):

biobootloader:
Nice, makes sense to have a verbatim log instead of having to try to piece things together.

Is there a reason people would want to turn this off / not turn it on? Maybe we don't need an option

jakethekoenig:
> Nice, makes sense to have a verbatim log instead of having to try to piece things together.
> 
> Is there a reason people would want to turn this off / not turn it on? Maybe we don't need an option

I was thinking the transcripts would grow quadratically and it'd maybe be better to not include them by default but I guess they grow quadratically in number of conversational steps which should be pretty small. I think you're right there's no need to have this off by default so there's no need to have the setting.

PR #123, opened 2023-10-04
turn all singletons into context vars, remove git_root from config

PCSwingle:
Moves git root out of config and turns all of the singletons into contextvars, instead of passing them around everywhere. 

Couple of notes:

I removed the 'self.config', 'self.code_file_manager', etc. references everywhere except for the Session still having self.stream; we can change that in the future if we need to, but with the upcoming RFC server change, I decided to wait until we know what it's all going to look like.

Do you think we should have CodeFileManager, CodeContext, and Conversation be regular instances? We might want to have multiple of them in the future. On the other hand, I think it makes a lot more sense in our current codebase for them to be singletons. Let me know what y'all think!

### DIFF
--- a/mentat/code_context.py
+++ b/mentat/code_context.py
@@ -215,7 +215,7 @@ class CodeContext:
         await stream.send(git_root.name, color="blue")
         await _print_path_tree(
             _build_path_tree(list(self.files.values()), git_root),
-            get_paths_with_git_diffs(git_root),
+            get_paths_with_git_diffs(),
             git_root,
         )
         await self.diff_context.display_context()
--- a/mentat/code_context.py
+++ b/mentat/code_context.py
@@ -3,25 +3,21 @@ from __future__ import annotations
 import glob
 import logging
 import os
+from contextvars import ContextVar
 from pathlib import Path
 from textwrap import dedent
-from typing import TYPE_CHECKING, Any, Dict, Optional
+from typing import Any, Dict, Optional
 
 import attr
 
 from .code_file import CodeFile
 from .code_map import CodeMap
-from .config_manager import ConfigManager
-from .diff_context import get_diff_context
+from .config_manager import CONFIG_MANAGER, ConfigManager
+from .diff_context import DiffContext
 from .errors import MentatError, UserError
-from .git_handler import get_non_gitignored_files, get_paths_with_git_diffs
+from .git_handler import GIT_ROOT, get_non_gitignored_files, get_paths_with_git_diffs
 from .llm_api import count_tokens, model_context_size
 from .session_stream import SESSION_STREAM
-
-if TYPE_CHECKING:
-    # This normally will cause a circular import
-    from mentat.code_file_manager import CodeFileManager
-    from mentat.parsers.parser import Parser
 
 
 def _is_file_text_encoded(file_path: Path):
@@ -67,7 +63,7 @@ def _abs_files_from_list(paths: list[Path], check_for_text: bool = True):
 
 
 def _get_files(
-    config: ConfigManager, paths: list[Path], exclude_paths: list[Path]
+    config: ConfigManager, git_root: Path, paths: list[Path], exclude_paths: list[Path]
 ) -> Dict[Path, CodeFile]:
     excluded_files_direct, excluded_files_from_dirs = _abs_files_from_list(
         exclude_paths, check_for_text=False
@@ -77,12 +73,12 @@ def _get_files(
     ), set(map(lambda f: f.path, excluded_files_from_dirs))
 
     glob_excluded_files = set(
-        os.path.join(config.git_root, file)
+        os.path.join(git_root, file)
         for glob_path in config.file_exclude_glob_list()
         # If the user puts a / at the beginning, it will try to look in root directory
         for file in glob.glob(
             pathname=glob_path,
-            root_dir=config.git_root,
+            root_dir=git_root,
             recursive=True,
         )
     )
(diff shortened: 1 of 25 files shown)


waydegg:
> Do you think we should have CodeFileManager, CodeContext, and Conversation be regular instances?

By regular instances do you mean passing them around normally and not fetching them from a Contextvar?

mentat/code_file_manager.py
```diff
@@ -2,33 +2,47 @@
...
+    from .parsers.file_edit import FileEdit
+
+CODE_FILE_MANAGER: ContextVar[CodeFileManager] = ContextVar("mentat:code_file_manager")
+
 
 class CodeFileManager:
-    def __init__(self, config: ConfigManager):
-        self.config = config
+    def __init__(self):
+        pass
```
waydegg, line 26: We can remove this method imo

mentat/conversation.py
```diff
@@ -149,20 +136,26 @@ async def _stream_model_response(
...
 
-    async def get_model_response(
-        self, parser: Parser, config: ConfigManager
-    ) -> list[FileEdit]:
+    async def get_model_response(self) -> list[FileEdit]:
+        stream = SESSION_STREAM.get()
+        code_context = CODE_CONTEXT.get()
+        cost_tracker = COST_TRACKER.get()
+        code_file_manager = CODE_FILE_MANAGER.get()
+        parser = PARSER.get()
```
waydegg, line 144: It's pretty common we need to `.get` multiple object instances stored in Contextvars in a lot of methods/functions, so I'm wondering if it would be better to have a single Contextvar (e.g. a `SessionContext`) that's just a dict holding the instances for `SessionStream`, `CostTracker`, `Parser`, etc.). Thoughts?

It's kind of like having a single state object in React you just pass around everywhere instead of having many state variables.

waydegg: All of these instances are tied to a specific session anyways

mentat/code_context.py
```diff
@@ -220,24 +205,29 @@ async def refresh(self, replace_paths: bool | None = False):
...
+            _build_path_tree(list(self.files.values()), git_root),
+            get_paths_with_git_diffs(git_root),
+            git_root,
         )
         await self.diff_context.display_context()
 
     async def get_code_message(
-        self, model: str, code_file_manager: CodeFileManager, parser: Parser
+        self,
+        all_file_lines: dict[Path, list[str]],
```
granawkins, line 225: In the `context-features` pr, the code_file_manager gets passed from here all the way down to the `CodeFile._get_file_message`. We can just use CODE_FILE_MANAGER.get() in that function and save a lot of ink.

mentat/conversation.py
granawkins: Ya that makes sense, lots of things tied to the session. I was also thinking this reminded me somehow of react.

mentat/git_handler.py
```diff
@@ -1,10 +1,13 @@
 import logging
 import os
 import subprocess
+from contextvars import ContextVar
 from pathlib import Path
 
 from mentat.errors import UserError
 
+GIT_ROOT: ContextVar[Path] = ContextVar("mentat:git_root")
```
granawkins, line 9: Should we use GIT_ROOT directly in the functions below?

mentat/session.py
```diff
@@ -59,44 +51,54 @@ async def create(
...
-        self = Session(stream, config, parser, code_context_settings)
+        # NOTE: Should codefilemanager, codecontext, and conversation be contextvars/singletons or regular instances?
+        code_file_manager = CodeFileManager()
+        CODE_FILE_MANAGER.set(code_file_manager)
 
+        conversation = Conversation()
+        CONVERSATION.set(conversation)
+
+        self = Session(stream)
         return self
```
granawkins, line 86:
```suggestion
        return cls(stream)
```

mentat/terminal/prompt_completer.py
```diff
@@ -67,7 +65,9 @@ def refresh_completions_for_file_path(self, file_path: Path):
         self.syntax_completions[file_path] = SyntaxCompletion(words=filtered_tokens)
 
     async def refresh_completions(self):
-        file_paths = list(self.session.code_context.files.keys())
+        code_context = CODE_CONTEXT.get()
+
+        file_paths = list(code_context.files.keys())
```
granawkins, line 70:
```suggestion
        file_paths = list(CODE_CONTEXT.get().files.keys())
```
I see this pattern a few places in the code, makes sense to use it if there's only one use.

mentat/conversation.py
```diff
@@ -70,9 +60,10 @@ async def create(
         prompt = parser.get_system_prompt()
         self.add_system_message(prompt)
```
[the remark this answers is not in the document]
PCSwingle: Good catch!

mentat/git_handler.py
PCSwingle: Yes, thanks for finding this! I probably missed a few other places I should have used the contextvar haha

mentat/terminal/prompt_completer.py
PCSwingle: I think I would prefer to as a rule get the contextvars at the start of the function; also, if we end up merging all of the contextvars, we'll have to get it at the start anyways.

mentat/conversation.py
PCSwingle: Hmm, I wouldn't be opposed to it; I like the idea of a SessionContext class, but on the other hand I feel like it bundles up a lot of unrelated stuff for not much reason. How about we keep it like this for now (since it's just a couple of extra lines of code), and if we want to in the future we can move to it?

biobootloader: I like the `SessionContext` idea, it'd make it explicit which things are part of the state

biobootloader:
I'll let @waydegg and/or @granawkins finish reviewing this!

mentat/conversation.py
waydegg: @PCSwingle which unrelated stuff would be bundled up for no reason?

waydegg: By having a bunch of different context vars we're passing around it feels like we'd just me moving the issue of passing a bunch of object instances around to having to make a bunch of `SOME_CONTEXT_VAR.get()` calls in every function.

waydegg: In most cases functions won't need to access everything in the context, but I think it's a fine tradeoff and makes the codebase simpler.

PCSwingle: Ok, I'll add in the SessionContext; although I can't say I'm 100% convinced haha

granawkins (approved):

PR #124, opened 2023-10-04
add util func for async subprocess

waydegg:
This PR adds a function that wraps async subprocess calls. 

I don't think we should make all the synchronous subprocesses calls async here, better to do that separately imo.


granawkins (approved):
Just changing the name to `run_subprocess_async` for clarity

PR #126, opened 2023-10-05
remove basicConfig and modify the root handler directly

PCSwingle:
Fixes some problems with the current logging setup we had

### DIFF
--- a/mentat/session.py
+++ b/mentat/session.py
@@ -6,6 +6,8 @@ from pathlib import Path
 from typing import List, Optional, Union, cast
 from uuid import uuid4
 
+from mentat.logging_config import setup_logging
+
 from .code_context import CodeContext, CodeContextSettings
 from .code_edit_feedback import get_user_feedback_on_edits
 from .code_file_manager import CodeFileManager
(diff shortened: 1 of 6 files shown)


mentat/session.py
```diff
@@ -148,6 +150,8 @@ def start(self) -> asyncio.Task[None]:
             logging.warning("Job already started")
             return self._main_task
 
+        setup_logging()
```
jakethekoenig, line 153: When I run the exercism benchmark on two exercises I see two copies of both exercises in transcripts. I think because setup_logging is run twice. Is there a way to make it idempotent?

PCSwingle:
Ah, that makes sense; it's because the costs and transcripts loggers aren't having old handlers removed. We can either a. call the logging setup in the client or b. remove old handlers in the logging setup. I think I prefer b. since it feels like logging should be handled by the server. I'll get that done!

PCSwingle:
I tested it and I think I fixed it, but let me know if it's still happening

jakethekoenig (approved):
Thanks so much!

PR #127, opened 2023-10-05
Exercism fix

jakethekoenig:
We were not awaiting input requested in the right place so we could sometimes run the tests before mentat had edited the files.


PCSwingle (approved):
Thanks, looks good!

PR #128, opened 2023-10-06
Add non git project support

waydegg:
This PR allows Mentat to start in non-git managed directories

### DIFF
--- a/mentat/terminal/client.py
+++ b/mentat/terminal/client.py
@@ -2,6 +2,7 @@ import argparse
 import asyncio
 import glob
 import logging
+import os
 import signal
 from pathlib import Path
 from typing import Any, Coroutine, List, Set
@@ -23,12 +24,14 @@ from mentat.terminal.prompt_session import MentatPromptSession
 class TerminalClient:
     def __init__(
         self,
+        cwd: Path,
         paths: List[Path] = [],
         exclude_paths: List[Path] = [],
         no_code_map: bool = False,
         diff: str | None = None,
         pr_diff: str | None = None,
     ):
+        self.cwd = cwd
         self.paths = paths
         self.exclude_paths = exclude_paths
         self.no_code_map = no_code_map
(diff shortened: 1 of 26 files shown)


tests/conftest.py
```diff
@@ -193,25 +200,28 @@ def run_git_command(cwd, *args):
 
 
 @pytest.fixture(autouse=True)
-def temp_testbed(monkeypatch):
+def temp_testbed(request, monkeypatch):
+    use_git = request.param if hasattr(request, "param") else True
```
waydegg, line 204: Maybe should add typing to `request`?

tests/code_file_manager_test.py
```diff
@@ -112,9 +121,14 @@ async def test_run_from_subdirectory(
         }
         @@code
         # Hello
-        @@end""")])
+        @@end"""
+            )
+        ]
+    )
```
waydegg, line 127: Black re-formatted a lot of strings/code in the tests as it doesn't look like we run any formatters in gh actions. Should we update github actions?

mentat/terminal/client.py
```diff
@@ -23,12 +24,14 @@
 class TerminalClient:
     def __init__(
         self,
+        cwd: Path,
```
waydegg, line 27: I made `cwd` explicit here since we will be running the "core" library on a separate process for extensions on some unknown filesystem, meaning we can use `os.getcwd` as a default

granawkins (commented):
Do we need to track both `cwd` and `git_root`? Would it be simpler to just track the `root`, and objects that refer to git like `DiffContext` can check if it's available with like `config.is_git() -> bool`.

tests/code_file_manager_test.py
PCSwingle: I'm pretty sure we do run black and isort in github actions? Maybe black got updated? Or maybe we're not running black and isort on the tests directory? I'll look into this

PR #129, opened 2023-10-06
A recursive FileNode for CodeFileManager

granawkins:
I have mentioned this in a couple comments so here is a start. I'm imagining:
- `CodeFileManager.root` is a Node for the git_root. This can be built and cached on startup.
- `CodeFile` and `FileEdit` can get the node via `CODE_FILE_MANAGER.get().root[path]`, and then use it for thread-safe read/write. We'll set CFM outside/above the session loops.
- Everything in `include_files.py` can be done here much easier.

So this is here as a draft for feedback. Planning to build it out after #115.

### DIFF
--- a/mentat/file_node.py
+++ b/mentat/file_node.py
@@ -1,0 +1,92 @@
+from __future__ import annotations
+
+import asyncio
+from pathlib import Path
+from typing import Generator
+
+from mentat.errors import MentatError
+from mentat.git_handler import get_non_gitignored_files
+from mentat.utils import sha256
+
+
+class Node:
+    def __init__(self, path: Path, parent: Node | None = None):
+        self.path = path.resolve()
+        self.parent = parent
+        self.children = dict[str, Node]()
+        self.lock = asyncio.Lock()
+        self.refresh()
+
+    def __getitem__(self, path: Path | str) -> "Node":
+        if isinstance(path, str):
+            path = Path(path)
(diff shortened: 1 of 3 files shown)


waydegg:
> thread-safe read/write. We'll set CFM outside/above the session loops

Instead of putting file read/writes on different threads, I'm thinking we could use [aiofiles](https://pypi.org/project/aiofiles/) and keep everything in the async event loop

waydegg:
What's the advantage of using this over the existing way we pull in files/partial files again?

granawkins:
> What's the advantage of using this over the existing way we pull in files/partial files again?

> Instead of putting file read/writes on different threads, I'm thinking we could use [aiofiles](https://pypi.org/project/aiofiles/) and keep everything in the async event loop

Keeping a map of the filetree in a separate thread will have a lot of uses, one of which is storing the Lock for that file. I think we'd still need the lock with `aiofiles`, right? 

Another is storing the checksum for each file. We'll need to check for updates across hundreds of files each time the user sends a message, and hopefully not read/hash every time. We could use git, but we also want to #128. If we base it on the last-updated time, we can store the expected time along with the checksum here.

We can also use it in code_context.get_code_message to iterate over the context, instead of calling `get_non_gitignored_files`.

What do you think? Premature optimization?

granawkins:
Hasn't been necessary

PR #134, opened 2023-10-09
Python client added

jakethekoenig:
A light client for mentat is created and used in the exercism test. I think we'll mostly find value in this for testing but I could imagine people wanting to interact with mentat programmatically. For instance maybe you want to give GPT-4 access to mentat.

### DIFF
--- a/mentat/python_client/client.py
+++ b/mentat/python_client/client.py
@@ -27,11 +27,11 @@ class PythonClient:
         self.acc_task: asyncio.Task[None] | None = None
         self._started = False
         self._accumulated_message = ""
-        self._input_request_message = None
 
     async def call_mentat(self, message: str):
         if not self._started:
             await self._startup()
+        assert isinstance(self.session, Session), "Client is not running"
         await self.session.stream.send(
             message,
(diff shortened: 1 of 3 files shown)


mentat/python_client/client.py
```diff
@@ -0,0 +1,71 @@
...
+            self._accumulated_message += message.data
+
+    async def _startup(self):
+        self.session = await Session.create(
+            self.paths, self.exclude_paths, self.no_code_map, self.diff, self.pr_diff
+        )
+        asyncio.ensure_future(self.session.start())
+        self._input_request_message = await self.session.stream.recv("input_request")
+
+        self.acc_task = asyncio.create_task(self._accumulate_messages())
```
granawkins, line 61: It would be useful to have access to the initial message, i.e. the context summary and user prompt. Moving this line before _input_req catches it, and then it's included in the first message by default.

granawkins (commented):
I like this a lot, will make benchmarking very simple.

PCSwingle (approved):
This looks super awesome, thanks for making this!

PR #136, opened 2023-10-10
Clojure exercism added

jakethekoenig:


PCSwingle (approved):
I really like how nice the exercise runner is and how easy it makes it to change languages!

PR #137, opened 2023-10-10
Update configuration.md

jakethekoenig:


PCSwingle (approved):
Looks good! I'm going to modify the ConfigManager real quick to automatically cast strings to ints, and then we can merge this

PR #138, opened 2023-10-10
use signal.signal instead of loop.add_signal_handler

PCSwingle:
Windows doesn't have signals the same way *nix systems do, so asyncio's loop.add_signal_handler isn't implemented in Windows. The main difference between signal.signal and add_signal_handler is that add_signal_handler is asyncio safe, while signal.signal isn't; when the SIG_INT gets caught by signal.signal now, we could be in the middle of an async coroutine and it would be interrupted. I don't think this causes any race conditions; all we do is send an interrupt signal through the sessionstream, but I would like if @waydegg looked over this and gave his thoughts before merging.

Additionally, the reason this wasn't caught by testing is because we weren't running any TerminalClient tests on Windows (due to an issue I discovered a while ago involving PromptToolkit and Github Actions Windows runners specifically); before we get this merged, I'm going to look into resolving that problem in a different way so that we can still run these tests on Windows.

### DIFF
--- a/mentat/terminal/client.py
+++ b/mentat/terminal/client.py
@@ -89,6 +89,9 @@ class TerminalClient:
             "", source=StreamMessageSource.CLIENT, channel="interrupt"
         )
 
+    # Be careful editing this function; since we use signal.signal instead of asyncio's
+    # add signal handler (which isn't available on Windows), this function can interrupt
+    # asyncio coroutines, potentially causing race conditions.
     def _handle_exit(self, sig: int, frame: FrameType | None):
         assert isinstance(self.session, Session), "TerminalClient is not running"
         if (
--- a/mentat/terminal/client.py
+++ b/mentat/terminal/client.py
@@ -3,6 +3,7 @@ import asyncio
 import logging
 import signal
 from pathlib import Path
+from types import FrameType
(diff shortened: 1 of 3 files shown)


waydegg:
> the reason this wasn't caught by testing is because we weren't running any TerminalClient tests on Windows

Was this disabled explicitly or were the tests not running b/c of issues with Prompt Toolkit and GHA?

PCSwingle:
> > the reason this wasn't caught by testing is because we weren't running any TerminalClient tests on Windows
> 
> Was this disabled explicitly or were the tests not running b/c of issues with Prompt Toolkit and GHA?

It was disabled explicitly, yeah. I'm working on changing the tests to just mock PromptToolkit while still running the tests

waydegg:
> I don't think this causes any race conditions

I don't think so either. The shared state we're modifying outside the asyncio event loop is ok to modify:
- `self.should_exit` and `self._force_exit` always start out as `False` and once toggled to `True` will stay `True`
- `asyncio.create_task` enqueues another task that won't run until the handler finishes (i.e. the asyncio event loop takes over)

We'll have to update things if we have multiple event loops running or need to modify more complex state tho.

waydegg (approved):

PR #142, opened 2023-10-11
Version bump to 0.1.16

jakethekoenig:
Our requirements cannot be read properly by brew because we read it with a relative path in setup.py. I made a few other small changes in setup.py which caused me to notice an import error in conversation.py so I fixed that too.


PCSwingle (approved):
:rocket:

PR #143, opened 2023-10-11
Add undo

PCSwingle:
Add undo and clear commands; refactor CodeContext, fix bugs

### DIFF
--- a/mentat/code_context.py
+++ b/mentat/code_context.py
@@ -65,12 +65,19 @@ class CodeContext:
     async def create(
         cls, paths: list[Path], exclude_paths: list[Path], settings: CodeContextSettings
     ):
+        stream = SESSION_STREAM.get()
+
         self = cls(settings)
 
         self.diff_context = await DiffContext.create(
             self.settings.diff, self.settings.pr_diff
         )
-        self.include_files = get_include_files(paths, exclude_paths)
+        self.include_files, invalid_paths = get_include_files(paths, exclude_paths)
+        for invalid_path in invalid_paths:
+            await stream.send(
+                f"File path {invalid_path} is not text encoded, and was skipped.",
+                color="light_yellow",
+            )
         await self._set_code_map()
 
         return self
@@ -177,10 +184,6 @@ class CodeContext:
     ) -> str:
         code_message = list[str]()
 
-        # Reset the diff context since it could have changed since last code message
-        self.diff_context = await DiffContext.create(
-            self.settings.diff, self.settings.pr_diff
-        )
         await self._set_code_map()
         if self.diff_context.files:
             code_message += [
@@ -268,12 +271,15 @@ class CodeContext:
 
         return sorted(all_features, key=_feature_relative_path)
 
-    def include_file(self, abs_path: Path):
-        for new_path, new_file in get_include_files([abs_path], []).items():
+    def include_file(self, path: Path):
+        paths, invalid_paths = get_include_files([path], [])
+        for new_path, new_file in paths.items():
             if new_path not in self.include_files:
                 self.include_files[new_path] = new_file
+        return invalid_paths
 
-    def exclude_file(self, abs_path: Path):
-        for new_path in get_include_files([abs_path], []).keys():
+    def exclude_file(self, path: Path):
+        paths, _ = get_include_files([path], [])
+        for new_path in paths.keys():
             if new_path in self.include_files:
                 del self.include_files[new_path]
--- a/mentat/code_context.py
+++ b/mentat/code_context.py
@@ -276,38 +276,31 @@ class CodeContext:
 
         return sorted(all_features, key=_feature_relative_path)
 
-    async def include_file(self, code_file: CodeFile):
+    async def include_file(self, abs_path: Path):
         stream = SESSION_STREAM.get()
-        if not os.path.exists(code_file.path):
-            await stream.send(f"File does not exist: {code_file.path}\n", color="red")
-            return
-        if code_file.path in self.settings.paths:
-            await stream.send(
-                f"File already in context: {code_file.path}\n", color="yellow"
-            )
-            return
-        if code_file.path in self.settings.exclude_paths:
-            self.settings.exclude_paths.remove(code_file.path)
-        self.settings.paths.append(code_file.path)
-        self._set_include_files()
-        await stream.send(
-            f"File included in context: {code_file.path}\n", color="green"
-        )
 
-    async def exclude_file(self, code_file: CodeFile):
-        stream = SESSION_STREAM.get()
-        if not os.path.exists(code_file.path):
-            await stream.send(f"File does not exist: {code_file.path}\n", color="red")
+        if not os.path.exists(abs_path):
+            await stream.send(f"File does not exist: {abs_path}\n", color="red")
             return
-        if code_file.path not in self.settings.paths:
-            await stream.send(
-                f"File not in context: {code_file.path}\n", color="yellow"
-            )
+        if abs_path in self.settings.paths:
+            await stream.send(f"File already in context: {abs_path}\n", color="yellow")
             return
-        if code_file.path in self.settings.exclude_paths:
-            self.settings.paths.remove(code_file.path)
-        self.settings.exclude_paths.append(code_file.path)
+        if abs_path in self.settings.exclude_paths:
+            self.settings.exclude_paths.remove(abs_path)
+        self.settings.paths.append(abs_path)
         self._set_include_files()
-        await stream.send(
-            f"File removed from context: {code_file.path}\n", color="green"
-        )
+        await stream.send(f"File included in context: {abs_path}\n", color="green")
+
+    async def exclude_file(self, abs_path: Path):
+        stream = SESSION_STREAM.get()
+        if not os.path.exists(abs_path):
+            await stream.send(f"File does not exist: {abs_path}\n", color="red")
+            return
+        if abs_path not in self.settings.paths:
+            await stream.send(f"File not in context: {abs_path}\n", color="yellow")
+            return
+        if abs_path in self.settings.exclude_paths:
+            self.settings.paths.remove(abs_path)
+        self.settings.exclude_paths.append(abs_path)
+        self._set_include_files()
+        await stream.send(f"File removed from context: {abs_path}\n", color="green")
--- a/mentat/code_context.py
+++ b/mentat/code_context.py
@@ -160,6 +160,7 @@ class CodeContext:
             features_checksum = sha256("".join(feature_file_checksums))
         settings = attr.asdict(self.settings)
         settings["max_tokens"] = max_tokens
+        settings["include_files"] = self.include_files
         settings_checksum = sha256(str(settings))
         return features_checksum + settings_checksum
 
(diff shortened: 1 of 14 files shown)


mentat/code_file_manager.py
```diff
@@ -36,25 +43,32 @@ def read_file(self, path: Path) -> list[str]:
...
+        code_context.exclude_file(abs_path)
         abs_path.unlink()
 
+    def _rename_file(
+        self, code_context: CodeContext, abs_path: Path, new_abs_path: Path
+    ):
+        logging.info(f"Renaming file {abs_path} to {new_abs_path}")
+        os.rename(abs_path, new_abs_path)
+        code_context.include_file(new_abs_path)
+        code_context.exclude_file(abs_path)
```
waydegg, line 65: Maybe add a test for this method?

mentat/code_file_manager.py
```diff
@@ -86,14 +101,20 @@ async def write_changes_to_files(
                 )
                 if await ask_yes_no(default_yes=False):
                     await stream.send(f"Deleting {rel_path}...", color="red")
+                    # We use the current lines rather than the stored lines for undo
```
waydegg, line 104: What's the difference between "current lines" and "stored lines" here?

mentat/code_file_manager.py
```diff
@@ -114,13 +135,23 @@ async def write_changes_to_files(
...
+                )
+                self._rename_file(
+                    code_context, file_edit.file_path, file_edit.rename_file_path
+                )
                 file_edit.file_path = file_edit.rename_file_path
 
             new_lines = file_edit.get_updated_file_lines(stored_lines)
-            with open(file_edit.file_path, "w") as f:
-                f.write("\n".join(new_lines))
+            if new_lines != stored_lines:
```
granawkins, line 147: Shouldn't this always be the case?

mentat/diff_context.py
```diff
@@ -109,15 +111,22 @@ def files(self) -> list[Path]:
...
 
         if diff and pr_diff:
-            raise UserError("Cannot specify more than one type of diff.")
+            await stream.send(
+                "Cannot specify more than one type of diff. Disabling diff and"
+                " pr-diff.",
+                color="light_yellow",
+            )
+            diff = None
+            pr_diff = None
```
waydegg, line 129: Is it ok to not raise a `UserError` here anymore? Assuming the user mistakenly puts in both `diff` and `pr_diff` (and just meant to put in one or the other) idk if it makes sense to continue running and call `merge-base` later with no diffs.

mentat/include_files.py
```diff
@@ -85,27 +72,29 @@ def abs_files_from_list(paths: list[Path], check_for_text: bool = True):
...
-    return files_direct, files_from_dirs
+    files_from_dirs = set(CodeFile(path.resolve()) for path in file_paths_from_dirs)
+    return files_direct, files_from_dirs, invalid_paths
 
 
 def get_include_files(
     paths: list[Path], exclude_paths: list[Path]
-) -> Dict[Path, CodeFile]:
+) -> tuple[Dict[Path, CodeFile], list[str]]:
     """Returns a complete list of text files in a given set of include/exclude Paths."""
```
waydegg, line 82: I think we should keep the previous behavior of not returning the invalid paths, and just return a dictionary of `CodeFile` where all the code files with invalid paths are filtered out. We don't use the returned list of `invalid_paths` other than just sending them to the `SessionStream` (which can be done in this method) anyways.

waydegg: Nicer to just return 1 object and not a tuple of objects also imo

mentat/terminal/client.py
```diff
@@ -115,6 +117,10 @@ def _init_signal_handlers(self):
     async def _startup(self):
         assert self.session is None, "TerminalClient already running"
 
+        # TODO: Right now, everything sent in Session.create is never printed because
+        # the PromptSessions and listening tasks haven't been set up yet; once we move
+        # the config to be client-side, we can get the config without setting up the
+        # session first, and we can move the session last
```
waydegg, line 123: I think we should update `broadcast.py` to handle this. A behavior that I think would be nice is having stream messages queued up when there's no listeners, then the second a new listener is "subscribed" the queue messages are copied over to the new subscriber queue which the subscriber (the Terminal Client in this case) can ingest and render.

Maybe make a github issue to pick this up later?

mentat/terminal/client.py
```diff
@@ -252,3 +258,39 @@ def run_cli():
         auto_tokens,
     )
     terminal_client.run()
+
+
+# This needs to stay in the terminal client since it's done on start and uses print/exit,
+# which is why we have an almost identical version in include_files
```
waydegg, line 264: Similar to [this comment I left](https://github.com/AbanteAI/mentat/pull/143/files#r1355908637), by copying over the queue messages to new subscriber queue's we can send "interrupt" messages before the Terminal Client is subscribed to the Session and have the shutdown work properly.

mentat/commands.py
```diff
@@ -162,3 +169,37 @@ def argument_names(cls) -> list[str]:
...
+    @classmethod
+    def argument_names(cls) -> list[str]:
+        return []
+
+    @classmethod
+    def help_message(cls) -> str:
+        return "Undo the last change made by Mentat"
+
+
+class ResetCommand(Command, command_name="reset"):
```
waydegg, line 191: My initial thought seeing this was that "reset" resets the context. What are your thoughts on having "undo-all" undo everything? More explicit and harder to mis-type

waydegg:
I don't think it's related to this PR since we haven't changed any of the prompts, but the `test_calculator_add_exp_then_log` and `test_calculator_continue_change` benchmark tests keep failing for me. Anyone else having this happen?

mentat/edit_history.py
```diff
@@ -0,0 +1,112 @@
...
+                f"File {self.old_file_name} already exists; unable to undo rename from"
+                f" {self.cur_file_name}",
+                color="light_red",
+            )
+        else:
+            os.rename(self.cur_file_name, self.old_file_name)
+
+
+@attr.define()
+class CreationAction:
```
granawkins, line 26: Maybe these can be `CreateAction` / `DeleteAction` to match `CreateFile`/`DeleteFile`

mentat/include_files.py
granawkins: +1

mentat/terminal/client.py
granawkins: Also like the queue idea

mentat/code_file_manager.py
PCSwingle: We already have tests for file renames, so I think this method should already be tested

mentat/code_file_manager.py
PCSwingle: We store the lines when we send the user's input to mentat; we do this because mentat's changes won't work correctly if the user changes the file (you can see in this same function we also check if the file was changed; if it was, we ask the user if they're sure they want their changes erased). But if you apply mentat's changes over your changes, I think it makes sense that you'd want undo to bring back your overwritten changes, rather than the stored lines. Let me know if you think we should do something else though, I'm open to suggestions!

mentat/diff_context.py
PCSwingle: Well for one, we can't actually send a message in the startup of Session yet (where this is), since Session is created before the prompt toolkit stuff is (because the prompttoolkit stuff uses the config contextvar); I added a TODO, which I plan on doing soon, to move the config to the client, which means we'll set up listening for the stream before we start the session, which means we'll actually be able to send a message and quit here (and other places). I actually prefer this to just outright quitting though, having no diff and no pr-diff is normal (it's what happens when you just call mentat without either argument) and it gives mentat the current git diff from the HEAD. If you think we should still quit though, I'm totally ok with that once we solve the config problem

mentat/include_files.py
PCSwingle: Well if we send it to sessionstream in this method not only do we have to make this method (and a few others) async, but we have no control over if we want to print the invalid files or not; additionally, I don't want to give our utility functions random side-effects (like printing), because it makes things a lot less clear and easily leads to bugs (where you use a function not realizing it was hand-built for this one purpose and has a lot of random side-effects). I agree it's messy having to return the invalid_paths, but I can't think of a better way; let me know if y'all think of anything though!

mentat/terminal/client.py
PCSwingle: This is a bit different (and won't be the same for every client); originally, the only reason we did this here is because zsh and some other shells automatically glob expand when you run commands, so by having this done at the very start we guaranteed no differences between different shells. In this PR I actually added glob expand to the get include files, so this function isn't actually necessary. Thinking about it more though, I don't think we need this function at all; if there are extremely small differences between shells because of it (which I don't even think there will be), I don't think it matters. I'll probably end up removing this function (as long as you agree)

mentat/commands.py
PCSwingle: Hmm, I do think that's clearer. I was planning on having something like /clear clear all context, but I agree it might be confusing. I'll change this to undo-all

mentat/code_context.py
```diff
@@ -58,36 +55,29 @@ class CodeContext:
...
+        cls, paths: list[Path], exclude_paths: list[Path], settings: CodeContextSettings
+    ):
         self = cls(settings)
 
-        await self._set_diff_context(replace_paths=True)
-        self._set_include_files()
+        self.diff_context = await DiffContext.create(
+            self.settings.diff, self.settings.pr_diff
+        )
+        self.include_files = get_include_files(paths, exclude_paths)
```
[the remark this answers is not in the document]
PCSwingle: Oh whoops I forgot to mention this! I kind of prefer that it doesn't put any files into context when I run just mentat, but I definitely can understand how this is useful; what does everyone else think?

mentat/code_file_manager.py
waydegg: Ah gotcha, thanks for the clarification! Yeah I think that behavior makes sense too

mentat/include_files.py
waydegg: >we send it to sessionstream in this method not only do we have to make this method (and a few others) async

I removed the method to synchronously enqueue stream messages (pushes into a queue without `await`, we don't have any queue size limits anyways), but maybe we should re-add this if it makes the development experience better. Thinking about it now, I don't see any downsides to re-add. 

>we have no control over if we want to print the invalid files or not

we could add a `print_invalid_files: bool = True` argument that toggles the prints

> random side-effects (like printing)

I feel like prints aren't a random side effect? Similar to log calls, I think its totally normal to have them wherever in your code base since they're capturing relevant context to whatever function/method they're being called from.

mentat/diff_context.py
waydegg: makes sense to me!

waydegg (approved):
Approving since all the remaining discussion is around preference for organizing code, functionally everything in this PR looks great. Left a reply regarding `get_include_files`, but will leave up to you for deciding how to go forward w/ it!

mentat/code_context.py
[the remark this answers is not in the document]
granawkins: Ya, works for me. We'll have better options for finding relevant code soon anyway.

granawkins:
For my part:
- The tuple thing is fine, I see the logic.
- The naming is also fine - realize `CreateAction` sounds like a verb so `CreationAction` is probably clearer.

So I'll merge now -

PR #144, opened 2023-10-12
Embeddings

granawkins:
A basic implementation of embeddings and context-generation benchmarks. This is a barebones implementation, off by default, to be built on in future PRs.

### Embeddings
- `llm_api.call_embedding_api`
- `embeddings.get_feature_similarity_scores(prompt, features)`: returns a list of scores for the given features (CodeFiles) based on their embedding similarity to a prompt. This can be used for context-generation or search.
- `EmbeddingsDatabase`, saved to .mentat/embeddings.json.gz, which is just {checksum: embedding} pairs.
- `CodeContext.settings.no_embeddings`, which is True (so no embeddings) by default.
- A new terminal arg `--embedding` to activate embeddings, and fill your remaining context with the most relevant files.

### Context Benchmarks
There are some notes at the top of `tests/benchmarks/context_benchmark.py`. Results with max_tokens = 7000:
```
True Positive:  2.000
True Negative:  96.000
False Positive: 10.000
False Negative: 6.000
Precision:      0.167   | How many selected features are relevant?
Recall:         0.250   | How many relevant features are selected?
F1:             0.200   | Weighted average of precision and recall```
```

### DIFF
--- a/mentat/code_context.py
+++ b/mentat/code_context.py
@@ -212,7 +212,9 @@ class CodeContext:
             self.features = features
         else:
             auto_tokens = _max_auto if _max_user is None else min(_max_auto, _max_user)
-            self.features = await self._get_auto_features(model, prompt, features, auto_tokens)
+            self.features = await self._get_auto_features(
+                model, prompt, features, auto_tokens
+            )
 
         for f in self.features:
             code_message += await f.get_code_message()
@@ -278,34 +280,36 @@ class CodeContext:
         # Sort by relative path
         def _feature_relative_path(f: CodeFile) -> str:
             return os.path.relpath(f.path, git_root)
+
         all_features = sorted(all_features, key=_feature_relative_path)
-        
 
         # If there's room, convert cmap to full-files, starting with the highest-scoring
-        cmap_features_tokens = await _count_tokens_in_features(
-            all_features, model
-        )
+        cmap_features_tokens = await _count_tokens_in_features(all_features, model)
         max_sim_tokens = max_tokens - cmap_features_tokens
         if max_sim_tokens > 0:
             sim_tokens = 0
 
             # Get embedding-similarity scores for all files
             sim_scores = await get_feature_similarity_scores(prompt, all_features)
-            file_scores = {(f.path, CodeMessageLevel.CODE): s 
-                           for f, s in zip(all_features, sim_scores)
-                           if f.path not in self.include_files}
-            
+            file_scores = {
+                (f.path, CodeMessageLevel.CODE): s
+                for f, s in zip(all_features, sim_scores)
+                if f.path not in self.include_files
+            }
+
             max_sim_tokens = max_tokens - include_features_tokens
             file_paths_sorted = sorted(
-                list(file_scores.keys()), 
-                key=lambda x: file_scores[x], 
-                reverse=True
+                list(file_scores.keys()), key=lambda x: file_scores[x], reverse=True
             )
             for path, _ in file_paths_sorted:
                 # Calculate the total change in length
-                i_cmap, cmap_feature = next((i, f) for i, f in enumerate(all_features) if f.path == path)
+                i_cmap, cmap_feature = next(
+                    (i, f) for i, f in enumerate(all_features) if f.path == path
+                )
                 recovered_tokens = await cmap_feature.count_tokens(model)
-                new_feature = CodeFile(path, level=CodeMessageLevel.CODE, diff=cmap_feature.diff)
+                new_feature = CodeFile(
+                    path, level=CodeMessageLevel.CODE, diff=cmap_feature.diff
+                )
                 new_tokens = await new_feature.count_tokens(model)
                 forecast = max_sim_tokens - sim_tokens + recovered_tokens - new_tokens
                 if forecast < 0:
@@ -313,7 +317,7 @@ class CodeContext:
                 # If it fits, add it
                 sim_tokens += recovered_tokens - new_tokens
                 all_features[i_cmap] = new_feature
-                
+
         return sorted(all_features, key=_feature_relative_path)
 
     async def include_file(self, code_file: CodeFile):
--- a/mentat/code_context.py
+++ b/mentat/code_context.py
@@ -13,6 +13,7 @@ from .code_file import CodeFile, CodeMessageLevel
 from .code_file_manager import CODE_FILE_MANAGER
 from .code_map import check_ctags_disabled
 from .diff_context import DiffContext
+from .embeddings import get_feature_similarity_scores
 from .errors import UserError
 from .git_handler import GIT_ROOT, get_non_gitignored_files, get_paths_with_git_diffs
 from .include_files import (
@@ -168,6 +169,7 @@ class CodeContext:
 
     async def get_code_message(
         self,
+        prompt: str,
         model: str,
         max_tokens: int,
     ) -> str:
@@ -176,12 +178,13 @@ class CodeContext:
             self._code_message is None
             or code_message_checksum != self._code_message_checksum
         ):
-            self._code_message = await self._get_code_message(model, max_tokens)
+            self._code_message = await self._get_code_message(prompt, model, max_tokens)
             self._code_message_checksum = self._get_code_message_checksum(max_tokens)
         return self._code_message
 
     async def _get_code_message(
         self,
+        prompt: str,
         model: str,
         max_tokens: int,
     ) -> str:
@@ -209,7 +212,7 @@ class CodeContext:
             self.features = features
         else:
             auto_tokens = _max_auto if _max_user is None else min(_max_auto, _max_user)
-            self.features = await self._get_auto_features(model, features, auto_tokens)
+            self.features = await self._get_auto_features(model, prompt, features, auto_tokens)
 
         for f in self.features:
             code_message += await f.get_code_message()
@@ -235,6 +238,7 @@ class CodeContext:
 
     async def _get_auto_features(
         self,
+        prompt: str,
         model: str,
         include_features: list[CodeFile],
         max_tokens: int,
@@ -273,6 +277,11 @@ class CodeContext:
 
         def _feature_relative_path(f: CodeFile) -> str:
             return os.path.relpath(f.path, git_root)
+        
+        # Get similarity scores
+        sim_scores = await get_feature_similarity_scores(prompt, all_features)
+        file_scores = {(f.path.name, f.level.key): s for f, s in zip(all_features, sim_scores)}
+        print(file_scores)
 
         return sorted(all_features, key=_feature_relative_path)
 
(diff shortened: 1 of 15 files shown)


mentat/conversation.py
```diff
@@ -151,6 +151,7 @@ async def get_model_response(self) -> list[FileEdit]:
         tokens = count_tokens(conversation_history, self.model)
         response_buffer = 1000
         code_message = await code_context.get_code_message(
+            messages[-1]["content"],
```
PCSwingle, line 154: This might not necessarily always be the user's input; it might have some extra text (check out code_edit_feedback.py), or it might just not be the user's input (it won't be soon with a PR I have coming out soon), not to mention the most recent feedback might have just been 'you made a mistake like this, fix it', which won't give us very good embeddings. I think it might be best to just pass all of the user's actual inputs; in general, the user's inputs aren't very long, so I think this could work pretty well. In an upcoming PR I automatically store all of the user's actual inputs (as opposed to a few things we give the role of user but are given by us) in a list, so we could use that then if you think that's a good idea

mentat/code_context.py
```diff
@@ -42,6 +43,7 @@ class CodeContextSettings:
     diff: Optional[str] = None
```
PCSwingle, line 43: (Referring to the _count_tokens_in_features function above this) I know you didn't add this in this PR, but why do we need a semaphore? I didn't think tiktoken made any http requests or used any outside resources at all

mentat/code_context.py
```diff
@@ -42,6 +43,7 @@ class CodeContextSettings:
...
     pr_diff: Optional[str] = None
     no_code_map: bool = False
+    no_embedding: bool = True
```
PCSwingle, line 46: I think this would make more sense (along with no_code_map) to be just embedding (or has_embeddings) and have it default to False

mentat/embeddings.py
```diff
@@ -0,0 +1,138 @@
...
+
+    tasks = [_count_tokens(f) for f in features]
+    return await asyncio.gather(*tasks)
+
+
+async def get_feature_similarity_scores(
+    prompt: str, features: list[CodeFile]
+) -> list[float]:
+    """Return the similarity scores for a given prompt and list of features."""
+    assert all([isinstance(f, CodeFile) for f in features]), "Invalid feature list"
```
PCSwingle, line 91: Is this necessary? We have pyright which should generally work

mentat/embeddings.py
```diff
@@ -0,0 +1,138 @@
...
+
+def _cosine_similarity(v1: list[float], v2: list[float]) -> float:
+    """Calculate the cosine similarity between two vectors."""
+    dot_product = np.dot(v1, v2)
+    norm_v1 = np.linalg.norm(v1)
+    norm_v2 = np.linalg.norm(v2)
+    return dot_product / (norm_v1 * norm_v2)
+
+
+async def _count_feature_tokens(features: list[CodeFile], model: str) -> list[int]:
```
PCSwingle, line 75: Is this not a duplicate of the other count feature tokens function? And just like my other comment, do we need a semaphore? If not, both of these functions wouldn't be needed

mentat/embeddings.py
```diff
@@ -0,0 +1,138 @@
...
+) -> list[float]:
+    """Return the similarity scores for a given prompt and list of features."""
+    assert all([isinstance(f, CodeFile) for f in features]), "Invalid feature list"
+    global database
+    stream = SESSION_STREAM.get()
+
+    # Keep things in the same order
+    checksums: list[str] = [f.get_checksum() for f in features]
+    tokens: list[int] = await _count_feature_tokens(features, EMBEDDING_MODEL)
+    skip: list[bool] = [t > EMBEDDING_MAX_TOKENS for t in tokens]
```
PCSwingle, line 98: This skip list seems unnecessary; why not just check for the tokens on line 108 instead of doing it here and zipping it up?

mentat/code_context.py
jakethekoenig: I think it's about mentat opening too many files at once. I believe @biobootloader had an issue with that before the semaphore was added

mentat/embeddings.py
PCSwingle: Same with the tokens list (as long as we can get rid of the semaphore)

mentat/embeddings.py
```diff
@@ -0,0 +1,138 @@
...
+async def get_feature_similarity_scores(
+    prompt: str, features: list[CodeFile]
+) -> list[float]:
+    """Return the similarity scores for a given prompt and list of features."""
+    assert all([isinstance(f, CodeFile) for f in features]), "Invalid feature list"
+    global database
+    stream = SESSION_STREAM.get()
+
+    # Keep things in the same order
+    checksums: list[str] = [f.get_checksum() for f in features]
```
PCSwingle, line 96: Do we need checksums at all? Couldn't we just store the full content as the key? Storing vector embeddings is already pretty expensive (space-wise) so I don't think we need to worry about space

tests/benchmarks/context_benchmark.py
PCSwingle, line 1: I like this benchmark a lot!

tests/benchmarks/utils.py
```diff
@@ -0,0 +1,18 @@
+import os
+
+from git import Repo
+
+CLONE_TO_DIR = f"{os.path.dirname(__file__)}/../../../"
```
PCSwingle, line 5: Shouldn't this just be the home file (Path.home())?

mentat/code_context.py
PCSwingle: Ah, I see, that's unfortunate

tests/benchmarks/context_benchmark.py
```diff
@@ -0,0 +1,122 @@
...
+from .utils import clone_repo
+
+pytestmark = pytest.mark.benchmark
+
+tests = [
+    {
+        "name": "Mentat: replace subprocess/git with GitPython",
+        "codebase_url": "http://github.com/AbanteAI/mentat",
+        "codebase_name": "mentat",
+        "commit": "",  # hexsha
```
biobootloader, line 47: Is the plan to add the commit this PR ends up being when the PR is merged?

I guess usually if adding a new benchmark you could use the latest commit on Mentat, as long as when adding the benchmark you don't change anything else.

tests/benchmarks/utils.py
biobootloader: This is cloning outside the Mentat repo, right? What if the repo is Mentat, like in your new benchmark? won't there be a conflict?

Maybe we should make a directory inside mentat, something like `repos_for_benchmarks`?

mentat/embeddings.py
granawkins: Yes - just combined these into a single function `count_feature_tokens` in CodeFile.py (for import reasons).

mentat/embeddings.py
granawkins: Right again - I needed this earlier but now its useless.

mentat/conversation.py
granawkins: I see, ya that makes sense. Shall we leave it as-is for this PR and update it when that other PR is merged?

tests/code_context_test.py
```diff
@@ -204,22 +204,22 @@ async def test_get_code_message_cache(
         "mentat.code_context.CodeContext._get_code_message"
     )
     mock_get_code_message.return_value = "test1"
-    value1 = await code_context.get_code_message("gpt-4", 1e6)
+    value1 = await code_context.get_code_message("", "gpt-4", 1e6)
```
biobootloader, line 207: no need to change, but it'd be more clear if we called with keyword args, i.e. `get_code_message(prompt="", model="gpt-4", ...)

mentat/llm_api.py
```diff
@@ -86,6 +86,17 @@ async def call_llm_api(
     return _add_newline(response)
 
 
+def call_embedding_api(
+    input: list[str], model: str = "text-embedding-ada-002"
+) -> list[list[float]]:
+    if is_test_environment():
+        logging.critical("OpenAI call attempted in non benchmark test environment!")
+        raise MentatError("OpenAI call attempted in non benchmark test environment!")
```
biobootloader, line 94: instead of duplicating these lines from `call_llm_api` lets make a new function that both call, `raise_if_in_test_environment`

mentat/conversation.py
biobootloader: Hmm it does get complex when the user has sent a second message / request. The embedding of just the second message might totally miss context that's in the first message. So the system would appear to get a lot dumber on the second prompt

biobootloader: let's not worry about it for now, as our goal here is getting embeddings added. Perhaps the right way to do it will be have gpt-3 create things to embed and search for, relevant to the last prompt, given the conversation history as well

mentat/embeddings.py
PCSwingle: Ok, that all makes sense. Thanks!

biobootloader: when we have it cached and maintained somewhere how about we have access to that just be through the full content, so the checksums would be just an internal optimization detail to the cache, and we wouldn't need to calculate any checksums here

PCSwingle (approved):
Looks good to me!

tests/benchmarks/context_benchmark.py
granawkins: Oh, if we follow your comment from above (`repos_for_benchmarks`) this wouldn't be an issue, that'd be great.

mentat/code_file.py
```diff
@@ -139,19 +142,35 @@ async def _get_code_message(self) -> list[str]:
...
         return self._code_message
 
     async def count_tokens(self, model: str) -> int:
         code_message = await self.get_code_message()
         return count_tokens("\n".join(code_message), model)
+
+
+async def count_feature_tokens(features: list[CodeFile], model: str) -> list[int]:
+    """Return the number of tokens in each feature."""
+    sem = asyncio.Semaphore(10)
```
biobootloader, line 169: wouldn't be part of this PR, but does it make sense to move the Semaphore to be part of CodeFileManager? since that's where the limit is?

mentat/embeddings.py
```diff
@@ -0,0 +1,121 @@
...
+        # Place each item in the first batch that it fits in
+        placed = False
+        for batch in batches:
+            if sum(data[k] for k in batch) + value <= batch_size:
+                batch.append(key)
+                placed = True
+                break
+        if not placed:
+            batches.append([key])
+    return batches
```
biobootloader, line 63: Nice algorithm! 🚀

biobootloader (approved):
Awesome! Go ahead and merge after addressing my comments / making issues for things to do later

tests/benchmarks/utils.py
granawkins: `mentat/tests/benchmarks/repos`?

mentat/code_file.py
granawkins: Yes, makes sense. And have it be configurable.

PR #145, opened 2023-10-12
Manifest added

jakethekoenig:
The `requirements.txt` was not being found by brew because it was not included, not because it was specified with relative path. Data files are now specified via Manifest and the `requirements.txt` is added. You can test that the `requirements.txt` and other files are included with the following command:
```
python setup.py sdist bdist_wheel
```


biobootloader (approved):
Thanks!

PR #150, opened 2023-10-13
Reversible parsing

jakethekoenig:
* An inverse method for block parser
* A script that translates transcriptions from on parser format to another. Run with:
```
./scripts/translate_transcript.py --transcript your_transcript.log --starting-format replacement --ending-format block
```
Only the block parser has an inverse for now. My hope is this will make it easy to change the training data between edit formats. I also hope this will allow us to make training examples from git diffs. We'll need to make a function that parses git diffs into file_edits. We don't necessarily have to make this compatible with the current parsing infrastructure because git diffs have all kinds of meta data you wouldn't want to expect an llm to output so just `string` to `FileEdits[]` is fine. We'll need to add inverses for all the formats we want to generate training examples for.

I think it'd be better to push the following in separate PRs:
* An inverse for the replacement format
* Git diff -> FileEdits[] function

Possibly the script would also be better in a separate PR but I think it makes the change useful in a way that it's not without it.


mentat/parsers/block_parser.py
```diff
@@ -197,3 +197,71 @@ def _add_code_block(
...
+                ans += json.dumps(tmp, indent=4) + "\n"
+                if (
+                    tmp[_BlockParserJsonKeys.Action.value]
+                    == _BlockParserAction.CreateFile.value
+                ):
+                    ans += _BlockParserIndicator.Code.value + "\n"
+                    ans += "\n".join(file_edit.replacements[0].new_lines) + "\n"
+                ans += _BlockParserIndicator.End.value + "\n"
+            else:
+                for replacement in file_edit.replacements:
```
PCSwingle, line 234: There can be replacements in a fileedit that is a renamefile action if the model makes changes and does a rename on the same file (and if this is meant to potentially convert from fileedits from other formats, the createfile action can have multiple replacements).

PCSwingle: In this case, I think it would be more trouble than it's worth to add multiple replacements to createfile (especially since it really shouldn't need to put more than one insert in a new file) but the renamefile I can see happening a decent amount of the time

tests/parser_tests/block_format_test.py
```diff
@@ -447,6 +450,82 @@ async def test_multiple_blocks(
...
+    mock_collect_user_input,
+    mock_setup_api_key,
+    mock_code_file_manager,
+    mock_git_root,
+):
+    # This test verifies that the inverse is a left inverse to stream_and_parse_llm_response. That is we can go:
+    #                               llm_message -> file_edits -> llm_message
+    # and get back where we started. Actually what we care about is that it's a right inverse. That if we go:
+    #                               file_edits -> llm_message -> file_edits
+    # we get back where we started. So this test verifies things we don't necessarily care about like the order of the
```
PCSwingle, line 466: Is there a reason we don't test file_edits -> llm_message -> file_edits?

mentat/parsers/block_parser.py
```diff
@@ -197,3 +197,71 @@ def _add_code_block(
             )
         )
         return ""
+
+    def file_edits_to_llm_message(self, generated: tuple[str, list[FileEdit]]) -> str:
+        """
+        Inverse of stream_and_parse_llm_response;
+        takes in the generated message and file edits and returns the original message
+        """
+        ans = generated[0].split("@")[0]  # BlockParserIndicator.Start.value)[0]
```
PCSwingle, line 206: Hold on, I'm a bit confused on what's going on here; is generated[0] just the actual message that the model output (that we're trying to get back, but potentially in a different format)?  If so, the @ won't always be the end of the 'conversation' that gets generated (since some formats have different markers). We could do one of 2 things: change parsers to output the actual 'conversation' that the model output (this would also catch conversation in between edits, which you drop right now), or we could check for any format's start marker.

tests/parser_tests/block_format_test.py
```diff
@@ -447,6 +450,84 @@ async def test_multiple_blocks(
...
+    # edits and white space.
+    cwd = Path(os.getcwd())
```
PCSwingle, line 468: This will get an absolute path; normally the model outputs relative to the git_root (set to temp_testbed by default); not a big deal but worth pointing out

PCSwingle:
Looks good to me except for a couple comments I left! Let me know if you need me to look at the Windows bug; one important thing to mention is that the model always outputs and recieves posix paths no matter what os it's on (for consistency)

tests/parser_tests/block_format_test.py
jakethekoenig: Mostly laziness. It's easier to compare strings than `FileEdit[]s`. But I think I'll change it.

mentat/parsers/block_parser.py
PCSwingle: I like that idea a lot! Definitely is a bit messy right now haha

jakethekoenig:
> Looks good to me except for a couple comments I left! Let me know if you need me to look at the Windows bug; one important thing to mention is that the model always outputs and recieves posix paths no matter what os it's on (for consistency)

Everything you pointed out should be fixed now except for the right vs. left inverse thing in the test which I think I'll leave as is if it's okay with you.

tests/parser_tests/block_format_test.py
```diff
@@ -465,63 +464,80 @@ async def test_inverse(
...
+        {
+            "file": "file1.txt",
+            "action": "replace",
+            "start-line": 4,
+            "end-line": 4
+        }
+        @@code
+        # better measure in a replacement
         @@end
                           """)
```
PCSwingle, line 535: What's going on here?

mentat/parsers/block_parser.py
```diff
@@ -232,11 +239,11 @@ def file_edits_to_llm_message(self, generated: tuple[str, list[FileEdit]]) -> st
                     ans += _BlockParserIndicator.Code.value + "\n"
                     ans += "\n".join(file_edit.replacements[0].new_lines) + "\n"
                 ans += _BlockParserIndicator.End.value + "\n"
-            else:
+            if not file_edit.is_creation:
                 for replacement in file_edit.replacements:
                     tmp = {}
-                    tmp[_BlockParserJsonKeys.File.value] = str(
-                        file_edit.file_path.as_posix()
+                    tmp[_BlockParserJsonKeys.File.value] = (
```
PCSwingle, line 245: This is a bit of a special case, but when the model does a rename and then does a change to the file it renamed, it normally outputs the file to change as the NEW file (because it thinks it already renamed the previous file); in the parser, you can see we handle this with a 'rename_map', which maps the 'new' file_names to old ones, so that the edits applied to the new name apply to the correct file. It isn't the end of the world if you don't handle this edge case (along with a few of the edge cases we already just don't handle, like the model renaming a file twice in a row haha), since the parser will still modify the correct file (unless the model renamed file_1 -> file_2 and then file_0 -> file_1, another extreme edge case), so I don't care if you leave it as is, but I wanted to make you aware

tests/parser_tests/block_format_test.py
```diff
@@ -465,63 +464,80 @@ async def test_inverse(
...
+            "file": "file1.txt",
             "action": "rename-file",
-            "name": "{(cwd / "file2.txt").as_posix() }"
-        }}
+            "name": "file2.txt"
+        }
+        @@end
+        @@start
+        {
+            "file": "file1.txt",
```
PCSwingle, line 517: As an example of the rename edge case I mentioned, normally the model would output file2.txt here and we would convert it to file1.txt. If you do end up fixing that edge case (which I'm perfectly fine with you not fixing it), you would change these 2 changes to have file file2.txt.

PCSwingle (approved):
All looks good to me, but take a look at the last couple of comments I left

tests/parser_tests/block_format_test.py
jakethekoenig: Sorry, can you say more? It's an insertion/replacement after a rename. I probably should have made it say "I inserted this comment in a rename". I didn't want it to be the exact same code as before as a sanity check.

mentat/parsers/block_parser.py
jakethekoenig: Thanks!

jakethekoenig: So to clarify: the parser works if the model outputs the original file name or the new file name but the original file name is more correct? Seems like inverse works as it should then right? Even though this is another way it may fail to be an inverse to `_parse_and_stream_llm_response`?

PCSwingle: I wouldn't say the original file name is 'more' correct, but yeah, this will work (although it does introduce another way there can be multiple inverses)

PR #152, opened 2023-10-13
Conversation viewer

PCSwingle:
TODO:

Increase aesthetic appeal
Should we make conversation command hidden? It probably isn't too useful to the average viewer, and it looks very hacked together

### DIFF
--- a/mentat/conversation.py
+++ b/mentat/conversation.py
@@ -3,9 +3,16 @@ from __future__ import annotations
 import json
 import logging
 from contextvars import ContextVar
+from enum import Enum
+from pathlib import Path
 from timeit import default_timer
 
 from openai.error import InvalidRequestError, RateLimitError
+
+from mentat.parsers.file_edit import FileEdit
+from mentat.parsers.parser import PARSER, Parser
+from mentat.utils import fetch_resource, mentat_dir_path
+from tests.conftest import SessionStream
 
 from .code_context import CODE_CONTEXT
 from .config_manager import CONFIG_MANAGER, user_config_path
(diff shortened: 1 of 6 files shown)


PCSwingle:
Current state screenshot:

image

PCSwingle:
Just uploaded a transcript with a fairly large conversation; feel free to use the transcript viewer to check it out!

[transcript_20231013_144235.log](https://github.com/AbanteAI/mentat/files/12899698/transcript_20231013_144235.log)

mentat/conversation.py
```diff
@@ -35,8 +44,12 @@ def __init__(self):
         self.model = config.model()
         self.messages = list[dict[str, str]]()
 
+        # This contain the messages the user actually sends and the messages the model output
+        # along with a snapshot of exactly what the modelg got before that message
```
jakethekoenig, line 48: s/modelg/model/g

mentat/conversation.py
```diff
@@ -18,13 +23,17 @@
...
 )
-from .parsers.file_edit import FileEdit
-from .parsers.parser import PARSER, Parser
-from .session_stream import SESSION_STREAM, SessionStream
+from .session_stream import SESSION_STREAM
 
 CONVERSATION: ContextVar[Conversation] = ContextVar("mentat:conversation")
 
 
+class MessageRole(Enum):
```
jakethekoenig, line 31: I think it's better to call this `Role`. Thanks for refactoring it in this way. I think it's a bit cleaner.

jakethekoenig (approved):
When you say hide `/conversation` you mean remove it from the autocompleter? I think that'd be a good idea. I also like the idea of using jinja for our templates. I think it's good enough to merge though. One feature requests going forward: It'd be really nice if `transcript_viewer` could automatically open the most recent transcript and if you could scroll through them with the arrow keys.

PCSwingle:
> When you say hide `/conversation` you mean remove it from the autocompleter? I think that'd be a good idea. I also like the idea of using jinja for our templates. I think it's good enough to merge though. One feature requests going forward: It'd be really nice if `transcript_viewer` could automatically open the most recent transcript and if you could scroll through them with the arrow keys.

Oohh, I really like that idea of being able to scroll through them. I'll add an issue for both jinja and the transcript scrolling

PR #166, opened 2023-10-17
Empty string handling

jakethekoenig:
Currently if you start mentat with no files or give it an empty prompt mentat crashes when it tries to embed an empty string. After fixing that I also made the terminal client not call gpt on empty prompts to prevent expensive errors.

### DIFF
--- a/mentat/terminal/client.py
+++ b/mentat/terminal/client.py
@@ -64,15 +64,17 @@ class TerminalClient:
         assert isinstance(self.session, Session), "TerminalClient is not running"
         while True:
             input_request_message = await self.session.stream.recv("input_request")
-            # TODO: Make extra kwargs like plain constants
-            if (
-                input_request_message.extra is not None
-                and input_request_message.extra.get("plain")
-            ):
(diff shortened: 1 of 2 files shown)


mentat/terminal/client.py
```diff
@@ -64,15 +64,17 @@ async def _handle_input_requests(self):
...
-            if (
-                input_request_message.extra is not None
-                and input_request_message.extra.get("plain")
-            ):
-                prompt_session = self._plain_session
-            else:
-                prompt_session = self._prompt_session
-            user_input = await prompt_session.prompt_async(handle_sigint=False)
+            user_input = None
+            while user_input is None or user_input == "":
```
PCSwingle, line 68: What about when we have a "Y/n" question and the user just hits enter (empty should default to yes)? I also think this would be best to do on the server anyways; we can just put this loop in session.py in the main loop where we wait for user input

jakethekoenig: Yeah that is a lot better

PCSwingle (approved):

PR #167, opened 2023-10-17
add way to exit main session loop, add command catching and quitting to sessioninput

### LINKED ISSUE #148: Running commands when in Y/n/i after response
When in Y/n/i mode, commands won't work, but still show up in the autocomplete. We either need to disable the command autocomplete or make commands work (I think the second one would be better).

PCSwingle:
Fixes #148. What does everyone think of the SessionExit as a way to quit the session? It will only work from within the _main coroutine of the session, and I'm not entirely sure I like using an exception to quit, but it is fairly easy and simple.

### DIFF
--- a/mentat/session_input.py
+++ b/mentat/session_input.py
@@ -3,38 +3,42 @@ import logging
 import shlex
 from typing import Any, Coroutine
 
-from mentat.commands import Command
-
 from .errors import RemoteKeyboardInterrupt, SessionExit
 from .session_stream import SESSION_STREAM, StreamMessage
 
 
-async def collect_user_input(**kwargs: Any) -> StreamMessage:
+async def _get_input_request(**kwargs: Any) -> StreamMessage:
+    stream = SESSION_STREAM.get()
+
+    message = await stream.send("", channel="input_request", **kwargs)
+    response = await stream.recv(f"input_request:{message.id}")
+    return response
+
+
+async def collect_user_input(plain: bool = False) -> StreamMessage:
     """Listens for user input on a new channel
 
     send a message requesting user to send a response
     create a new broadcast channel that listens for the input
     close the channel after receiving the input
     """
-    stream = SESSION_STREAM.get()
+    # This fixes a circular import
+    from mentat.commands import Command
 
-    message = await stream.send("", channel="input_request", **kwargs)
-    response = await stream.recv(f"input_request:{message.id}")
-
+    response = await _get_input_request(plain=plain)
     logging.debug(f"User Input: {response.data}")
 
     # Intercept and run commands
-    if not kwargs.get("plain"):
+    if not plain:
         while isinstance(response.data, str) and response.data.startswith("/"):
-            arguments = shlex.split(message.data[1:])
+            arguments = shlex.split(response.data[1:])
             command = Command.create_command(arguments[0])
             await command.apply(*arguments[1:])
 
-            message = await stream.send("", channel="input_request", **kwargs)
-            response = await stream.recv(f"input_request:{message.id}")
+            response = await _get_input_request(plain=plain)
 
     # Quit on q
-    if response.data == "q":
+    if isinstance(response.data, str) and response.data.strip() == "q":
         raise SessionExit
 
     return response
--- a/mentat/session_input.py
+++ b/mentat/session_input.py
@@ -1,8 +1,11 @@
 import asyncio
 import logging
+import shlex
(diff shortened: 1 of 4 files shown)


waydegg:
Getting circular import errors:

```
~/Development/mentat-main yni-commands !1 ❯ mentat pyproject.toml
  File "/Users/waydegg/Development/mentat-main/venv/bin/mentat", line 5, in <module>
    from mentat.terminal.client import run_cli
  File "/Users/waydegg/Development/mentat-main/mentat/terminal/client.py", line 14, in <module>
    from mentat.session import Session
  File "/Users/waydegg/Development/mentat-main/mentat/session.py", line 10, in <module>
    from .code_context import CODE_CONTEXT, CodeContext, CodeContextSettings
  File "/Users/waydegg/Development/mentat-main/mentat/code_context.py", line 11, in <module>
    from .code_file import CodeFile, CodeMessageLevel, count_feature_tokens
  File "/Users/waydegg/Development/mentat-main/mentat/code_file.py", line 9, in <module>
    from .code_file_manager import CODE_FILE_MANAGER
  File "/Users/waydegg/Development/mentat-main/mentat/code_file_manager.py", line 19, in <module>
    from .session_input import ask_yes_no
  File "/Users/waydegg/Development/mentat-main/mentat/session_input.py", line 6, in <module>
    from mentat.commands import Command
  File "/Users/waydegg/Development/mentat-main/mentat/commands.py", line 7, in <module>
    from mentat.code_file_manager import CODE_FILE_MANAGER
ImportError: cannot import name 'CODE_FILE_MANAGER' from partially initialized module 'mentat.code_file_manager' (most likely due to a circular import) (/Users/waydegg/Development/mentat-main/mentat/code_file_manager.py)
```

mentat/session_input.py
```diff
@@ -20,6 +23,20 @@ async def collect_user_input(**kwargs: Any) -> StreamMessage:
 
     logging.debug(f"User Input: {response.data}")
 
+    # Intercept and run commands
+    if not kwargs.get("plain"):
```
waydegg: Instead of supporting and `**kwargs` for ` collect_user_input`, I think we should add a specific flag argument (e.g. `disable_commands: bool`) for disabling command interception since it's a lot clearer what's going on.

PCSwingle:
Just fixed... this does make me think we might want to switch to the SESSION_CONTEXT variable after all though (since that would fix pretty much all circular imports ever)

mentat/session_input.py
PCSwingle: Just changed how things work a tiny bit; I still have plain disabling commands, but it isn't in kwargs anymore. I think we should keep plain and disable_commands as one thing though, since the entire point of the plain is that it doesn't have autocomplete (like commands)

waydegg:
We should update this too right?

https://github.com/AbanteAI/mentat/blob/a6d7343aa82033037aaa3ae74227223698f11d3a/mentat/code_edit_feedback.py#L22

waydegg:
Do we always want users to be able to quit a session with "q"? For example, when we prompt "y/n/i" do we want "q" to still quit? Before this PR this wouldn't be possible.

waydegg:
I think `SessionExit` is fine for now since we aren't interrupting anything important business logic by raising an Exception.

mentat/session_input.py
```diff
@@ -1,25 +1,46 @@
...
     send a message requesting user to send a response
     create a new broadcast channel that listens for the input
     close the channel after receiving the input
     """
-    stream = SESSION_STREAM.get()
-
-    message = await stream.send("", channel="input_request", **kwargs)
-    response = await stream.recv(f"input_request:{message.id}")
+    # This fixes a circular import
+    from mentat.commands import Command
```
waydegg, line 26: This also makes me want to have a single `SESSION_CONTEXT` lol

mentat/session_input.py
waydegg: Sg, I didn't understand what "plain" meant until now ("a 'plain' prompt")

waydegg (approved):

mentat/session_input.py
PCSwingle: 🙃

PCSwingle:
> Do we always want users to be able to quit a session with "q"? For example, when we prompt "y/n/i" do we want "q" to still quit? Before this PR this wouldn't be possible.

I think so, yeah

> We should update this too right?

What do you mean? I think that part is fine, right?

Issue #160, opened 2023-10-16
Try using Markdown to format prompts


biobootloader:
https://x.com/vagabondjack/status/1713602004969205804?s=20

use-the-fork:
Take a look at my pr #432  covers this.

PR #168, opened 2023-10-17
Unified diff tweaking

PCSwingle:
Fix unified diff bugs; tweak prompt a bit to be slightly more similar to the git diff format. Things I still haven't tried yet:

Adding a/ and b/ to filenames (like a git diff would have)
Having diff --git file_name_1 file_name_2 line at start

Working on seeing if adding those has any significant effect (although I doubt it will)

### DIFF
--- a/mentat/parsers/diff_utils.py
+++ b/mentat/parsers/diff_utils.py
@@ -14,7 +14,7 @@ def matching_index(orig_lines: list[str], new_lines: list[str]) -> int:
                 new_orig_lines = [s for s in orig_lines if s]
                 new_new_lines = [s for s in new_lines if s]
                 index = _exact_match(new_orig_lines, new_new_lines)
-                if index != -1:
+                if new_orig_lines and index != -1:
                     index = orig_lines.index(new_orig_lines[index])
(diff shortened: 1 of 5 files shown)


biobootloader:
How are benchmark comparisons looking? If you've been running them to test tweaks it'd be good to share anything that's made a difference or not, to help improve our intuitions as well.

Also related is using markdown might help it understand prompts better: https://github.com/AbanteAI/mentat/issues/160

jakethekoenig (approved):

PCSwingle:
> How are benchmark comparisons looking? If you've been running them to test tweaks it'd be good to share anything that's made a difference or not, to help improve our intuitions as well.
> 
> Also related is using markdown might help it understand prompts better: #160

It was pretty much the exact same as the other formats unfortunately

PCSwingle:
Just finished the results for adding a/ b/ and diff --git and it performed worse (although strangely with much fewer tokens than last time? I think with this format especially tokens used is going to vary a lot). I'll push these changes to a branch so we don't lose this format, but I don't think it works well.

PR #171, opened 2023-10-17
Backoff llm api calls

PCSwingle:
Add backoff to llm api calls and stop errors from bricking benchmarks. @jakethekoenig I don't really like the way I had to set up the python client to detect when the session is stopped, but because of the way we run the benchmarks (where we wait for the next input request to confirm that all changes have been made) I'm not sure of a better way to do it. Maybe in the future we should have the server send a message confirming that changes have been written? This could be useful for the vscode extension too.

### DIFF
--- a/mentat/llm_api.py
+++ b/mentat/llm_api.py
@@ -5,12 +5,14 @@ import os
 import sys
 from contextvars import ContextVar
 from dataclasses import dataclass
+from functools import partial
 from typing import Any, AsyncGenerator, Optional, cast
 
 import backoff
 import openai
 import openai.error
 import tiktoken
+from backoff.types import Details
 from dotenv import load_dotenv
 from openai.error import AuthenticationError, RateLimitError, Timeout
 
@@ -73,14 +75,23 @@ def raise_if_in_test_environment():
         raise MentatError("OpenAI call attempted in non benchmark test environment!")
 
 
+async def warn_user(message: str, max_tries: int, details: Details):
+    stream = SESSION_STREAM.get()
+
+    warning = f"{message}: Retry number {details['tries']}/{max_tries - 1}..."
+    await stream.send(warning, color="light_yellow")
+
+
 @backoff.on_exception(
     wait_gen=backoff.expo,
     exception=Timeout,
     max_tries=5,
     base=2,
     factor=2,
+    jitter=None,
     logger="",
     giveup_log_level=logging.INFO,
+    on_backoff=partial(warn_user, "Error reaching OpenAI's servers", 5),
 )
 @backoff.on_exception(
     wait_gen=backoff.expo,
@@ -88,8 +99,10 @@ def raise_if_in_test_environment():
     max_tries=3,
     base=2,
     factor=10,
+    jitter=None,
     logger="",
     giveup_log_level=logging.INFO,
+    on_backoff=partial(warn_user, "Rate limit recieved from OpenAI's servers", 3),
 )
 async def call_llm_api(
     messages: list[dict[str, str]], model: str
--- a/mentat/llm_api.py
+++ b/mentat/llm_api.py
@@ -7,11 +7,12 @@ from contextvars import ContextVar
 from dataclasses import dataclass
 from typing import Any, AsyncGenerator, Optional, cast
 
+import backoff
 import openai
 import openai.error
 import tiktoken
 from dotenv import load_dotenv
-from openai.error import AuthenticationError
+from openai.error import AuthenticationError, RateLimitError, Timeout
 
 from mentat.session_stream import SESSION_STREAM
 
@@ -72,6 +73,12 @@ def raise_if_in_test_environment():
         raise MentatError("OpenAI call attempted in non benchmark test environment!")
 
 
+@backoff.on_exception(
+    wait_gen=backoff.expo, exception=Timeout, max_tries=5, base=2, factor=2
+)
+@backoff.on_exception(
+    wait_gen=backoff.expo, exception=RateLimitError, max_tries=3, base=2, factor=10
+)
 async def call_llm_api(
     messages: list[dict[str, str]], model: str
 ) -> AsyncGenerator[Any, None]:
--- a/mentat/llm_api.py
+++ b/mentat/llm_api.py
@@ -74,10 +74,22 @@ def raise_if_in_test_environment():
 
 
 @backoff.on_exception(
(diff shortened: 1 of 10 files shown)


tests/benchmarks/exercism_practice.py
```diff
@@ -138,7 +139,7 @@ async def run_exercise(problem_dir, language="python", max_iterations=2):
     )
 
     iterations = 0
-    while iterations < max_iterations:
+    while iterations < max_iterations and client.started:
```
PCSwingle, line 142: If it exits because of the session turning off (presumably because of an error), should we mark the test reason for failure as 'error'? I think it makes sense, but was wondering what y'all might think

mentat/llm_api.py
```diff
@@ -72,6 +73,12 @@ def raise_if_in_test_environment():
         raise MentatError("OpenAI call attempted in non benchmark test environment!")
 
 
+@backoff.on_exception(
+    wait_gen=backoff.expo, exception=Timeout, max_tries=5, base=2, factor=2
+)
+@backoff.on_exception(
+    wait_gen=backoff.expo, exception=RateLimitError, max_tries=3, base=2, factor=10
+)
```
biobootloader, line 81: iirc these have some options for logging? It'd be nice to know how often these are being encountered / realize it's slowing things down

tests/benchmarks/exercism_practice.py
biobootloader: We should definitely record failures due to rate limiting or api issues separately from other failures

mentat/llm_api.py
jakethekoenig: It might also be good to write something to the terminal client if these are encountered as well. If the user keeps hitting `RateLimitErrors` then mentat will hang for 3 minutes before giving up right? That seems not ideal. Maybe a fringe case.

mentat/python_client/client.py
```diff
@@ -29,25 +29,36 @@ def __init__(
...
             source=StreamMessageSource.CLIENT,
             channel=f"input_request:{self._input_request_message.id}",
         )
-        self._input_request_message = await self.session.stream.recv("input_request")
+        # If the Session shuts down, we'll be waiting here forever;
+        # but because we need to wait for this request to insure that all changes have been made,
+        # setting a timeout is the best way to determine if the Session has shut down
+        try:
+            self._input_request_message = await asyncio.wait_for(
+                self.session.stream.recv("input_request"), 3
```
jakethekoenig, line 49: Is 3 enough for you? I get `TImeoutError` 100% of the time but it works if I change it to 300. In theory we should only wait here when the session encounters an exception and shuts down right? It seems fine to bias towards a high number.

PCSwingle: Actually, I think I might just add in the stream message that indicates that the changes are finished now

mentat/llm_api.py
PCSwingle: It should only hang for 10 + 20 + 40 seconds, so a bit more than one minute; I think it's good as is, since if it fails all 3 times it does print the error it encountered, but if you still want to I can add that.

biobootloader:
So if a user is running Mentat what will this look like to them? ~1 minute freeze and then it will show the errors?

If it's easy to print something as soon as each error his hit, like "Encountered RateLimitError, retrying...", that'd be ideal I think

mentat/python_client/client.py
jakethekoenig: I'm a little confused. Shouldn't the input_request signal the edits have been made? And isn't a number greater than the max LLM response time be necessary here? Which should be at least 30? Session exits should be rare so I'd bias towards not killing sessions that would have succeeded

mentat/llm_api.py
jakethekoenig: Oh I misunderstood. I thought 2, 20, 200. That seems fine-ish. There's sort of a tension here where the optimal behavior in user and benchmark seems different:
* For benchmark retrying a dozen times would make sense. I don't want my overnight benchmark to fail because OpenAI had a short outage
* for users retry once in ~2 seconds then report to user seems best to me

I think the current change is good enough and better than the status quo but could be better

biobootloader:
> There's sort of a tension here where the optimal behavior in user and benchmark seems different:
> For benchmark retrying a dozen times would make sense. I don't want my overnight benchmark to fail because OpenAI had a short outage for users retry once in ~2 seconds then report to user seems best to me

Yeah I agree with this, could we set the numbers for users and then have benchmarks use different ones?

mentat/llm_api.py
granawkins: I believe OpenAI's rate limit is reset every minute, and every day. So couldn't we just retry once, after 60 seconds, on `RateLimitError`?

mentat/python_client/client.py
PCSwingle: I completely changed this, now I have the stream send a message in a different channel signaling that the edits are completed and we listen to that instead of waiting for an input request message (which always felt a bit hacky anyways).

mentat/llm_api.py
PCSwingle: Hmm, I think you've convinced me, I'll add sending a message.

PCSwingle:
Just added the backoff error messages. I also added synchronous stream sending but eventually found out that backoff could take async function as well so I didn't end up using them, but now we have them.

jakethekoenig (approved):

mentat/llm_api.py
```diff
@@ -82,11 +82,15 @@ async def warn_user(message: str, max_tries: int, details: Details):
     await stream.send(warning, color="light_yellow")
 
 
+# This can be mocked in benchmarks to increase the wait time
+backoff_delay_base = 2
```
biobootloader, line 86: does this work? I don't know when python decorator args get applied

biobootloader (approved):
🚀

mentat/llm_api.py
biobootloader: I think they'd get applied at import time, as the decorator would run then to create the decorated functions

PCSwingle: Good point, I'm not entirely sure. I'll test it and if it doesn't work I'll remove this

PCSwingle: It didn't work; we could probably increase the delay only in benchmarks by checking if PYTEST is active, but I'm honestly not sure we need to (if the openai api is down for 40 seconds, it's probably down for a while); I'm just going to merge and we can figure this out later if we do need to.

PR #183, opened 2023-10-18
switch to use abs-paths for parsers renames

### LINKED ISSUE #181: Parser rename file name isn't an absolute path

PCSwingle:
Fixes #181. I meant for this to be the default when I designed the FileEdit and Parser classes, but clearly I forgot about it along the way.

### DIFF
--- a/tests/parser_tests/inverse.py
+++ b/tests/parser_tests/inverse.py
@@ -70,7 +70,7 @@ async def verify_inverse(parser):
                 ],
                 is_creation=False,
                 is_deletion=False,
-                rename_file_path=Path("file2.txt"),
+                rename_file_path=Path("file2.txt").resolve(),
             ),
         ],
     )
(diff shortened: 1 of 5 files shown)


granawkins (approved):
Looks fine to me and tests all passing. I like using absolute paths here instead of stubs.

tests/parser_tests/inverse.py
```diff
@@ -70,7 +70,7 @@ async def verify_inverse(parser):
                 ],
                 is_creation=False,
                 is_deletion=False,
-                rename_file_path=Path("file2.txt"),
+                rename_file_path=Path("file2.txt").resolve(),
```
jakethekoenig, line 73: Why does this need resolve but not the other paths in the test?

PCSwingle: Oh, I guess you just did cwd / "file_name" for the other paths; resolve just makes it an abs path, but I'll change it to cwd / for consistency

jakethekoenig: Right. Sorry I probably could have figured that out myself but I was reviewing from my phone. Thanks for fixing this!

PR #184, opened 2023-10-18
fix context checking in fileedit, remove test.conftest imports

### LINKED ISSUE #173: Warning of "Attempted to edit file not in context" should be caught before accepting changes
If the model tries to edit a file that's not in context, we don't do anything about it until after the user accepts changes. And then we say "Changes applied." even if nothing has been applied (if all the edits were for out-of-context files).

Really the root issue is the model trying to edit files that aren't in context, but still we could handle this better

image

PCSwingle:
Fixes #173. I'm going to work on adding SessionContext after this, since it clearly would help with circular imports quite a bit

### DIFF
--- a/mentat/parsers/file_edit.py
+++ b/mentat/parsers/file_edit.py
@@ -66,8 +66,12 @@ class FileEdit:
     rename_file_path: Path | None = attr.field(default=None)
 
     async def is_valid(self) -> bool:
+        # TODO: Remove this when we create SessionContext
+        # This fixes a circular import
+        from mentat.code_context import CODE_CONTEXT
+
         stream = SESSION_STREAM.get()
-        code_file_manager = CODE_FILE_MANAGER.get()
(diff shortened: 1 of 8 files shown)


mentat/parsers/file_edit.py
```diff
@@ -83,7 +87,7 @@ async def is_valid(self) -> bool:
                     f"File {rel_path} does not exist, canceling all edits to file."
                 )
                 return False
-            elif rel_path not in code_file_manager.file_lines:
+            elif self.file_path not in code_context.include_files:
```
granawkins, line 90: When auto-context gets kicking we'll want a custom method for this (probably in code_context) that says:
- Check if there's at least one feature in code_context.features matching the path
- Check that one of those features contains all the lines flagged for editing here.

Just a heads up for now.

granawkins (approved):

PR #185, opened 2023-10-19
add session context and remove old contextvars

### LINKED ISSUE #131: Add SessionContext
followup from this discussion: https://github.com/AbanteAI/mentat/pull/123#discussion_r1346638556

PCSwingle:
Fixes #131; adds sessioncontext, which will help us not have circular imports.

### DIFF
--- a/mentat/code_context.py
+++ b/mentat/code_context.py
@@ -8,6 +8,7 @@ from typing import Optional
 import attr
 
 from mentat.session_context import SESSION_CONTEXT
+from mentat.session_stream import SessionStream
 
 from .code_file import CodeFile, CodeMessageLevel, count_feature_tokens
 from .code_map import check_ctags_disabled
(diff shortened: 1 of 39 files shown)


granawkins (approved):
Looks good to me. No changes needed but a couple thoughts I had while reading:
- A lot of files' imports have a `from mentat.session..` block followed by a `from .session..` block. These could be combined but doesn't need to happen here. I'd say remove `mentat` from imports where possible, if nothing else so our code has fewer tokens.
- If a function uses only one context variable, I think it's fine to say `stream = SESSION_CONTEXT.get().stream`, instead of having two lines. This is of course preference, but I do see it in a couple places here.

PCSwingle:
> Looks good to me. No changes needed but a couple thoughts I had while reading:
> 
> * A lot of files' imports have a `from mentat.session..` block followed by a `from .session..` block. These could be combined but doesn't need to happen here. I'd say remove `mentat` from imports where possible, if nothing else so our code has fewer tokens.
> * If a function uses only one context variable, I think it's fine to say `stream = SESSION_CONTEXT.get().stream`, instead of having two lines. This is of course preference, but I do see it in a couple places here.

I'll add an issue for the imports, because it's definitely true that we need some formatter for standardized imports

waydegg (approved):

PR #186, opened 2023-10-19
Auto exclude

granawkins:
This PR fixes the context_benchmark by setting auto_tokens manually (it's default 0 now). 

I also realize, as I add benchmarks, that there should be a way to manually exclude files from auto-context, at least in the short term. I've just applied the same `exclude_paths` arg in this PR. 

There could be a scenario where you want to exclude something from your *included* files, but still let auto-context add it. In this case I would say we have a separate arg, like `exclude-auto`, but I thought repurposing is cleaner and those scenarios will be rare.

### DIFF
--- a/mentat/include_files.py
+++ b/mentat/include_files.py
@@ -76,16 +76,13 @@ def abs_files_from_list(paths: list[Path], check_for_text: bool = True):
     return files_direct, files_from_dirs, invalid_paths
 
 
-def get_include_files(
-    paths: list[Path], exclude_paths: list[Path]
-) -> tuple[Dict[Path, CodeFile], list[str]]:
-    """Returns a complete list of text files in a given set of include/exclude Paths."""
+def get_exclude_files(exclude_paths: list[Path]) -> list[Path]:
+    """Get the set of excluded files from the given exclude paths."""
     git_root = GIT_ROOT.get()
     config = CONFIG_MANAGER.get()
 
-    paths = expand_paths(paths)
+    # Paths excluded by Session
     exclude_paths = expand_paths(exclude_paths)
-
     excluded_files_direct, excluded_files_from_dirs, _ = abs_files_from_list(
         exclude_paths, check_for_text=False
     )
@@ -93,6 +90,7 @@ def get_include_files(
         map(lambda f: f.path, excluded_files_direct | excluded_files_from_dirs)
     )
 
+    # Paths excluded by config
     glob_excluded_files = set(
         Path(os.path.join(git_root, file))
         for glob_path in config.file_exclude_glob_list()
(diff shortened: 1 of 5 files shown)


mentat/include_files.py
```diff
@@ -76,23 +76,21 @@ def abs_files_from_list(paths: list[Path], check_for_text: bool = True):
...
-
     excluded_files_direct, excluded_files_from_dirs, _ = abs_files_from_list(
         exclude_paths, check_for_text=False
     )
     excluded_files = set(
         map(lambda f: f.path, excluded_files_direct | excluded_files_from_dirs)
     )
 
+    # Paths excluded by config
     glob_excluded_files = set(
```
PCSwingle, line 94: Right now, the config's glob_excluded_files only applied to files added by directory (so you can put in in *.json for example in your config to exclude all json files, but if you add a json file manually it will still work). In this PR you kind of merge both --exclude and the glob exclude so that they both only apply to directories, whereas before the --exclude argument would take precedence over manually included files; it might seem weird that we make the --exclude take preference over files you add individually one by one, but because zsh expands globs in the shell, if you make a glob include every file you're including is individual, so now there's no way to exclude files from an include glob (before, the --exclude argument was in place to do exactly this). I do think this is all overly convoluted, so I would be ok with this change (even though I think it's useful to be able to exclude parts of globs), but you need to change the help message for --exclude to make it clear that it only applies to manually added directories.

mentat/include_files.py
```diff
@@ -103,20 +101,35 @@ def get_include_files(
...
+) -> tuple[Dict[Path, CodeFile], list[str]]:
+    """Returns a complete list of text files in a given set of include/exclude Paths."""
+    paths = expand_paths(paths)
+    excluded_files = get_exclude_files(exclude_paths)
+
     files_direct, files_from_dirs, invalid_paths = abs_files_from_list(
         paths, check_for_text=True
     )
 
     # config glob excluded files only apply to files added from directories
```
PCSwingle, line 120: You would also probably want to change this comment to just all excluded files

mentat/include_files.py
granawkins: Aha - that was actually a good explanation and I see what you mean. This might be a good excuse to just have a separate argument for auto-context, like 'ignore'.

granawkins:
Reverted everything to do with `exclude_paths` to how it was before, and added a new arg `ignore_paths` to be applied to auto-context and search. This along with a test for the new arg and some misc tweaks to `context_benchmark`.

PCSwingle (approved):
Looks good to me; I will say that I think we might be overcomplicating things a bit having so many different exclude / ignore things, but I do also think we need this so I'm ok with merging it.

biobootloader (approved):
> Looks good to me; I will say that I think we might be overcomplicating things a bit having so many different exclude / ignore things, but I do also think we need this so I'm ok with merging it.

Agreed, it feels like we need a way for users to specify these things, but it is rather complicated! If anyone has a proposal for a more simple approach I'd like to hear it. As long as regular usage is simple and users only need to see these options when they want to do a more complex thing, that seems good to me. As long as it doesn't slow us down trying to support.

PR #194, opened 2023-10-19
catch shlex error

### LINKED ISSUE #192: Commands with arguments with unclosed quotes throw exceptions
For example `/search where's the parser` throws an exception. I propose we put a try catch [here](https://github.com/AbanteAI/mentat/blob/2a61ead72e841f74f8a993565ef65238003fc959/mentat/session_input.py#L35) and if there's an exception we can split on the first space and take every after as a singly string typed argument. Alternatively we could give a warning and it can be up to the user to rerun with quotes.

PCSwingle:
Fixes #192. I looked into shlex.split and I think it can only ever raise two errors: The unclosed quotation error you found, and a no escaped character error (just type a \ with no characters following it)

### DIFF
--- a/mentat/session_input.py
+++ b/mentat/session_input.py
@@ -26,15 +26,22 @@ async def collect_user_input(plain: bool = False) -> StreamMessage:
     # This fixes a circular import
     from mentat.commands import Command
 
+    stream = SESSION_STREAM.get()
+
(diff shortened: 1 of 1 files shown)


jakethekoenig (approved):
Thanks!

Issue #189, opened 2023-10-19
GPT added TODOs could trigger a different flow


jakethekoenig:
I find when I ask mentat to write very large blocks of code it often adds some TODOs (example attached). I don't think this is necessarily undesirable because:
1. gpt might think better step by step in this way
2. If the method outline it writes is bad it gives the user an opportunity to save tokens/time and ask for higher level changes before implementation

But it would be cool I think if mentat detected TODOs in the code and instead of `Apply these changes Y/n/i` it said something like: `Would you like me to implement the TODOs (Y), Just apply these changes (J) or Throw them out (N)`.
For example. I'm not sure what the exact phrasing should be.
[transcript_20231019_073927.log](https://github.com/AbanteAI/mentat/files/13044196/transcript_20231019_073927.log)

jakethekoenig:
Also side note on our html conversation viewers: github doesn't let you upload html files, presumably because the browser would try to open them and that could be confusing or a security risk. It might be good to also support markdown for easy sharing.

Issue #188, opened 2023-10-19
Mentat created files are not automatically added to context


jakethekoenig:
I asked mentat to create a file. It left some todos so I asked it to fill them in. But then it gave a warning `Attempted to edit file $file not in context`.

PR #195, opened 2023-10-19
Git diff parsing

jakethekoenig:
A parser for the git diff format is introduced. And a test for the translation script between block, replacement and git_diff formats is introduced.

### DIFF
--- a/mentat/parsers/parser.py
+++ b/mentat/parsers/parser.py
@@ -235,6 +235,7 @@ class Parser(ABC):
                                 cur_file_edit.rename_file_path = (
                                     file_edit.rename_file_path
                                 )
+                            cur_file_edit.replacements.extend(file_edit.replacements)
                             file_edit = cur_file_edit
 
                         # Print file header
--- a/scripts/translate_transcript.py
+++ b/scripts/translate_transcript.py
@@ -2,20 +2,19 @@
 import argparse
 import asyncio
 import json
-from unittest.mock import AsyncMock
 from pathlib import Path
+from unittest.mock import AsyncMock
 
 from mentat.code_file_manager import CODE_FILE_MANAGER, CodeFileManager
 from mentat.git_handler import GIT_ROOT
 from mentat.parsers.block_parser import BlockParser
+from mentat.parsers.git_parser import GitParser
 from mentat.parsers.parser import Parser
 from mentat.parsers.replacement_parser import ReplacementParser
 from mentat.parsers.split_diff_parser import SplitDiffParser
 from mentat.parsers.unified_diff_parser import UnifiedDiffParser
-from mentat.parsers.git_parser import GitParser
 from mentat.session_stream import SESSION_STREAM
 from mentat.utils import convert_string_to_asyncgen
-
 
 parser_map: dict[str, Parser] = {
     "block": BlockParser(),
--- a/scripts/translate_transcript.py
+++ b/scripts/translate_transcript.py
@@ -26,14 +26,11 @@ parser_map: dict[str, Parser] = {
 
 
 def translate_message(message: str, starting_parser, ending_parser) -> str:
-    if hasattr(starting_parser, "parse_string"):
-        parsedLLMResponse = starting_parser.parse_string(message)
-    else:
-        parsedLLMResponse = asyncio.run(
-            starting_parser.stream_and_parse_llm_response(
-                convert_string_to_asyncgen(message, 100)
-            )
+    parsedLLMResponse = asyncio.run(
+        starting_parser.stream_and_parse_llm_response(
+            convert_string_to_asyncgen(message, 100)
         )
+    )
     return ending_parser.file_edits_to_llm_message(parsedLLMResponse)
 
 
--- a/scripts/translate_transcript.py
+++ b/scripts/translate_transcript.py
@@ -3,6 +3,7 @@ import argparse
 import asyncio
 import json
 from unittest.mock import AsyncMock
+from pathlib import Path
 
 from mentat.code_file_manager import CODE_FILE_MANAGER, CodeFileManager
 from mentat.git_handler import GIT_ROOT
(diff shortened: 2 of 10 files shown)


tests/parser_tests/translate_scripts_test.py
```diff
@@ -0,0 +1,60 @@
...
+        return f.read()
+
+
+@pytest.fixture
+def git_diff_format():
+    with open("format_examples/git_diff.txt") as f:
+        return f.read()
+
+
+# This one doesn't pass. I think it's actually a bug with the parser as if I print out the final parsed
```
jakethekoenig, line 27: @PCSwingle I'll look into this tomorrow but I wanted to bring your attention to it because I thought you might just know what the problem is.

jakethekoenig:
I also have no idea why the `code_context` test is failing. I guess it's probably because I changed `testbed`.

scripts/translate_transcript.py
```diff
@@ -2,65 +2,78 @@
...
-CODE_FILE_MANAGER.set(CodeFileManager())
-SESSION_STREAM.set(AsyncMock())
-
-
 parser_map: dict[str, Parser] = {
     "block": BlockParser(),
     "replacement": ReplacementParser(),
     "split-diff": SplitDiffParser(),
     "unified-diff": UnifiedDiffParser(),
+    "git": GitParser(),
```
PCSwingle, line 24: I don't really think it should be added to this map; it isn't a parser, we don't have a prompt for it, and the user can't select it

PCSwingle: Oh whoops, somehow I completely missed what file this was in! Ignore this haha

testbed/format_examples/git_diff.txt
PCSwingle, line 1: Are you going to add any example of creating/deleting files (it would have a /dev/null instead of a/a or b/a)

scripts/translate_transcript.py
```diff
@@ -2,65 +2,78 @@
...
-)
-parser.add_argument("--git-root", type=str, default=".", help="Git root directory")
-args = parser.parse_args()
-GIT_ROOT.set(args.git_root)
 
-starting_parser = parser_map[args.starting_format]
-ending_parser = parser_map[args.ending_format]
+def translate_message(message: str, starting_parser, ending_parser) -> str:
+    if hasattr(starting_parser, "parse_string"):
+        parsedLLMResponse = starting_parser.parse_string(message)
```
PCSwingle, line 30: We could get rid of this hasattr if we made a stream_and_parse_llm_response that just collected the async generator back into a string and passed it to parse_string; just an idea though, if you don't like it no need to do it (it is a bit weird converting a string to a generator and back haha

tests/parser_tests/translate_scripts_test.py
PCSwingle: I'm looking into it. You can also mark tests like this with @pytest.mark.xfail to mark it as failing but still let everything pass

testbed/format_examples/git_diff.txt
jakethekoenig: I believe it has both. See later in the file. I think moves, creates and deleted are all correctly supported

scripts/translate_transcript.py
jakethekoenig: That's probably a good idea. I'll do that. Could also see giving parsers a parse_string method since it might be useful in other contexts and makes more sense

tests/parser_tests/translate_scripts_test.py
PCSwingle: Holy cow I found the bug and it's actually a massive bug; I can't believe this isn't caught by any of our tests. I'll push a commit soon that fixes it (super easy fix) and will make it obvious what the bug was haha. Thanks for finding this!!

scripts/translate_transcript.py
PCSwingle: Maybe, but I think that would take a lot of work making a parse_string method for all of the parsers, since it's so interlinked with streaming right now.

testbed/format_examples/git_diff.txt
PCSwingle: Ok I must be blind haha, I could've sworn I didn't see any /dev/null

mentat/parsers/parser.py
```diff
@@ -235,6 +235,7 @@ async def stream_and_parse_llm_response(
                                 cur_file_edit.rename_file_path = (
                                     file_edit.rename_file_path
                                 )
+                            cur_file_edit.replacements.extend(file_edit.replacements)
```
jakethekoenig, line 238: oof. Thanks for fixing!

scripts/translate_transcript.py
jakethekoenig: Yeah, you're right. I did it your way.

jakethekoenig:
> I also have no idea why the `code_context` test is failing. I guess it's probably because I changed `testbed`.

I went ahead and just changed the test. I think any time any file in testbed is edited the exact token counts should change. I wonder if we should fix the test to a git commit to prevent that.

jakethekoenig:
I tried to get gpt-4-32k (because I wanted to fit parser.py into context) to write this for me. It didn't work the first try and I actually implementing all the parser methods was more than we needed for this pr so I decided to do it myself. Interestingly it first wrote out some methods and added todos which is why I opened https://github.com/AbanteAI/mentat/issues/189 and https://github.com/AbanteAI/mentat/issues/188
[transcript_20231019_073927.log](https://github.com/AbanteAI/mentat/files/13056078/transcript_20231019_073927.log)

PCSwingle:
> > I also have no idea why the `code_context` test is failing. I guess it's probably because I changed `testbed`.
> 
> I went ahead and just changed the test. I think any time any file in testbed is edited the exact token counts should change. I wonder if we should fix the test to a git commit to prevent that.

Yeah, that test is definitely pretty brittle; I definitely think it would be good to completely remove the testbed in the test and run it on only like 2 files (but we don't need to do that in this pr)

PCSwingle (approved):
Looks good to me!

biobootloader:
I added an issue for improving the brittle CodeContext test: https://github.com/AbanteAI/mentat/issues/201

Nice work!

PR #198, opened 2023-10-19
fix abs path auto completions

### LINKED ISSUE #191: Autocomplete suggests full paths
If you run `mentat README.md` and then start typing `README.md` autocomplete will suggest the full path of the readme instead of the relative path.

PCSwingle:
Fixes #191. We still suggest paths relative to the git root (rather than just the file_name) because that path is what the model sees and outputs and will most likely be the best one to use in the prompt.


jakethekoenig (approved):
Thanks!

Issue #154, opened 2023-10-14
Investigate regression in `tests/record_benchmark.py`


biobootloader:
@waydegg noticed some of these aren't doing well. I ran a few times and saw that as well. Let's run several times and determine our current performance for each one and determine the most common causes of failure.

<img width="942" alt="image" src="https://github.com/AbanteAI/mentat/assets/128252497/18742766-53ca-4a15-b940-8a73a8f528d0">

biobootloader:
Related: https://github.com/AbanteAI/mentat/issues/155

PCSwingle:
I turned off codemap and put auto tokens to 0 and it immediately started working most of the time

biobootloader:
> I turned off codemap and put auto tokens to 0 and it immediately started working. It failed both times I tried at first, and succeeded all 4 times I tried after turning off codemap and autotokens

it may not even realize which file it's supposed to be working in...

biobootloader:
Two possible solutions to try:

First, from @granawkins : 

>How about a tag after the fname saying who added it, and an explanation in the prompt. e.g.:
>"Your input will consist of a user request, some user-included code files (filenames appended ' USER'), some >automatically selected files (appended ' AUTO'), and sometimes the git diff of current code files.
>...
>mentat/conversation.py AUTO
>\<cmap>
>mentat/code_context.py USER
>\<code>

Second:

Add something like a list of user-added files as part of the beginning prompt, like "The user has run the system, specifically requesting to include files A, B, C in your context. User requests may implicitly refer to these files or their content. Other files and information are added automatically by the system, and may be useful to you."

PCSwingle:
I honestly think the best solution is to not have any auto-tokens or codemap by default; it increases the token size (and therefore cost), and it's going to distract it from what the user actually wanted it to focus on. Instead, I think we should focus on having the model determine what files it needs (via the codemap and possibly embeddings) in a separate query with separate conversation history (so nothing bleeds over to the main conversation) and then tell the user that the model added these files before we show it's actual result. The workflow would be something like this:

On mentat start: Get user provided files (multifile_calculator/calculator.py) -> user gives query (add exp and log to calculator) -> we run gpt with codemap (and/or embeddings) to determine files needed for this query (I need multifile_calculator/operations.py) -> Tell user 'Model added operations.py to context' -> Run and stream main model response (has no idea that operations.py was added by previous model, has no idea what a codemap is or embeddings are).

Let me know what you think @biobootloader and @granawkins

[...]

PR #199, opened 2023-10-19
update diff_context and tests to use absolute path

granawkins:
This addresses an issue in the latest comment in #154. We recently switched from relative to absolute paths in several places, and diff_context was overlooked.

### DIFF
--- a/mentat/diff_context.py
+++ b/mentat/diff_context.py
@@ -105,10 +105,10 @@ class DiffContext:
     @property
     def files(self) -> list[Path]:
         if self._files_cache is None:
+            git_root = GIT_ROOT.get()
             if self.target == "HEAD" and not check_head_exists():
                 return []  # A new repo without any commits
-            diff_files = get_files_in_diff(self.target)
-            self._files_cache = [f.resolve() for f in diff_files]
+            self._files_cache = [git_root / f for f in get_files_in_diff(self.target)]
         return self._files_cache
 
     @classmethod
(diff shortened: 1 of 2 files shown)


mentat/diff_context.py
```diff
@@ -107,7 +107,8 @@ def files(self) -> list[Path]:
         if self._files_cache is None:
             if self.target == "HEAD" and not check_head_exists():
                 return []  # A new repo without any commits
-            self._files_cache = get_files_in_diff(self.target)
+            diff_files = get_files_in_diff(self.target)
+            self._files_cache = [f.resolve() for f in diff_files]
```
PCSwingle, line 111: We need to use git_root / f, not f.resolve() (since these paths are relative to the git root, not the cwd). I also think it would be cleaner to do it in one line (instead of making this diff_files varialbe), but that's up to you

tests/diff_context_test.py
```diff
@@ -6,7 +6,10 @@
 
 from mentat.diff_context import DiffContext
 
-rel_path = Path("multifile_calculator/operations.py")
+
+@pytest.fixture
+def abs_path(temp_testbed):
+    return Path(temp_testbed).resolve() / "multifile_calculator/operations.py"
```
PCSwingle, line 12: I feel like this shouldn't be a fixture; how about we just calculate it in each test? Also, temp_testbed is already absolute, so no need to resolve it

granawkins:
Good calls - took care of both of those.

PCSwingle (approved):
Thanks for fixing this!

PR #204, opened 2023-10-20
fix codecontext test to not be brittle

### LINKED ISSUE #201: Improve CodeContext Test to decrease brittleness
The exact token counts change when files in the testbed change, which means this test has to be updated often:

https://github.com/AbanteAI/mentat/pull/195/files#diff-d4964c636201547e3ac8f7f6af61b0895b68658a6f669849e5d1561c6056b53dL282-R284

Maybe we could use a specific git commit for this test, or only use a couple of files?

PCSwingle:
Fixes #201

### DIFF
--- a/.github/workflows/lint_and_test.yml
+++ b/.github/workflows/lint_and_test.yml
@@ -37,6 +37,15 @@ jobs:
 
     steps:
       - uses: actions/checkout@v3
+      - name: Install universal-ctags (Ubuntu)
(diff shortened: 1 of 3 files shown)


granawkins (approved):
Yes - good idea, no notes.

granawkins:
Added ctags to the github actions to get the tests passing.

biobootloader (approved):

PR #205, opened 2023-10-20
Add "Agent Mode"

biobootloader:
This isn't something I expect users to use, but I thought it could be useful to experiment with having GPT-4 replace the human using Mentat. It's a pretty basic, limited version right now that works with just a single file.

Perhaps we could iterate on this and get something useful for automatic testing / benchmarking of Mentat? I.e. we could have it run several times and record things Mentat does wrong for us so we can automatically identify common issues. Or perhaps it'd be useful for benchmarking - seeing how frequently / in how many tries does it take to accomplish various tasks?

Here's an example where Mentat created some duplicate lines and GPT-4 noticed and asked for them to be fixed (Magenta text are the GPT-4 responses that are directing Mentat):

image

Sometimes it gets confused and thinks it is Mentat:

image

### DIFF
--- a/mentat/python_client/client.py
+++ b/mentat/python_client/client.py
@@ -32,6 +32,25 @@ class PythonClient:
         self.started = False
         self._accumulated_message = ""
 
+    async def wait_for_input_request(self):
+        assert isinstance(self.session, Session), "Client is not running"
+
+        input_request_message = await self.session.stream.recv("input_request")
+        await asyncio.sleep(0.01)
+
+        messages = self._accumulated_message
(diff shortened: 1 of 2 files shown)


PCSwingle:
This looks pretty awesome! As a heads up, I made a few changes to the python client in a recent PR, so after you merge you might have to fix a few conflicts there (and hopefully you won't have to edit it at all)

biobootloader:
Closing as https://github.com/AbanteAI/mentat/pull/343 replaces!

Issue #202, opened 2023-10-20
Import formatter


PCSwingle:
Right now we mix a lot of mentat._file_ and ._file_; we need to get a formatter to standardize our imports.

PCSwingle:
After looking into it a bit, the only library I could find was this small one here https://github.com/MarcoGorelli/absolufy-imports, which I'm not sure is being maintained anymore.

biobootloader:
If we decide on a format maybe we don't need an auto-formatter for this. Should we just make every absolute imports?

jakethekoenig:
@biobootloader I prefer absolute aesthetically but @granawkins had a good point that relative saves tokens.

PCSwingle:
I think we should prefer absolute; I don't think saving tokens should matter (although it does bring up a good point that it might be useful to trim imports from files)

PCSwingle:
> If we decide on a format maybe we don't need an auto-formatter for this. Should we just make every absolute imports?

I also still think it would be good to have an auto-formatter; there's no way we catch them all in the future without one

biobootloader:
Let's go with absolute. Costs more tokens but the clarity might help models more too.

Auto-formatter would be good if it's not a hassle

biobootloader:
Thanks to @granawkins for standardizing all our imports. Leaving this ticket open to add a github action to keep them that way

PR #207, opened 2023-10-21
standardize import format to absolute

granawkins:
Addresses #202. 

As far as auto-formatting, I used [absolutify](https://github.com/MarcoGorelli/absolufy-imports) locally and it worked great. This PR is the output of that, followed by isort. However it's not maintained, and it doesn't have a 'check' option for our gh-action.


biobootloader (approved):
Thanks

> it doesn't have a 'check' option for our gh-action

I'm sure we could just run it and then check if something has changed some other way (run git diff?). But no need to set that up as part of this!

PR #208, opened 2023-10-21
rename code_file to code_feature, CodeFile to CodeFeature

granawkins:


biobootloader (approved):
I don't love the name `CodeFeature` but I haven't thought of anything better

PR #209, opened 2023-10-23
Function-level Features

granawkins:
- Use `ctags` to split files into smaller chunks
- Return chunks rather than full files for search
- Handful of misc related fixes

In the next PR, I'll apply this to generating auto-features in `get_code_message`.

### DIFF
--- a/mentat/code_context.py
+++ b/mentat/code_context.py
@@ -48,18 +48,14 @@ def _get_all_features(
 ) -> list[CodeFeature]:
     """Return a list of all features in the git root with given properties."""
 
-    _features = list[CodeFeature]()
+    all_features = list[CodeFeature]()
     for path in get_non_gitignored_files(git_root):
         abs_path = git_root / path
-        if (
-            abs_path in include_files
-            or abs_path.is_dir()
-            or not is_file_text_encoded(abs_path)
-        ):
+        if abs_path.is_dir() or not is_file_text_encoded(abs_path):
             continue
 
         diff_target = diff_context.target if abs_path in diff_context.files else None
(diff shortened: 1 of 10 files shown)


mentat/code_feature.py
```diff
@@ -1,42 +1,83 @@
...
-from mentat.code_map import get_code_map
+from mentat.code_map import get_code_map, get_ctags
 from mentat.diff_context import annotate_file_message, parse_diff
 from mentat.git_handler import get_diff_for_file
+from mentat.interval import Interval, parse_intervals
 from mentat.llm_api import count_tokens
 from mentat.session_context import SESSION_CONTEXT
 from mentat.utils import sha256
 
+MIN_INTERVAL_LINES = 10
```
waydegg, line 18: Any reason why "10" (arbitrary number or works best for a specific reason)? Just curious

waydegg (approved):

mentat/code_context.py
```diff
@@ -34,6 +39,53 @@ class CodeContextSettings:
...
+        abs_path = git_root / path
+        if (
+            abs_path in include_files
+            or abs_path.is_dir()
+            or not is_file_text_encoded(abs_path)
+        ):
+            continue
+
+        diff_target = diff_context.target if abs_path in diff_context.files else None
+        user_included = path in include_files
```
PCSwingle, line 62: Won't this always be false? (first of all because of line 55 and second of all because include_files has absolute paths as keys)?

granawkins: You're right, good catch. I've added a test for this too.

PR #212, opened 2023-10-23
Fix Interactive Mode suggestion count

### LINKED ISSUE #153: Bug in interactive mode in what we send to model
After the user selects changes from interactive mode, we tell the model "User accepted x/y of your changes"; the x and y are incorrect.

waydegg:
Fixes #153

### DIFF
--- a/mentat/code_edit_feedback.py
+++ b/mentat/code_edit_feedback.py
@@ -35,14 +35,16 @@ async def get_user_feedback_on_edits(
                 MessageRole.User, "User chose not to apply any of your changes."
             )
         case "i":
-            suggested_edits_count = sum([len(fe.replacements) for fe in file_edits])
             edits_to_apply = await _user_filter_changes(file_edits)
-            conversation.add_message(
-                MessageRole.User,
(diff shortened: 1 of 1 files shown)


PCSwingle:
Hmm, thinking about this further I'm thinking we might just want to get rid of this line; what does telling the model that x of it's changes were accepted even mean to it? How does it help it? Additionally, with some formats the number of 'changes' isn't very clear; for example, with this PR we only look at replacements, but with the original block format a 'change' to the model also includes file creations, deletions, and renames (while in some formats those are combined into a single change). This also wouldn't work too well for formats like the unified diff format (where multiple replacements sometimes have to be created by one change). @biobootloader what do you think? I'm thinking we might just say something like 'the user selected some of your changes'

waydegg:
Yeah I think GPT4 is smart enough to infer what changes were accepted and which ones weren't. Also just giving an "X out of Y changes were accepted" doesn't really help the model in any way.

biobootloader (approved):
I'm fine either way, X/Y or "some were accepted".

PR #214, opened 2023-10-24
Benchmark generation from git commits

jakethekoenig:
The commit to transcript script can now make benchmarks. One is added as an example to our existing context benchmark script. Flags which control which benchmarks run are added to the context benchmark.


PCSwingle (approved):
Looks good to me! How does it perform on these benchmarks?

jakethekoenig:
> Looks good to me! How does it perform on these benchmarks?

I'm not sure the context benchmark is doing much right now because we don't have much of a context generator.

I need to write a benchmark that sends the output to GPT-4 for grading. Right now the "expected_edits" field isn't doing anything.

PR #217, opened 2023-10-25
Remove git dependency for CodeMap

waydegg:
This PR removes the need for CodeMap to be run on a file inside of a git repository. 

This is part of a larger effort to allow Mentat Sessions to be started in a non-git directory, which is required by the VSCode extension executable along with just being a very nice feature to have (users can open Mentat anywhere on the filesystem).

### DIFF
--- a/tests/code_feature_test.py
+++ b/tests/code_feature_test.py
@@ -6,15 +6,20 @@ from mentat.code_feature import CodeFeature, CodeMessageLevel, split_file_into_i
 
 def test_split_file_into_intervals(temp_testbed, mock_session_context):
     with open("file_1.py", "w") as f:
-        f.write(dedent("""\
+        f.write(
+            dedent(
+                """\
             def func_1(x, y):
                 return x + y
             
             def func_2():
                 return 3
-            """))
+            """
+            )
+        )
     code_feature = CodeFeature(Path("file_1.py"), CodeMessageLevel.CODE)
     interval_features = split_file_into_intervals(temp_testbed, code_feature, 1)
+
     assert len(interval_features) == 2
 
     interval_1 = interval_features[0].intervals[0]
--- a/tests/code_feature_test.py
+++ b/tests/code_feature_test.py
@@ -6,17 +6,13 @@ from mentat.code_feature import CodeFeature, CodeMessageLevel, split_file_into_i
 
 def test_split_file_into_intervals(temp_testbed, mock_session_context):
     with open("file_1.py", "w") as f:
-        f.write(
-            dedent(
-                """\
+        f.write(dedent("""\
             def func_1(x, y):
                 return x + y
             
             def func_2():
                 return 3
-            """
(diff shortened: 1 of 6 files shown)


tests/code_feature_test.py
```diff
@@ -6,15 +6,20 @@
...
+                """\
             def func_1(x, y):
                 return x + y
             
             def func_2():
                 return 3
-            """))
+            """
+            )
+        )
```
waydegg, line 19: Black did this for me automatically as this file wasn't formatted earlier.

tests/conftest.py
```diff
@@ -154,6 +154,7 @@ def mock_setup_api_key(mocker):
 # Despite not using any awaits here, this has to be async or there won't be a running event loop
 @pytest_asyncio.fixture()
 async def _mock_session_context(temp_testbed):
+    # TODO make this `None` if there's no git (SessionContext needs to allow it)
```
waydegg, line 157: Will fix this in a later PR that removes the requirement of needing git for `SessionContext`

waydegg:
Not running `black` here since it reformats ~22 files...

tests/code_feature_test.py
granawkins: When I run `black --preview .`, it just changes this back, and no other reformats [🤔](https://emojipedia.org/thinking-face)

tests/conftest.py
```diff
@@ -224,35 +225,59 @@ def add_permissions(func, path, exc_info):
...
+def temp_testbed(monkeypatch, get_marks, no_git):
     # create temporary copy of testbed, complete with git repo
     # realpath() resolves symlinks, required for paths to match on macOS
     temp_dir = os.path.realpath(tempfile.mkdtemp())
     temp_testbed = os.path.join(temp_dir, "testbed")
     os.mkdir(temp_testbed)
 
-    # Initialize git repo
-    run_git_command(temp_testbed, "init")
+    if no_git is not True:
```
granawkins, line 258: Would it be simpler to add `no_git` the way `clear_testbed` is added, and then just get the value from get_marks? 

Then instead of `pytest.parametrize(...)`, just `pytest.mark.no_git`.

granawkins (approved):
Runs fine, looks good with just a couple comments.

tests/conftest.py
PCSwingle: I agree, I think this would be a lot cleaner

tests/code_feature_test.py
waydegg: Oof might be something on my end then

tests/conftest.py
waydegg: Nice yea will change this

tests/code_feature_test.py
waydegg: Ah looks like I wasn't running black in "preview" mode

PR #219, opened 2023-10-25
Config refactor

jakethekoenig:
Sorry this ended up being such a large PR. I wanted to do this before I added a couple of settings I wanted but looking at it I'm not 100% sure it was worth it to do right now.

Changes:
* The config is made by clients before being passed to Session. The one thing I didn't like about this change is it makes it hard to know what to do with error messages. 
* The config adds its fields to argparser so new settings can be added in one place and be settable by flag or file.
* The ConfigManager is now just Config
* Some things were importing mentat_dir_path from config but it's defined in utils so I fixed that.
* Three fields maintained by CodeContext are now in the Config. As the CodeContextSettings only had 2 attributes I got rid of it.
* I added a temperature flag.

Things I would like to do in future PRs:
* A `/config` command to change the config (I was going to do that in this PR but it ended up being a much bigger change than I anticipated)
* `file_exclude_glob_list` and `exclude` are now substantially similar. I'd like to consolidate them.


mentat/config.py
```diff
@@ -0,0 +1,174 @@
...
+                    )
+
+    def error(self, message: str) -> None:
+        self._errors.append(message)
+        try:
+            self.send_errors_to_stream()
+        except LookupError:
+            pass
+
+    def send_errors_to_stream(self):
```
PCSwingle, line 169: I agree this error stuff feels pretty weird; I noticed you call this function when the Session is created (right after SessionContext is set); why don't we just only call this function there? That way we don't need any of this except LookupError stuff.

mentat/session.py
```diff
@@ -84,6 +76,9 @@ def __init__(
         )
         SESSION_CONTEXT.set(session_context)
 
+        # Functions that require stream
```
PCSwingle, line 79: Requiring the stream is the same as requiring the session_context, so I think this would fit under the other comment

PCSwingle (approved):
This is super awesome! Thanks for making this change Jake! I left one comment for you to look at regarding the config errors, but other than that it looks great!

mentat/config.py
jakethekoenig: I did it this way because the config might have errors later in it's life cycle when the stream is initialized. This current version can't but when there's a `/config` command there could be type or setting dne errors.

jakethekoenig: I don't really like the try/catch though. Mind if I set `SESSION_CONTEXT` to none right after it's declaration and then check that instead of catching an exception?

PR #221, opened 2023-10-25
Config Command

jakethekoenig:
* You can set or get a config with `/config`
* I realized conversation was holding onto its own copy of model preventing it being changed. It now gets it from the config as necessary. (Perhaps we should go all the way and have the llm_api be responsible for choosing the model? I could imagine wanting to keep track of which model generated a message for the future though).
* I updated the use_embeddings comment so users know they don't need to restart mentat.
* Note attr is taking care of type checking so we can't end up with illegal config properties after a call to `/config`
* Some commands streamed trailing new lines which I removed. UX note: maybe instead of a yellow error on a `/command` that has nothing to do without an argument we should print the commands help message instead.

Please share UX ideas. It's possible there should be a `/save-config` command or a `/set` command that writes the config change to your .mentat_config.json


mentat/commands.py
```diff
@@ -338,3 +341,48 @@ def argument_names(cls) -> list[str]:
...
+                    if attr.fields_dict(Config)[setting].metadata.get(
+                        "no_midsession_change"
+                    ):
+                        stream.send(
+                            f"Cannot change {setting} mid-session. Please restart"
+                            " Mentat to change this setting.",
+                            color="yellow",
+                        )
+                        return
+                    setattr(config, setting, value)
```
PCSwingle, line 375: Could this set it to a type it shouldn't be? I also thing the Config should have functions for getting and setting config attributes instead of using setattr and getattr

jakethekoenig: I should read the docs for attrs more carefully but I believe this is safe and will use the conversions I specified or the obvious ones e.g. if you try `/config temperature lol` you get the error message: `Error processing command arguments: could not convert string to float: 'lol'`

jakethekoenig: It'd probably be a good idea to add validators though because you can set temperature to a number outside 0 to 1.

jakethekoenig: I added validators for format, temperature, maximum_context and auto_context.

PCSwingle: I'm pretty sure setattr and getattr have nothing to do with attrs; they're just built into python. Either way, looks good to me!

PCSwingle (approved):

mentat/commands.py
jakethekoenig: Yeah `setattr` and `getattr` are built in but somehow when you use them on a class with `attr` attributes `attr` is able to run its converts and validaters and either change what you set it to (like automatically calling int) or throw an error. I should probably read the attrs source to learn how it works.

PR #222, opened 2023-10-27
Run github actions on all pushes / PRs

biobootloader:
Github actions are free since it's a public repo, this will help people catch things faster as @granawkins suggested

### DIFF
--- a/.github/workflows/lint_and_test.yml
+++ b/.github/workflows/lint_and_test.yml
@@ -4,12 +4,13 @@
 name: lint and test
 
 on: [push, pull_request]
+
 env:
   SHOULD_RUN_JOB: ${{ github.event_name != 'pull_request' || github.event.pull_request.head.repo.full_name != github.event.pull_request.base.repo.full_name }}
 
 jobs:
   ruff-black-isort:
-    if: $SHOULD_RUN_JOB
+    if: ${{ env.SHOULD_RUN_JOB }}
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v3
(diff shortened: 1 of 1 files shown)


biobootloader:
hmm annoying that now there are duplicated checks on PRs:
image

PCSwingle (approved):

.github/workflows/lint_and_test.yml
```diff
@@ -3,13 +3,7 @@
...
 name: lint and test
 
-on:
-  push:
-    branches:
-      - main
-  pull_request:
-    branches:
-      - main
+on: [push, pull_request]
```
granawkins, line 6:
```suggestion
on:
  push:
    branches:
      - main
  pull_request:
    branches:
      - main
  workflow_dispatch:
    inputs:
      branch:
        description: 'Branch to run on'
        required: true
        default: 'main'
```
Saw your comment about duplicating tests, that's not ideal.

What about instead allowing you to manually run the tests on a particular branch?

biobootloader: I found an alternative approach here: https://wildwolf.name/github-actions-how-to-avoid-running-the-same-workflow-multiple-times/

biobootloader:
I added these checks (found them here https://wildwolf.name/github-actions-how-to-avoid-running-the-same-workflow-multiple-times/). This should prevent the "pull-request" events from running unless the PR is from another repo. That should fix the duplicate checks being run on PRs that come from branches in this repo, but we'll see...

image

biobootloader:
well I tried several ways to not have to duplicate the check on each job but to no avail

PR #225, opened 2023-10-27
LLM_API goes to config for model

jakethekoenig:
* llm_api methods now have model as optional and use the config if unspecified.
* config holds onto parser. Taken out of session_context.
* Bug fix for boolean flags
* New flag to omit system prompt for fine tuned models
* Conversation changed so system prompt and parser can be changed midconversation.

### DIFF
--- a/mentat/config.py
+++ b/mentat/config.py
@@ -157,12 +157,14 @@ class Config:
         return config
 
     def load_namespace(self, args: Namespace) -> None:
-        for field in vars(args):
-            if hasattr(self, field) and getattr(args, field) is not None:
-                try:
-                    setattr(self, field, getattr(args, field))
-                except (ValueError, TypeError) as e:
-                    self.error(f"Warning: Illegal value for {field}: {e}")
+        for field in attr.fields(Config):
+            if field.name in args and field.name != "_errors":
+                value = getattr(args, field.name)
+                if value is not None and value != field.default:
+                    try:
+                        setattr(self, field.name, value)
+                    except (ValueError, TypeError) as e:
+                        self.error(f"Warning: Illegal value for {field}: {e}")
 
     def load_file(self, path: Path) -> None:
         if path.exists():
--- a/mentat/config.py
+++ b/mentat/config.py
@@ -6,23 +6,16 @@ from json import JSONDecodeError
 from pathlib import Path
 
 import attr
-from attr import validators
+from attr import converters, validators
 
 from mentat.git_handler import get_shared_git_root_for_paths
+from mentat.parsers.parser import Parser
 from mentat.parsers.parser_map import parser_map
 from mentat.session_context import SESSION_CONTEXT
 from mentat.utils import mentat_dir_path
 
 config_file_name = Path(".mentat_config.json")
 user_config_path = mentat_dir_path / config_file_name
-
-
-def parse_bool(s: str | bool | None) -> bool:
-    if isinstance(s, bool):
-        return s
-    if s is None:
-        return False
-    return s.lower() in ("true", "1", "t", "y", "yes")
 
 
 def int_or_none(s: str | None) -> int | None:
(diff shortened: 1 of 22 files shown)


mentat/config.py
```diff
@@ -53,16 +46,26 @@ class Config:
...
         },
-        validator=validators.optional(validators.in_(parser_map)),
+        converter=parser_map.get,  # pyright: ignore
+        validator=validators.instance_of(Parser),  # pyright: ignore
+    )
+    no_parser_prompt: bool = attr.field(
+        default=False,
+        metadata={
+            "description": (
+                "Whether to include the parser prompt in the system message. Only set"
```
PCSwingle, line 64: A bit of a nit picky change but I think this should read 'This should only be set to true for fine tuned models'; the first time I read it I interpreted it as 'This is only set to true for fine tuned models'

mentat/conversation.py
```diff
@@ -127,17 +90,28 @@ def add_message(self, role: MessageRole, message: str):
         """Used for adding messages to the models conversation"""
         self.messages.append({"role": role.value, "content": message})
 
+    def get_messages(self) -> list[dict[str, str]]:
+        """Returns the messages in the conversation. The system messsage may change throughout
```
PCSwingle, line 94: Typo

mentat/conversation.py
```diff
@@ -33,62 +32,26 @@ class MessageRole(Enum):
 class Conversation:
     max_tokens: int
 
-    def __init__(self, parser: Parser):
+    def __init__(self):
         self.messages = list[dict[str, str]]()
```
PCSwingle, line 36: Maybe self.messages should be private now that we have a function to get it?

mentat/llm_api.py
```diff
@@ -55,6 +55,12 @@ def is_test_environment():
     )
 
 
+def get_model():
```
PCSwingle, line 58: I feel like this function isn't necessary

mentat/llm_api.py
```diff
@@ -104,11 +110,31 @@ def warn_user(message: str, max_tries: int, details: Details):
...
     session_context = SESSION_CONTEXT.get()
+    stream = session_context.stream
     config = session_context.config
+    if model is None:
+        model = config.model
+    if not is_model_available(model):
+        raise MentatError(
+            f"Model {model} is not available. Please try again with a different model."
+        )
+    if "gpt-4" not in model:
```
PCSwingle, line 125: I think the main reason we sent this message at the start was because we only wanted to print it once. Maybe we should print this warning any time config.model is changed (including at the start)? This might be another reason to have a specific set_config function in the config (instead of using setattr)

mentat/llm_api.py
```diff
@@ -176,8 +209,31 @@ def model_context_size(model: str) -> Optional[int]:
         return None
 
 
-def model_price_per_1000_tokens(model: str) -> Optional[tuple[float, float]]:
+def maximum_context_size() -> int:
```
PCSwingle, line 212: I think it would make sense for this function to also take a model argument (instead of using session context) since it's very standalone

mentat/llm_api.py
```diff
@@ -176,8 +209,31 @@ def model_context_size(model: str) -> Optional[int]:
...
+    session_context = SESSION_CONTEXT.get()
+    config = session_context.config
+    model = config.model
+    model_maximum = model_context_size(model)
+    maximum_context = config.maximum_context
+    if maximum_context is None:
+        if model_maximum is None:
+            raise MentatError(
```
PCSwingle, line 220: How about we just print a warning instead of quitting?

mentat/llm_api.py
```diff
@@ -176,8 +209,31 @@ def model_context_size(model: str) -> Optional[int]:
...
+        else:
+            return model_maximum
+    else:
+        if model_maximum is None:
+            return maximum_context
+        else:
+            return min(model_maximum, maximum_context)
+
+
+def price_per_1000_tokens(model: str | None = None) -> Optional[tuple[float, float]]:
```
PCSwingle, line 233: I don't think it makes sense to allow None as an argument here

PCSwingle: This applies to all of the functions that have None as an option for model

PCSwingle (changes requested):
Super happy with this! It's definitely good to have the model be a global thing; I also like that you moved the parser to be part of the config. Left a few comments, but once those are resolved it should be good to merge!

granawkins:
With #228, it's possible we'd be calling multiple models on the same command: one for context-selection and one for the main response. I'm in favor of moving `model` to the config generally, but maybe we specify different models for different tasks?

jakethekoenig:
> With #228, it's possible we'd be calling multiple models on the same command: one for context-selection and one for the main response. I'm in favor of moving `model` to the config generally, but maybe we specify different models for different tasks?

That's a good point. And we're already sort of doing that with the embedding model. I think all the models mentat uses should go in the config. The default conversational one can be called `model`, the embedding model can be called `embedding-model` and the model you use in #228 can be called something like `select-features-model` or `build-context-model`. I guess this raises the question: why even have the llm_api aware of our conversational model default since it will only slightly simplify `conversation.py`

granawkins:
> I guess this raises the question: why even have the llm_api aware of our conversational model default since it will only slightly simplify `conversation.py`

Ya that also makes sense, and goes with @PCSwingle's comment avoiding too many `None`s.

jakethekoenig:
Going to split this PR up. First one: https://github.com/AbanteAI/mentat/pull/238

PR #227, opened 2023-10-29
Fix initializing mentat with a --diff argument

granawkins:
Crashing currently if you run e.g. `mentat --diff HEAD~1`

### DIFF
--- a/mentat/diff_context.py
+++ b/mentat/diff_context.py
@@ -131,7 +131,7 @@ class DiffContext:
                 self.name = "HEAD (last commit)"
                 return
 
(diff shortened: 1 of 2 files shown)


PCSwingle (changes requested):
@jakethekoenig create #230 which does the same thing as this PR but it also adds a test, so I think we should merge that one instead.

jakethekoenig:
Oops. Should have checked if anyone else had a PR before making mine.

biobootloader:
Thanks for catching this @granawkins, sorry it took a couple days to review!

PR #228, opened 2023-10-30
Auto context with llm

granawkins:
Add the ability to use llm for context selection in benchmarks. In future PRs we'll add a command/arg to use this in the actual mentat - this lays the groundwork for that.

- Move auto-context logic from `CodeContext._get_auto_features` to `mentat.auto_context.py`
- Define two algos: _greedy (reproduce current results) and _llm
- Use _llm algo in `get_git_log_transcripts` to build benchmark['expected_features'].

What this branch can do is:
1. Generate benchmarks using @jakethekoenig's script: `./scripts/git_log_to_transcripts.py --repo <gh-link>`
2. Run context_benchmarks against the created benchmarks: `pytest -s tests/benchmarks/context_benchmark.py --benchmarks`. This will print your precision/recall first with the greedy method, then llm.

### DIFF
--- a/mentat/auto_context.py
+++ b/mentat/auto_context.py
@@ -112,7 +112,6 @@ class LLMFeatureSelector(FeatureSelector, selector_name="llm"):
     ) -> list[CodeFeature]:
         session_context = SESSION_CONTEXT.get()
         config = session_context.config
-        git_root = session_context.git_root
 
         # Preselect as many features as fit in the context window
         model = config.feature_selection_model
--- a/mentat/auto_context.py
+++ b/mentat/auto_context.py
@@ -1,11 +1,13 @@
 import json
-from textwrap import dedent
+from pathlib import Path
 from typing import Optional, cast
 
 import openai
 
 from mentat.code_feature import CodeFeature, CodeMessageLevel
-from mentat.llm_api import count_tokens
+from mentat.errors import UserError
+from mentat.llm_api import count_tokens, model_context_size
+from mentat.prompts.prompts import read_prompt
 from mentat.session_context import SESSION_CONTEXT
 
 """
@@ -43,32 +45,9 @@ def select_features_greedy(
     return output
 
 
-feature_selection_model = "gpt-4"
-feature_selection_max_tokens = 8192
-feature_selection_temperature = 0.5
+feature_selection_prompt_path = Path("feature_selection_prompt.txt")
+feature_selection_prompt_training_path = Path("feature_selection_prompt_training.txt")
 feature_selection_response_buffer = 500
-feature_selection_prompt = dedent("""\
-    You are part of an automated coding system. 
-    Below you will see the line 'User Query:', followed by a user query, followed by the line 'Code Files:' and then a pre-selected subset of a codebase.
-    Your job is to select portions of the code which are relevant to answering that query.
-    The process after you will read the lines code you select, and then return a plaintext plan of action and a list of Code Edits.
-    {{training_prompt}}
-    Each item in the Code Files below will include a relative path and line numbers.
-    Identify lines of code which are relevant to the query.
-    Return a json-serializable list of relevant items following the same format you receive: <rel_path>:<start_line>-<end_line>,<start_line>-<end_line>.
-    It's important to include lines which would be edited in order to generate the answer as well as lines which are required to understand the context.
-    It's equally important to exclude irrelevant code, as it has a negative impact on the system performance and cost.
-    For example: if a question requires creating a new method related to a class, and the method uses an attribute of that
-    class, include the location for the edit as well as where the attribute is defined. If a typing system is used, include
-    the type definition as well, and the location of the expected import.
-    Make sure your response is valid json, for example:
-    ["path/to/file1.py:1-10,53-60", "path/to/file2.py:10-20"]
-    """)
-training_prompt = (
-    "You are currently being used to generate benchmarking data, so you'll be shown"
-    " those Code Edits as well in order to maximize your performance."
-)
-
 
 async def select_features_llm(
     features: list[CodeFeature],
@@ -79,22 +58,31 @@ async def select_features_llm(
     expected_edits: Optional[list[str]] = None,  # For benchmarks/training
 ) -> list[CodeFeature]:
     session_context = SESSION_CONTEXT.get()
+    config = session_context.config
     git_root = session_context.git_root
 
     # Preselect as many features as fit in the context window
+    model = config.feature_selection_model
+    context_size = model_context_size(model)
+    if context_size is None:
+        raise UserError(
+            f"Unknown context size for feature selection model: {config.feature_selection_model}"
+        )
+    system_prompt = read_prompt(feature_selection_prompt_path)
+    training_prompt = read_prompt(feature_selection_prompt_training_path)
     user_prompt_tokens = count_tokens(user_prompt, model)
-    system_prompt = feature_selection_prompt.format(
+    system_prompt = system_prompt.format(
         training_prompt=training_prompt if expected_edits else ""
     )
-    system_prompt_tokens = count_tokens(system_prompt, feature_selection_model)
+    system_prompt_tokens = count_tokens(system_prompt, config.feature_selection_model)
     preselect_max_tokens = (
-        feature_selection_max_tokens
+        context_size
         - system_prompt_tokens
         - user_prompt_tokens
         - feature_selection_response_buffer
     )
     preselected_features = select_features_greedy(
-        features, preselect_max_tokens, feature_selection_model, levels
+        features, preselect_max_tokens, model, levels
     )
 
     # Ask the model to return only relevant features
@@ -113,7 +101,7 @@ async def select_features_llm(
     response = await openai.ChatCompletion.acreate(  # type: ignore
         model=model,
         messages=messages,
-        temperature=feature_selection_temperature,
+        temperature=config.temperature,
     )
 
     # Create output features from the response
--- a/mentat/auto_context.py
+++ b/mentat/auto_context.py
@@ -1,4 +1,5 @@
 import json
+from abc import ABC, abstractmethod
 from pathlib import Path
 from typing import Optional, cast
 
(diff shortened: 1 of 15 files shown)


mentat/auto_context.py
```diff
@@ -0,0 +1,150 @@
...
+    Each item in the Code Files below will include a relative path and line numbers.
+    Identify lines of code which are relevant to the query.
+    Return a json-serializable list of relevant items following the same format you receive: <rel_path>:<start_line>-<end_line>,<start_line>-<end_line>.
+    It's important to include lines which would be edited in order to generate the answer as well as lines which are required to understand the context.
+    It's equally important to exclude irrelevant code, as it has a negative impact on the system performance and cost.
+    For example: if a question requires creating a new method related to a class, and the method uses an attribute of that
+    class, include the location for the edit as well as where the attribute is defined. If a typing system is used, include
+    the type definition as well, and the location of the expected import.
+    Make sure your response is valid json, for example:
+    ["path/to/file1.py:1-10,53-60", "path/to/file2.py:10-20"]
```
jakethekoenig, line 65: This is just an idea for something to try in the future: We could experiment with showing one file at a time (or rather per prompt, we could do it simultaneously) and asking if this file has relevant information for achieving the user request. Upsides:
1. The simpler we keep it the more likely gpt-3.5 is to be able to do it.
2. We can ask about more files than fit in context

Downsides:
1. Maybe seeing everything will help understand what is necessary to understand
2. We have to pay for the user request every time

tests/benchmarks/context_benchmark.py
jakethekoenig, line 1: I often get results that aren't between 0 and 1. Does that make sense?
```

Conversation viewer (#152)
Add a conversation viewer that allows users to view the conversation as seen by the AI. Fix scroll issue in the viewer and use monospace font. Also, fix a double transcript bug.
Embedding batch 1/2...
Embedding batch 2/2...
Cost: $0.0016
True Positive:  0.000
False Positive: 991.000
False Negative: 489.000
Precision:      0.000   | How many selected features are relevant?
Recall:         0.000   | How many relevant features are selected?
True Positive:  88.000
False Positive: 591.000
False Negative: 401.000
Precision:      0.130   | How many selected features are relevant?
Recall:         0.180   | How many relevant features are selected?
F1:             0.151   | Weighted average of precision and recall
```

scripts/git_log_to_transcripts.py
```diff
@@ -140,19 +140,56 @@ async def translate_commits_to_transcripts(repo, count=10):
...
+            code_context.use_llm = True
+            await code_context.get_code_message(
+                prompt, model, max_context_tokens, benchmark["expected_edits"]
+            )
+
+            git_root_length = len(str(git_root)) + 1
+            selected_features = [
+                f.ref()[git_root_length:] for f in code_context.features
+            ]
+            selected_files = {f.split(":")[0] for f in selected_features}
```
jakethekoenig, line 169: When I run the benchmark I get errors like: `E       FileNotFoundError: [Errno 2] No such file or directory: '/home/jake/Development/mentat/tests/benchmarks/repos/for_transcripts/mentat/prompts/replacement_parser_prompt.txt'`. I think we need to filter out files that don't exist which happens when files are created as part of a PR. Sorry I probably should have done that in the original script.

mentat/code_context.py
```diff
@@ -74,7 +76,47 @@ def _get_all_features(
             )
             all_features.append(_feature)
 
-    return all_features
+    return sorted(all_features, key=lambda f: f.path.relative_to(git_root))
+
+
+def _merge_features(a: list[CodeFeature], b: list[CodeFeature]) -> list[CodeFeature]:
```
jakethekoenig, line 82: Could you add a test for this function?

mentat/code_context.py
```diff
@@ -74,7 +76,47 @@ def _get_all_features(
...
+        matching[0].intervals = [
+            Interval(
+                min(feature_start.start, matching_start.start),
+                max(feature_end.end, matching_end.end),
+            )
+        ]
+        if feature.diff and not matching[0].diff:
+            matching[0].diff = feature.diff
+        if feature.user_included and not matching[0].user_included:
+            matching[0].user_included = feature.user_included
```
jakethekoenig, line 117: I'm confused. Does this matching feature ever get added to output?

mentat/auto_context.py
```diff
@@ -0,0 +1,150 @@
...
+    """)
+training_prompt = (
+    "You are currently being used to generate benchmarking data, so you'll be shown"
+    " those Code Edits as well in order to maximize your performance."
+)
+
+
+async def select_features_llm(
```
jakethekoenig, line 73: Could we add test for `select_features_llm` and `select_features_greedy`

jakethekoenig:
Sorry for leaving you with a lot of merge conflicts. I can fix them for you tomorrow if you don't want to bother with them.

granawkins:
No worries, I have a handful of other changes too.

mentat/auto_context.py
```diff
@@ -0,0 +1,150 @@
...
+    remaining_tokens = max_tokens
+    for feature in features:
+        _levels = list(set(levels) | {feature.level})
+        _levels = sorted(list(_levels), key=lambda v: v.rank)
+        for level in _levels:
+            feature.level = level
+            if feature.count_tokens(model) <= remaining_tokens:
+                output.append(feature)
+                remaining_tokens -= feature.count_tokens(model)
+                break
```
biobootloader, line 41: just to confirm, the result of this we'll get a bunch of features at the highest level of detail, and then probably just a couple features at a lower level? And then nothing for remaining features?

mentat/auto_context.py
biobootloader: Another downside is latency as we'd make multiple calls, but we could do them in parallel

mentat/auto_context.py
```diff
@@ -0,0 +1,150 @@
...
+    system_prompt_tokens = count_tokens(system_prompt, feature_selection_model)
+    preselect_max_tokens = (
+        feature_selection_max_tokens
+        - system_prompt_tokens
+        - user_prompt_tokens
+        - feature_selection_response_buffer
+    )
+    preselected_features = select_features_greedy(
+        features, preselect_max_tokens, feature_selection_model, levels
+    )
```
biobootloader, line 98: this will return some features at codemap or other levels, right? will the llm be able to see line numbers to select those?

scripts/git_log_to_transcripts.py
```diff
@@ -140,19 +140,56 @@ async def translate_commits_to_transcripts(repo, count=10):
...
+            if not edited_files.issubset(selected_files):
+                print(
+                    "Auto-context missing required files:",
+                    edited_files - selected_files,
+                )
+                print("Using edited_files instead")
```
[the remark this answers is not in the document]
biobootloader: @jakethekoenig how are you thinking about this now? I also wrote up some of my thoughts here: https://abanteai.slack.com/archives/C05T11DMCF3/p1698879054329529

tests/benchmarks/context_benchmark.py
granawkins: For TP/FP/TN, ya it will be an integer. For Precision/Recall/F1, they should always be 0<i<1, like they are.

mentat/code_context.py
granawkins: It should - that was a mistake

mentat/auto_context.py
granawkins: Correct. Kindof a depth-first selection instead of our previous breadth-first.

mentat/auto_context.py
granawkins: Currently yes - the levels list here includes `cmap`, `cmap_full` and `file_name`. But actually it should be only `interval` (or `code` if ctags is unavailable).

mentat/auto_context.py
```diff
@@ -0,0 +1,150 @@
+import json
```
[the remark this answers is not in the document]
granawkins: A couple important points from implementing this:
1. The `LLMSelector` needs access to all features (including user-selected features) in order for the LLM to 'understand' the problem fully. This means every selector needs to take the full list, and it must include the user-selected features it received in its response. For the `GreedySelector`, this means putting user-selected features at the beginning of the list, as if they're the first search results.
2. Allowing for multiple-features-per-file *and* multiple-intervals-per-feature was getting really complex, so I made all features have only one interval. This required changing self.include_files from {path: feature} to {path: [feature]}.

mentat/code_context.py
granawkins: Even better, I axed it 🙌

scripts/git_log_to_transcripts.py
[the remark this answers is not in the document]
jakethekoenig: I think I'd like to have both versions of the benchmark:
1. A benchmark that compares our feature selector's selection to edited files. If our feature selector is good it should have a very high recall (anything short of 1 would actually be a problem) but not a very high precision. In fact a high precision would indicate a problem.
2. A benchmark that compares our feature selector's selection with our best guess for a necessary and sufficient subset of the codebase to see to make the change. The problem with this is we need to use our feature selector to generate the necessary and sufficient subset. But we can bootstrap by making some expensive/slow feature selectors for benchmark data generation not intended for use in mentat (though we should maybe make them available in the config and maybe people might want to use them with the PythonClient). If we genuinely have a necessary and sufficient subset then the higher the precision and recall the better.

What I don't like about this benchmark generation as is is:
1. Some benchmarks will be 1 and others will be our attempt at 2 which will make it hard to evaluate performance.
2. We're generating the benchmark with the same process we're generating contexts that will be compared against it. I guess the difference is the benchmark gets the full change and in the test it only gets the prompt.

tests/benchmarks/context_benchmark.py
jakethekoenig: Oh I see. Those are counts of tokens, not percentages? Makes sense.

biobootloader:
> I'd like the selectors to be aware of what files were manually included. Exactly how they respect that can be up to them. For instance maybe the LLM selector would always send it to the LLM to ask if it should be included but if the LLM says no it wouldn't be put in the final context even if the user included it.

It'd be good if the LLM saw those files as part of the selection process, as it gives it a hint on what else need to be included. And the user's prompt might refer to things in those files that would be ambiguous if you didn't know what files they included

scripts/git_log_to_transcripts.py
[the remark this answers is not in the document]
biobootloader: I agree it'd be nice to have both those versions of the benchmark

> Some benchmarks will be 1 and others will be our attempt at 2 which will make it hard to evaluate performance.

Why can't we make at an attempt at 2 on all benchmarks? And just run both ways of evaluation on all benchmarks?

> We're generating the benchmark with the same process we're generating contexts that will be compared against it. I guess the difference is the benchmark gets the full change and in the test it only gets the prompt.

Can you clarify this?

granawkins:
> > I'd like the selectors to be aware of what files were manually included. Exactly how they respect that can be up to them. For instance maybe the LLM selector would always send it to the LLM to ask if it should be included but if the LLM says no it wouldn't be put in the final context even if the user included it.
> 
> It'd be good if the LLM saw those files as part of the selection process, as it gives it a hint on what else need to be included. And the user's prompt might refer to things in those files that would be ambiguous if you didn't know what files they included

As it stands, the `CodeFeature` itself has a `user_included` boolean flag, which is set correctly for all features passed to the Selectors, so they can decide if/how to respond.

For the case of the GreedySelector, checking for and prioritizing those seemed contrary to the spirit of it, so I move user-included features to the top of the list (formerly sorted by search), and that achieves the same thing. 

For the LLMSelector, it doesn't know currently, but we can add some indication in the code_message like we've talked about elsewhere. e.g. add it to the filename line: `mentat/conversation.py USER-INCLUDED`. I'll leave that update for a future PR.

jakethekoenig:
Given there isn't much user facing change in this PR, just another context selector hidden behind a flag, I think I'm okay to merge this to make it easier to work on the benchmark and the more expensive context selector algorithm (which I could do tomorrow?) in parallel. What do you think @biobootloader ?

granawkins:
I got it working as intended on the git-transcripts-script and pushed the small changes required for that.

Currently the LLMSelector's outputs still aren't usable for the benchmarks because they are missing ~10% of the edited lines. However I think this has to do with the prompt, so that's what I'm planning to work on next.

So I agree with @jakethekoenig that we ought to merge this as-is, switched off.

biobootloader (approved):

PR #230, opened 2023-10-30
Fix: --diff/--pr-diff

jakethekoenig:
These flags weren't working because the SESSION_CONTEXT wasn't initialized when get_treeish_metadata was called.

### DIFF
--- a/mentat/git_handler.py
+++ b/mentat/git_handler.py
@@ -124,9 +124,10 @@ def get_diff_for_file(target: str, path: Path) -> str:
         raise UserError()
 
 
-def get_treeish_metadata(target: str) -> dict[str, str]:
-    session_context = SESSION_CONTEXT.get()
-    git_root = session_context.git_root
+def get_treeish_metadata(target: str, git_root: Path | None = None) -> dict[str, str]:
+    if git_root is None:
+        session_context = SESSION_CONTEXT.get()
+        git_root = session_context.git_root
 
     try:
         commit_info = subprocess.check_output(
--- a/mentat/diff_context.py
+++ b/mentat/diff_context.py
@@ -131,7 +131,7 @@ class DiffContext:
                 self.name = "HEAD (last commit)"
                 return
 
-        meta = get_treeish_metadata(target)
+        meta = get_treeish_metadata(target, git_root=git_root)
         name += f'{meta["hexsha"][:8]}: {meta["summary"]}'
 
         self.target = target


tests/diff_context_test.py
```diff
@@ -159,3 +161,26 @@ async def test_diff_context_pr(temp_testbed, git_history, mock_session_context):
...
+    session_context = SESSION_CONTEXT.get()
+    code_context = session_context.code_context
+    diff_context = code_context.diff_context
+
+    assert diff_context.target == "HEAD~2"
+    assert diff_context.name.startswith("HEAD~2: ")
+    assert diff_context.name.endswith(": add testbed")
+    assert diff_context.files == [abs_path]
+
+    await python_client.call_mentat("Conversation")
```
jakethekoenig, line 185: If you stop the python_client without calling it at least once you get a `Task is destroyed but it is pending` which is likely indicative of a bug with the python client.

jakethekoenig:
If you fail to call `mock_setup_api_key` on a test that needs it the test will pass locally (because locally you probably have an openai key). I wonder if we should add something to conftest that would make the test fail.

PCSwingle:
> If you fail to call `mock_setup_api_key` on a test that needs it the test will pass locally (because locally you probably have an openai key). I wonder if we should add something to conftest that would make the test fail.

Yeah, we should either have an auto fixture that resets the environment variable or have the mock_setup_api_key fixture be autoused (I'm leaning towards the second one because the first one would have to always run before mock_setup_api_key and I'm not sure the order auto-used fixtures run in)

PCSwingle (approved):
This is a duplicate of #227, but I like that you added a test, so let's merge this one

mentat/git_handler.py
```diff
@@ -124,9 +124,10 @@ def get_diff_for_file(target: str, path: Path) -> str:
         raise UserError()
 
 
-def get_treeish_metadata(target: str) -> dict[str, str]:
-    session_context = SESSION_CONTEXT.get()
-    git_root = session_context.git_root
+def get_treeish_metadata(target: str, git_root: Path | None = None) -> dict[str, str]:
```
PCSwingle, line 127: Let's just make this git_root: Path, and not have it ever be None

jakethekoenig:
> This is a duplicate of #227, but I like that you added a test, so let's merge this one

Oh nice. I've got to pay more attention.

PR #232, opened 2023-10-31
Git Parser Fix

jakethekoenig:
When a git diff shows exactly one line changed the number which says how many lines were changed is omitted. This is rarer than it sounds because git usually gives context around a change. But it happens for one line files like .nvmrcs.


PCSwingle (approved):
Good find!

PR #233, opened 2023-10-31
Git transcript processing tweaks

jakethekoenig:
* We ask GPT to respond in json and give a little more information about the commit.
  * The mocked transcript is stored in the same cache.
* A new script is made to make the jsonls which selects transcripts with specified properties from the commit_information.json
* `git show --first-parent` is used so we can get a git diff for pull requests (thanks gpt)
* The repo name is inferred from the url instead of being constant.
* We checkout main instead of head.

### DIFF
--- a/scripts/git_log_to_transcripts.py
+++ b/scripts/git_log_to_transcripts.py
@@ -21,23 +21,34 @@ from mentat.parsers.replacement_parser import ReplacementParser
 from mentat.session_context import SESSION_CONTEXT, SessionContext
 from tests.benchmarks.utils import clone_repo
 
+
 system_prompt = dedent("""\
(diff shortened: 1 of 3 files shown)


PCSwingle (commented):
Looks good to me; have you tested this new system prompt? Does it work well?

PCSwingle (approved):

jakethekoenig:
> Looks good to me; have you tested this new system prompt? Does it work well?

Yeah it works pretty well. It correctly judges things to be docs or configuration. The complexity score is pretty tightly concentrated in 3-4 but I think it does provide enough signal. It does sometimes fail to reply in valid json. But only like 20% of the time.

PR #236, opened 2023-10-31
Rubric benchmark

jakethekoenig:
[benchmark_results.json](https://github.com/AbanteAI/mentat/files/13221332/benchmark_results.json)
[benchmarks.json](https://github.com/AbanteAI/mentat/files/13221333/benchmarks.json)

Mentat is given a repo, commit and prompt and attempts to make edits. gpt-4 grades. The actual edits can also be graded. You can generate benchmarks with `./scripts/git_log_to_transcripts.py`. You can also use the attached `benchmarks.json`. If you copy it to `tests/benchmarks/repos/mentat` (where you should clone mentat). I'd like to keep this PR small but in the future we should have:
* An html results page like for exercism
* Comparative quality benchmarks where we put both the actual and generated diffs in context. Both a version where we tell gpt which is real and one where we don't.

My first impression looking at the grading is gpt-4 overrates it's correctness. For instance in the exercism clojure benchmark runner gpt doesn't override the right methods but they aren't in context so it has no way to know.

### DIFF
--- a/tests/benchmarks/edit_rubric_benchmark.py
+++ b/tests/benchmarks/edit_rubric_benchmark.py
@@ -1,0 +1,154 @@
+import json
+import os
+import subprocess
+from itertools import islice
+from pathlib import Path
+from textwrap import dedent
+
+import openai
+import pytest
+from git import Repo
+
+from mentat.python_client.client import PythonClient
+from tests.benchmarks.utils import clone_repo
+
+pytestmark = pytest.mark.benchmark
+
+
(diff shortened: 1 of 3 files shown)


biobootloader (approved):
Ok to merge this now, but I don't think it can be too useful until we include more in context. It'd need the task the diff was supposed to accomplish as well as the codebase context to understand better. After those two are added let's try the comparative diff as well like you mentioned

PR #238, opened 2023-11-02
Fix: cli flag parsing

jakethekoenig:
The boolean flags would still be set to true or false if no flag was added so settings set in the config file would be overwritten with no flag. Namespace values are now not stored if they are equal to the default.

Also I realized attrs provided a boolean converter so I removed mine.

### DIFF
--- a/mentat/config.py
+++ b/mentat/config.py
@@ -6,7 +6,7 @@ from json import JSONDecodeError
 from pathlib import Path
 
 import attr
-from attr import validators
+from attr import converters, validators
 
 from mentat.git_handler import get_shared_git_root_for_paths
 from mentat.parsers.parser_map import parser_map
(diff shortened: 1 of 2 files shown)


biobootloader:
> The boolean flags would still be set to true or false if no flag was added so settings set in the config file would be overwritten with no flag. Namespace values are now not stored if they are equal to the default.

I don't quite understand - when would the file be overwritten? And does this solution create another problem of not being able to overwrite to set something back to default values?

jakethekoenig:
> > The boolean flags would still be set to true or false if no flag was added so settings set in the config file would be overwritten with no flag. Namespace values are now not stored if they are equal to the default.
> 
> I don't quite understand - when would the file be overwritten? And does this solution create another problem of not being able to overwrite to set something back to default values?

Because the boolean flags have an action of `--store-true` or `--store-false` there's now way to use a cli flag to set them to their default. e.g. if you want to enable embeddings you use the flag `--use-embeddings` but there is no `--use-embeddings false` option. The problem this fixes is that the namespace always has a use-embeddings field. It's just set to false if the user didn't use the flag. This is processed after the .mentat-config.json so if the user sets `use-embeddings` there it won't have any effect.

biobootloader (approved):

Issue #239, opened 2023-11-02
--pr-diff isn't including the changed files


biobootloader:
The model doesn't see them:

<img width="1002" alt="image" src="https://github.com/AbanteAI/mentat/assets/128252497/f242ebda-7759-471c-96f5-ddabd28558ed">

biobootloader:
@granawkins want to take a look at this? I can look at too if you don't get a chance soon

Issue #237, opened 2023-11-02
Warn instead of crash on invalid tree-ish


jakethekoenig:
If you run `mentat --diff lol` you get the following exception:
```
Traceback (most recent call last):
  File "/home/jake/Development/mentat/.venv/bin/mentat", line 10, in <module>
    sys.exit(run_cli())
  File "/home/jake/Development/mentat/mentat/terminal/client.py", line 239, in run_cli
    terminal_client.run()
  File "/home/jake/Development/mentat/mentat/terminal/client.py", line 176, in run
    asyncio.run(self._run())
  File "/usr/lib/python3.10/asyncio/runners.py", line 44, in run
    return loop.run_until_complete(main)
  File "/usr/lib/python3.10/asyncio/base_events.py", line 649, in run_until_complete
    return future.result()
  File "/home/jake/Development/mentat/mentat/terminal/client.py", line 171, in _run
    await self._startup()
  File "/home/jake/Development/mentat/mentat/terminal/client.py", line 114, in _startup
    self.session = Session(
  File "/home/jake/Development/mentat/mentat/session.py", line 50, in __init__
    code_context = CodeContext(stream, git_root, diff, pr_diff)
  File "/home/jake/Development/mentat/mentat/code_context.py", line 98, in __init__
    self.diff_context = DiffContext(stream, git_root, self.diff, self.pr_diff)
  File "/home/jake/Development/mentat/mentat/diff_context.py", line 114, in __init__
    treeish_type = _get_treeish_type(git_root, target)
  File "/home/jake/Development/mentat/mentat/diff_context.py", line 203, in _get_treeish_type
    raise UserError(f"Invalid treeish: {treeish}")
mentat.errors.UserError: Invalid treeish: lol
```
I think it'd be better to catch the exception and display a red warning: `Unrecognized diff target: lol`

PR #241, opened 2023-11-03
Include paths from diff/pr-diff if no paths given

granawkins:
Addresses #239: Use --diff or --pr-diff to set paths if no paths arg given. We used to do this but it must've gotten lost at some point.

Another issue: If the diff/pr-diff is too big for context, it raises a MentatError but doesn't close Mentat. Seems to be a lifecycle issue - the shutdown listeners maybe not yet in place. I'd defer to @waydegg on the best way to address this. Might be related to #237 too - handling errors in the session startup phase.

### DIFF
--- a/mentat/code_context.py
+++ b/mentat/code_context.py
@@ -107,6 +107,8 @@ class CodeContext:
         exclude_paths: list[Path],
         ignore_paths: list[Path] = [],
     ):
+        if not paths and (self.diff or self.pr_diff) and self.diff_context.files:
+            paths = self.diff_context.files
         self.include_files, invalid_paths = get_include_files(paths, exclude_paths)
         for invalid_path in invalid_paths:
             print_invalid_path(invalid_path)
--- a/mentat/session.py
+++ b/mentat/session.py
@@ -67,6 +67,8 @@ class Session:
 
         # Functions that require session_context
         config.send_errors_to_stream()
+        if not paths and (diff or pr_diff) and code_context.diff_context.files:
+            paths = code_context.diff_context.files
         code_context.set_paths(paths, exclude_paths, ignore_paths)
         code_context.set_code_map()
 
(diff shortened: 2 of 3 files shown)


jakethekoenig:
I wanted to add a test and it was easier if the logic was in `code_context.set_paths` then the session constructor. You can verify this test fails before this PR (and passed without the assert, it didn't require any paths to be passed to the python client).

I'm okay with this PR but will wait for someone else to LGTM since I made the last change.

I'm not entirely sure the behavior is what users will expect. I personally would expect all the changed files to always be added to the context if --diff was included. This is a more complicated change but maybe it would be best to include intervals around the actual changes so much larger changes would have a hope of fitting in context?

jakethekoenig:
[conversation_viewer.json](https://github.com/AbanteAI/mentat/files/13249946/conversation_viewer.json)
Oh one more thing. I was sort of unhappy with this conversation with mentat (rename to .html, github doesn't like you uploading htmls). I don't like how it said `The added lines are marked with a "+" symbol at the beginning.`. Not sure how common such a sentence is in diff explanations. Maybe a fluke or we should do some prompt engineering to suppress it.

biobootloader (approved):
Thanks for the fix @granawkins and the test @jakethekoenig 

> I'm not entirely sure the behavior is what users will expect. I personally would expect all the changed files to always be added to the context if --diff was included. This is a more complicated change but maybe it would be best to include intervals around the actual changes so much larger changes would have a hope of fitting in context?

Yeah that seems like it'd be the more expected behavior. I'll make an issue to discuss and potentially change this, it'll be easier once we have function-level features instead of whole files, so even diffs touching a lot of files can be added.

> Oh one more thing. I was sort of unhappy with this conversation with mentat (rename to .html, github doesn't like you uploading htmls). I don't like how it said The added lines are marked with a "+" symbol at the beginning.. Not sure how common such a sentence is in diff explanations. Maybe a fluke or we should do some prompt engineering to suppress it.

I'll also make an issue for this

PR #246, opened 2023-11-04
Remove `git_root` from Code Context

waydegg:
# To-do

- [ ] Update existing test cases
- [ ] Add test cases for non-git testbed (need to enforce everything works w/o git)
- [ ] Add test cases for glob pattern matching
- [ ] Re-add `DiffContext` into `CodeContext` 
- [ ] Remove random comments


mentat/utils/__init__.py
waydegg, line 1: Probably should separate these utils out into their own domain-specific modules (e.g. "hashing", "template")

pyrightconfig.json
```diff
@@ -1,5 +1,5 @@
 {
-    "include": ["mentat"],
-    "ignore": ["testbed", "tests", "scripts"],
-    "typeCheckingMode": "strict",
+  "include": ["mentat", "tests"],
```
waydegg, line 2: Added this for development which has helped catch a lot of edge-cases and bugs when updating the tests. I think a balance would be to just add non-strict typing to tests

mentat/utils/path.py
```diff
@@ -0,0 +1,153 @@
...
+    else:
+        raise PathValidationException(f"Unable to validate path {path}")
+
+    return abs_path
+
+
+def match_path_with_patterns(path: Path, patterns: Set[str]) -> bool:
+    """Check if the given absolute path matches any of the patterns.
+
+    TODO: enforce valid glob patterns? Right now we allow glob-like
```
waydegg, line 66: @biobootloader do you have any thoughts on this?

mentat/utils/path.py
```diff
@@ -0,0 +1,153 @@
...
+    patterns (the one's that .gitignore uses). This feels weird imo.
```
waydegg, line 67: @biobootloader do you have any thoughts on this?

mentat/session_context.py
```diff
@@ -22,7 +22,7 @@
 class SessionContext:
     stream: SessionStream = attr.field()
     cost_tracker: CostTracker = attr.field()
-    git_root: Path = attr.field()
+    cwd: Path = attr.field()
```
waydegg, line 25: This change forces us to update every file basically :/

mentat/session_context.py
```diff
@@ -22,7 +22,7 @@
 class SessionContext:
     stream: SessionStream = attr.field()
     cost_tracker: CostTracker = attr.field()
-    git_root: Path = attr.field()
```
waydegg, line 25: `git_root` is now assessed at a per-file and per-cwd basis

mentat/git_handler.py
waydegg, line 1: I updated all subprocess calls to send their stderr to DEVNULL

mentat/code_context.py
```diff
@@ -63,61 +55,39 @@ def _get_all_features(
...
-        self.ignore_files = get_ignore_files(ignore_paths)
+        self.include_files: Dict[Path, CodeFeature] = dict()
+        self.features: List[CodeFeature] = list()
+        self.code_map = True
 
     def set_code_map(self):
-        session_context = SESSION_CONTEXT.get()
-        config = session_context.config
-        stream = session_context.stream
+        ctx = SESSION_CONTEXT.get()
```
jakethekoenig, line 88: Thanks for changing this. It's a lot cleaner. Though it makes it even harder to tell what depends on what. And I'd rather it wasn't done as part of this PR. It makes the total number of lines changed massive. And it makes it hard to notice the little bits of logic that may have changed anywhere `git_root` was used before.

mentat/code_context.py
```diff
@@ -127,84 +97,70 @@ def set_code_map(self):
...
+            ctx.stream.send("Auto-Selected Features:", color="blue")
             for level, count in auto_features.items():
                 if count:
-                    stream.send(f"  {count} {level.description}")
+                    ctx.stream.send(f"  {count} {level.description}")
 
     _code_message: str | None = None
     _code_message_checksum: str | None = None
 
     def _get_code_message_checksum(self, max_tokens: Optional[int] = None) -> str:
```
jakethekoenig, line 149: I just noticed this doesn't depend on the user prompt. Now that we're using embeddings and LLMs it should. @granawkins

mentat/code_context.py
```diff
@@ -232,59 +185,60 @@ async def _get_code_message(
...
                 "Diff References:",
                 f' "-" = {self.diff_context.name}',
                 ' "+" = Active Changes',
                 "",
             ]
         code_message += ["Code Files:\n"]
-
         features = self._get_include_features()
-        meta_tokens = count_tokens("\n".join(code_message), model)
+        meta_tokens = count_tokens("\n".join(code_message), model)  # NOTE: why does this take so long to run?
```
jakethekoenig, line 204: Are you sure this is the bottle neck? I don't notice much latency when I put print statements around it.

mentat/code_feature.py
```diff
@@ -99,42 +98,39 @@ class CodeFeature:
...
         user_included: bool = False,
         name: Optional[str] = None,
     ):
-        if Path(path).exists():
-            self.path = Path(path)
-            self.intervals = [Interval(0, math.inf)]
+        ctx = SESSION_CONTEXT.get()
+
+        validated_path = validate_and_format_path(path=Path(path), cwd=ctx.cwd, check_for_text=False)
+        if ":" in str(validated_path):
```
jakethekoenig, line 118: This isn't safe. Paths are allowed to have colons in them. And in the case of windows they almost always do.

mentat/code_file_manager.py
```diff
@@ -46,89 +45,72 @@ def _create_file(self, code_context: CodeContext, abs_path: Path):
...
         for file_edit in file_edits:
-            rel_path = Path(os.path.relpath(file_edit.file_path, git_root))
+            rel_path = Path(os.path.relpath(file_edit.file_path, ctx.cwd))
             if file_edit.is_creation:
                 if file_edit.file_path.exists():
-                    raise MentatError(
-                        f"Model attempted to create file {file_edit.file_path} which"
-                        " already exists"
-                    )
+                    raise MentatError(f"Model attempted to create file {file_edit.file_path} which" " already exists")
```
jakethekoenig, line 73: There are two strings here that can be combined.

mentat/code_map.py
```diff
@@ -26,9 +25,7 @@ def get_ctags(
     else:
         ctags_cmd_args.append("--fields=+S")
     ctags_cmd = ["ctags", *ctags_cmd_args, str(abs_file_path)]
-    output = subprocess.check_output(
-        ctags_cmd, stderr=subprocess.DEVNULL, start_new_session=True, text=True
-    ).strip()
+    output = subprocess.check_output(ctags_cmd, stderr=subprocess.DEVNULL, start_new_session=True, text=True).strip()
```
jakethekoenig, line 28: There's no actual change here or in the previous line right? Did black formatting change?

mentat/commands.py
```diff
@@ -71,8 +66,7 @@ async def apply(self, *args: str) -> None:
         stream = session_context.stream
 
         stream.send(
-            f"{self.invalid_name} is not a valid command. Use /help to see a list of"
-            " all valid commands",
+            f"{self.invalid_name} is not a valid command. Use /help to see a list of" " all valid commands",
```
jakethekoenig, line 69: Combine strings

mentat/utils/path.py
```diff
@@ -0,0 +1,160 @@
...
+
+    # Validate different path types
+    # File
+    if abs_path.is_file():
+        if not abs_path.exists():
+            raise PathValidationException(f"File {abs_path} does not exist")
+        if check_for_text and not is_file_text_encoded(abs_path):
+            raise PathValidationException(f"Unable to read file {abs_path}")
+    # File interval
+    elif ":" in str(abs_path):
```
PCSwingle, line 49: Similar to @jakethekoenig's other comment we can't just check for a colon; we need to put this after the check for dirs and glob patterns (and also fix a few bugs in this elif statement; make sure we split the last colon -- in case the filename has a colon -- and that we actually check that the interval is valid, since otherwise we might split on a path like file/with:colon/filename and check for file/with which might actually exist). I feel like we should make a class function in Interval that does this all automatically.

mentat/utils/path.py
```diff
@@ -0,0 +1,160 @@
...
+        _interval_path, interval_str = str(abs_path).split(":", 1)
+        interval_path = Path(_interval_path)
+        if not interval_path.exists():
+            raise PathValidationException(f"File interval {abs_path} does not exist")
+        if check_for_text and not is_file_text_encoded(interval_path):
+            raise PathValidationException(f"Unable to read file interval {abs_path}")
+        abs_path = Path(f"{interval_path}:{interval_str}")
+    # Directory
+    elif abs_path.is_dir():
+        if not abs_path.exists():
```
PCSwingle, line 59: This will never happen; is_dir will only return true if it exists

mentat/code_context.py
```diff
@@ -366,57 +322,117 @@ def _feature_relative_path(f: CodeFeature) -> str:
...
+        try:
+            # file
+            if validated_path.is_file():
+                if validated_path not in self.include_files:
+                    ctx.stream.send(f"Path {validated_path} not in context", color="light_red")
+                    return excluded_paths
+                excluded_paths.add(validated_path)
+                del self.include_files[validated_path]
+            # file interval
+            elif ":" in str(validated_path):
```
PCSwingle, line 381: Same as other comments on intervals (again, I think Interval should have a helper function for this)

mentat/code_context.py
```diff
@@ -366,57 +322,117 @@ def _feature_relative_path(f: CodeFeature) -> str:
...
+                _interval_path, _ = str(validated_path).split(":", 1)
+                interval_path = Path(_interval_path)
+                if interval_path not in self.include_files:
+                    ctx.stream.send(f"Path interval {validated_path} not in context", color="light_red")
+                    return excluded_paths
+                excluded_paths.add(interval_path)
+                del self.include_files[interval_path]
+            # directory
+            elif validated_path.is_dir():
+                if validated_path not in self.include_files:
```
PCSwingle, line 391: I don't think directories are ever in self.include_files; the point of self.include_files is that it has all of the individual files; if a user puts a directory in, they want all the children of the directory that are included to be excluded

mentat/code_context.py
```diff
@@ -366,57 +322,117 @@ def _feature_relative_path(f: CodeFeature) -> str:
...
+                    ctx.stream.send(f"Directory path {validated_path} not in context", color="light_red")
+                    return excluded_paths
+                excluded_paths.add(validated_path)
+                del self.include_files[validated_path]
+            # glob
+            else:
+                for included_path in self.include_files.keys():
+                    if match_path_with_patterns(included_path, set(str(validated_path))):
```
PCSwingle, line 399: Why don't we just use glob.glob to get all matching files and then get the intersection instead of making our own helper glob function?

mentat/code_context.py
```diff
@@ -366,57 +322,117 @@ def _feature_relative_path(f: CodeFeature) -> str:
...
+                        excluded_paths.add(included_path)
+                        del self.include_files[included_path]
+
+        except PathValidationException as e:
```
PCSwingle, line 403: I don't think anything in this try except raises this; did you mean to wrap validate_and_format_path instead?

mentat/utils/path.py
PCSwingle: I don't think we need this function; can't we stick with glob.glob like we were before?

mentat/code_context.py
waydegg: Yea that makes sense, I'll revert this

mentat/code_context.py
waydegg: Anecdotal but I this hung for like 5-6 seconds when I was testing things a while ago. Out of scope for this PR anyways but will look into later

mentat/code_map.py
waydegg: Yeah I think something is wrong with my formatter...

mentat/code_feature.py
waydegg: Will update this

mentat/code_context.py
waydegg: Agree also

mentat/code_context.py
waydegg: Yup that's correct. Will fix this

mentat/code_context.py
granawkins: I'm digging into code_context performance now. I'm seeing this line takes 0.15 seconds the first time, but if I call it immediately after it only takes .00004 seconds, so I guess it has to do with loading `tiktoken` the first time.

waydegg:
Breaking this up in a series of stacked PRs!

PR #248, opened 2023-11-06
Feature Selection Refactor

jakethekoenig:
Logic for feature selection is moved to feature_filters. I felt it was hard to understand what was going on in get_code_message because it was responsible for three things:
* Collecting all the features that could end up in the LLMs context.
* Filtering down to a final list that does.
* Transforming it to a string. Though quite complicated default behavior is put into the DefaultFilter and written as a composition of simpler steps gated behind flags.

While I was doing that I cleaned up a few other things:
* _get_all_features made more sense as an instance method as all of its arguments were field variables anyway.
* The code_map field is removed and there's just the method code_map_enabled. I cached the code_map check. The function sets config.no_code_map if it fails so it won't log multiple time unless the user sets config.no_code_map themselves after it failed once.
* I renamed the selectors to filters. They aren't really filters either as they can reorder and drop CodeMessageLevels.
* I renamed the GreedySelector to the TruncationFilter. It's being greedy in the sense that it's taking whatever it sees first. But to me greedy would imply it's picking on some naive criterion. It's sort of doing that but Truncation seems more evocative to me and less likely to mislead someone into thinking we have some theoretical justification for why it should be good.
* I removed the comment comparing the context generation problem to the knapsack problem: I don't feel they are very similar as there are real costs to filling up the context and putting in sufficiently unrelated things seems more likely to harm performance than help.
* The non-streamed llm calling method is put in the llm api. I think if we actually want to use the llm selector as the default we should find a way to stream the intermediate results to not keep the user waiting.


mentat/code_context.py
```diff
@@ -262,62 +206,26 @@ async def _get_code_message(
...
-        )
-        remaining_tokens = max(0, max_tokens - meta_tokens - included_feature_tokens)
         auto_tokens = config.auto_tokens
-        if remaining_tokens == 0 or config.auto_tokens == 0:
-            self.features = sorted(
-                included_features, key=lambda f: f.path.relative_to(git_root)
-            )
+        remaining_tokens = max(0, min(auto_tokens, max_tokens - meta_tokens))
+
+        if remaining_tokens == 0:
```
jakethekoenig, line 212: I was disappointed this check was necessary. Logically the function should be the same if you always call `_get_all_features` and then you don't even need to compute `remaining_tokens`. But it's unfortunately pretty slow.

mentat/code_context.py
```diff
@@ -345,6 +253,54 @@ def _feature_relative_path(f: CodeFeature) -> str:
...
+                        user_features=self.include_files.get(abs_path, []),
+                    )
+                    all_features += _split_features
+            else:
+                _feature = CodeFeature(
+                    abs_path, level=level, diff=diff_target, user_included=user_included
+                )
+                all_features.append(_feature)
+
+        return sorted(all_features, key=lambda f: f.path.relative_to(git_root))
```
jakethekoenig, line 302: Is `relative_to(git_root)` necessary? Shouldn't they all have the same prefix anyway so it won't affect the sort?

mentat/feature_filters/truncate_filter.py
```diff
@@ -0,0 +1,45 @@
...
+        self.model = model
+        self.code_map = code_map
+        self.respect_user_include = respect_user_include
+
+    async def filter(
+        self,
+        features: list[CodeFeature],
+    ) -> list[CodeFeature]:
+        """Truncate the features to max_token tokens."""
+        levels = [CodeMessageLevel.FILE_NAME]
```
granawkins, line 23: I would keep `levels` as a kwarg and move it to __init__. There are times, e.g. at the end of LLMFeatureFilter, where we'd want to truncate without modifying the levels.

granawkins (approved):
I like the FeatureFilter interface, nice and clean. Couple comments otherwise looks great.

mentat/code_context.py
jakethekoenig: Oh sorry I didn't mean to change the behavior. I feel like a test should have caught that. I think the behavior is better this way because I don't think users should think about their `include_token` budget and `auto_token` budget. They should just think about their token budget and it should be up to us how to best use it.

mentat/code_context.py
```diff
@@ -262,62 +206,26 @@ async def _get_code_message(
...
-            feature_selector = get_feature_selector(self.use_llm)
-            self.features = await feature_selector.select(
-                candidate_features,
-                selectable_tokens,
-                model,
-                alt_levels,
-                prompt,
-                expected_edits,
-            )
+        feature_filter = DefaultFilter(
```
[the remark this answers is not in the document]
jakethekoenig: I put flags in `DefaultFilter` and `EmbeddingSimilarityFIlter` (Running `/search ""` was also throwing an error). I moved this filter into the if statement for now. The reason I didn't originally is because I think we should move to a place where `mentat .` works and to do that we'll need filtering on the included features as well.

PR #250, opened 2023-11-06
Display exception

jakethekoenig:
Currently if mentat encounters an uncaught exception it hangs and displays nothing. This change makes it print the stack trace and exit.


biobootloader (approved):
Nice! I added a test, someone else merge if that looks good.

PR #251, opened 2023-11-06
Conversation command shows old transcripts

jakethekoenig:
Clicking right and left goes through the history of logs.


jakethekoenig:
One thing: I was going to put this in a separate command or behind arguments to not slow down the current use case but it ended up being fast enough I thought it was better to keep it simple. I think I'll make it accept an optional argument that can be a log a json or jsonl and have it display those as well in a future pr

granawkins:
LGTM

granawkins (approved):

PR #253, opened 2023-11-07
add json parser

PCSwingle:
Must be used with either gpt-3.5-turbo-1106 or gpt-4-1106-preview (for the json response_format arg). So far from my testing (with gpt-4 turbo) it seems significantly worse than the other formats. I have yet to run any benchmarks on it.


granawkins (approved):
Looks good - I got it working by setting

config.parser = "json"
config.model = "gpt-4-1106-preview"

in Session.__init__()

PR #254, opened 2023-11-08
Speed up embeddings/search

granawkins:
Two changes:
- Use sqlite database instead of json/gzip. We save ~10s by not writing the gzip file, and lose ~0.3 seconds reading from the database instead of just from memory.
- Make up to n=10 (arbitrarily) simultaneous calls to the embeddings API, handle minute-level RateLimits

A future optimization could be to use [Annoy](https://github.com/spotify/annoy). This would replace that 0.3s call to the database and ~0.1s calculating cosine similarity. 

(all these numbers are from running mentat on the mentat repo)

### DIFF
--- a/mentat/embeddings.py
+++ b/mentat/embeddings.py
@@ -1,9 +1,8 @@
-import gzip
 import json
-import logging
 import os
 from pathlib import Path
 from timeit import default_timer
+import sqlite3
 
 import numpy as np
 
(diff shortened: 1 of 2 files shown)


mentat/embeddings.py
```diff
@@ -21,36 +23,52 @@
...
+                "CREATE TABLE IF NOT EXISTS embeddings "
+                "(checksum TEXT PRIMARY KEY, vector BLOB)"
+            )
 
-    def save(self):
-        with gzip.open(self.path, "wt") as f:
-            json.dump(self._dict, f)
+    def set(self, items: dict[str, list[float]]):
+        with self.conn:
+            self.conn.executemany(
```
jakethekoenig, line 46: I think here and line 38 you should write `with self.conn as db` and then `db.executemany` instead of repeating `self.conn`

jakethekoenig (approved):
This is awesome, I really feel the speed up.

PR #255, opened 2023-11-08
Auto context interface

granawkins:
This PR consolidates a lot of args and settings related to context.

Removed:
- use_embeddings (always on)
- no_code_map (always try to use)
- auto_tokens (config.maximum_context achieves the same thing)

Added:
- auto_context: bool = False (use cmaps, embeddings and llm to generate the best possible context)

I leave `CodeContext.use_llm` and set it to True by default. This is currently used by tests so that we can test filter behavior without calling the LLM.

### DIFF
--- a/mentat/code_context.py
+++ b/mentat/code_context.py
@@ -51,6 +51,7 @@ class CodeContext:
         # or the CodeFeatures (value) and their intervals. Redundant.
         self.include_files = {}
         self.ignore_files = set()
+        self.ctags_disabled = check_ctags_disabled()
 
     def set_paths(
         self,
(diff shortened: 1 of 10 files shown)


mentat/code_context.py
```diff
@@ -205,11 +190,9 @@ async def _get_code_message(
             ]
         code_message += ["Code Files:\n"]
         meta_tokens = count_tokens("\n".join(code_message), model)
+        remaining_tokens = max(0, max_tokens - meta_tokens)
 
-        auto_tokens = config.auto_tokens
-        remaining_tokens = max(0, min(auto_tokens, max_tokens - meta_tokens))
-
-        if remaining_tokens == 0:
+        if not config.auto_context or remaining_tokens <= 0:
```
jakethekoenig, line 195: I like the `<=` comparison. It means we can remove the max from the `remaining_tokens` computation as well.

jakethekoenig (approved):
I like the simplified interface. Thanks! At first I was worried there was no way to turn off embeddings so users of local models without local embedding models wouldn't be able to use mentat. But I guess they can just turn the auto context off entirely.

PR #257, opened 2023-11-09
add sentry and version checking message

PCSwingle:
Adds sentry and a version checking message.

TODO before merge:
Don't track tests
Only track if installed from pip? (Will prevent dev exceptions from sending and any exceptions from modified source)

### DIFF
--- a/mentat/__init__.py
+++ b/mentat/__init__.py
@@ -1,0 +1,2 @@
+# Make sure to bump this on Release x.y.z PR's!
+__version__ = "0.1.21"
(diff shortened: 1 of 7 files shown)


granawkins (approved):
Looks fine, save for the emails/environments issues you mentioned.

After this initial commit, I guess I'll think about other places it'd be useful to have these.

I understand the issue with the built-in decorator (can't set description/tags).. having a decorator would make it a lot easier to use. Maybe we could build our own?

PR #258, opened 2023-11-09
Aider attribution

jakethekoenig:
We used their prompts in our exercism benchmark and it seems polite to attribute it.


tests/benchmarks/exercism_practice.py
```diff
@@ -112,6 +112,8 @@ async def run_exercise(problem_dir, language="python", max_iterations=2):
     )
     await client.startup()
 
+    # These prompts are copied from Aider: https://aider.chat/docs/benchmarks.html to
+    # allow direct comparison.
```
biobootloader, line 116:
```suggestion
    # allow for direct comparison.
```

biobootloader (approved):

PR #263, opened 2023-11-09
fix error handling

PCSwingle:
Added some better error handling so that errors that we throw automatically quit (sending everything to the stream, instead of printing it to the console where the user might never see it if they aren't using the CLI)


granawkins (approved):
LGTM 🚀

.github/workflows/lint_and_test.yml
```diff
@@ -55,7 +55,9 @@ jobs:
         run: |
           pytest
       - name: Can run Mentat
-        # Ensure that python doesn't import local mentat folder and that 'mentat' command calls mentat instead of switching folder
+        # Unfortunately Github Actions Runners have trouble with prompt toolkit, so we can't do this on Windows.
```
waydegg, line 58: What is the issue exactly for windows?

mentat/broadcast.py
```diff
@@ -99,9 +99,7 @@ def connect(self) -> None:
         self._listener_task = asyncio.create_task(self._listener())
 
     def disconnect(self) -> None:
-        if self._listener_task.done():
-            self._listener_task.result()
-        else:
```
waydegg, line 104: Why this change? I'm pretty sure we want to keep this since `.result()` will [propagate Exceptions](https://docs.python.org/3.10/library/asyncio-task.html#asyncio.Task.result) correctly.

mentat/errors.py
```diff
@@ -3,20 +3,23 @@ class ModelError(Exception):
...
 
-# Used to indicate a KeyboardInterupt thrown by a remote client
-class RemoteKeyboardInterrupt(Exception):
+class UserError(Exception):
     pass
 
 
+# This will exit the session without showing any sign of error
 class SessionExit(Exception):
     pass
```
waydegg, line 25: Instead of comments we can make these docstrings (will show up in VSCode/LSP intellisense descriptions).

E.g.

```
class RemoteKeyboardInterrupt(Exception):
   """Used to indicate a KeyboardInterrupt thrown by a remote client"""
```

mentat/session.py
```diff
@@ -116,24 +119,37 @@ def start(self) -> asyncio.Task[None]:
...
+                await self.stop()
 
         setup_logging()
         self._main_task: Task[None] = asyncio.create_task(run_main())
         return self._main_task
 
     async def stop(self):
+        if self.stopped:
+            return
+        self.stopped = True
```
waydegg, line 141: We're setting the session as "stopped" here, but there's still stopping-logic that's running afterwards that can raise Exceptions (and therefore not complete the whole stop-sequence).

I think `self.stopped = True` should be set at the end of the `stop` method, and there should be some variable/lock that prevents `stop` from being called concurrently.

mentat/session.py
```diff
@@ -116,24 +119,37 @@ def start(self) -> asyncio.Task[None]:
...
         session_context = SESSION_CONTEXT.get()
         cost_tracker = session_context.cost_tracker
 
         cost_tracker.display_total_cost()
         logging.shutdown()
         self._main_task.cancel()
-        await self._main_task
+        try:
+            await self._main_task
+        except CancelledError:
```
waydegg, line 151: When does `await self._main_task` raise a `CancelledError`?

waydegg (changes requested):
Left some comments on a few things that may create edge-cases during the Session's "stop" sequence.

.github/workflows/lint_and_test.yml
PCSwingle: It throws some NoConsoleScreenBufferError exception; it's weird because prompt toolkit works on my windows, so it really is just GHA Windows runners. It throws the error on the creation of a prompt toolkit session.

mentat/broadcast.py
PCSwingle: Oh hmm I didn't think about exceptions. I changed it because it throws a CancelledError (since it's pretty much always cancelled at this point); I think I'll just wrap it in a try except for CancelledError and keep it like it was. Thanks!

mentat/errors.py
PCSwingle: Good catch!

mentat/session.py
PCSwingle: We cancel it on the line before this, so I'm pretty sure that's why it raises a CancelledError.

mentat/session.py
PCSwingle: What Exceptions can be raised? I'm not entirely sure what better way there is to do this; if we raise an Exception during shutdown, there's really no way to recover (since running stop again is just going to raise the Exception again most likely). I only added in the stopped variable to prevent this function from being called twice, not to indicate that the session is stopped. The reason the stop function is called twice is because we stop the Client, which calls to stop the Session again; I think I might fix this, since I think the Client should instead signal to the server that it needs to stop, and then the server will be in charge of stopping itself and telling the client to stop with the exit signal.

PCSwingle:
@waydegg I addressed your issues and also switched to a Session/Server first lifecycle; now, the server always decides when to shutdown. The client can expedite this by sending a session_exit signal, and the server (when finished) will send a client_exit signal telling the client to shutdown. Let me know if you have any additional thoughts!

.github/workflows/lint_and_test.yml
waydegg: Interesting. I'll make a GH issue for this

mentat/session.py
waydegg: Ah that would make sense haha

mentat/session.py
waydegg: > if we raise an Exception during shutdown, there's really no way to recover (since running stop again is just going to raise the Exception again most likely)

Yeah that makes sense. If an Exception is thrown during shutdown it should be up to the Server (which is managing the Session) on how to handle it. Now that I'm looking at it again, there's not really any way an Exception can be thrown anyways which is good.

How do we check if a Session is stopped though? There's a race condition of `self.stopped` being `True` while `self._exit_task` and `self._main_task` are cancelling (i.e. the Session is still stopping)? With how the Terminal Client handles Session shutdown (in `self._listen_for_client_exit`) this isn't a problem, however I think it makes more sense if we have a `self.stopping` (which replaces `self.stopped` here) and a `self.stopped` which is set at the end of the `self._stop` method (which can be set in the `finally` section of a try-catch block). Thoughts?

waydegg (approved):
LGTM! Left 1 comment on the Session stop sequence (not blocking tho)

PR #264, opened 2023-11-10
Loading handler

granawkins:
Adds a tqdm loader when generating auto-context. 

I setup a new stream channel, "loading", and our terminal client listens to it builds/updates the tqdm bar.


biobootloader:
Probably this is a pre-existing problem, but I don't think the reported auto-selected features and token counts are accurate: it says it only added 1 specific range and that the total token count is 1367, but that seems wrong. The conversation viewer looks pretty full (potentially full 128k limit?)

image

biobootloader:
The tqdm loader looks awesome though!

waydegg:
Was running some tests manually and found a couple output bugs w/ tqdm and prompt toolkit:

- The LLM response is printed before the progress bar is complete. I can see this being intentional tho, especially if we want to show the user that other stuff is happening/processing in the background as the LLM response is being streamed in.
- After the LLM response is printed, the progress bar is printed again but is never updated visually

Screenshot 2023-11-10 at 9 25 16 AM

jakethekoenig:
> Probably this is a pre-existing problem, but I don't think the reported auto-selected features and token counts are accurate: it says it only added 1 specific range and that the total token count is 1367, but that seems wrong. The conversation viewer looks pretty full (potentially full 128k limit?)
> 
> image

I have a fix. I re-added an auto-token setting with a limit of 3000. I also turned off llm selection by default as this can cause exceptions when users set model but not `llm_selection_model`.

biobootloader:
> I have a fix.

@jakethekoenig does this fix the issue of accuracy for token counts? in my screenshot it showed "total token count" of 1367, not much more than the prompt of 1263, so I think most/all of what it added wasn't counted there

jakethekoenig:
> > I have a fix.
> 
> @jakethekoenig does this fix the issue of accuracy for token counts? in my screenshot it showed "total token count" of 1367, not much more than the prompt of 1263, so I think most/all of what it added wasn't counted there

Huh, I don't actually see that behavior. I ran mentat with only the `README.md` included and it said

mentat/feature_filters/llm_feature_filter.py
```diff
@@ -86,11 +89,15 @@ async def filter(
...
+
+        self.report_loading("Asking LLM to filter out irrelevant context...", 80)
         message = await call_llm_api_sync(model, messages)
+
+        self.report_loading("Parsing LLM response...", 10)
         try:
             selected_refs = json.loads(message)  # type: ignore
         except json.JSONDecodeError:
-            raise ValueError(f"The response is not valid json: {message}")
+            raise ModelError(f"The response is not valid json: {message}")
```
jakethekoenig, line 100: In my testing I hit this exception a lot.

jakethekoenig:
> > > I have a fix.
> > 
> > 
> > @jakethekoenig does this fix the issue of accuracy for token counts? in my screenshot it showed "total token count" of 1367, not much more than the prompt of 1263, so I think most/all of what it added wasn't counted there
> 
> Huh, I don't actually see that behavior. I ran mentat with only the `README.md` included and it said 2975 tokens in context. When I set `auto_context` to `True` and re-ran I saw 4238 tokens in context (Sorry for close/re-open. Misclick). Both looked accurate from `/conversation`.

I ran from this branch with the prompt "Can you summarize the code base for me?" and it said it included 93801 tokens. Which seems about right. It responded with "The code base cannot be summarized. Please provide a specific request for code changes or clarifications". Which is less than ideal.

biobootloader:
> Huh, I don't actually see that behavior. I ran mentat with only the `README.md` included and it said 2975 tokens in context. When I set `auto_context` to `True` and re-ran I saw 4238 tokens in context (Sorry for close/re-open. Misclick). Both looked accurate from `/conversation`.

ah you're right, no issue with token counts!

biobootloader:
> I ran from this branch with the prompt "Can you summarize the code base for me?" and it said it included 93801 tokens. Which seems about right. It responded with "The code base cannot be summarized. Please provide a specific request for code changes or clarifications". Which is less than ideal.

haha yeah not ideal... I have asked similar things though and gotten summaries (not necessarily good ones, I didn't look too closely but it does list major things).

mentat/feature_filters/llm_feature_filter.py
granawkins: Per your other PR and the conversation there, I'm going to switch off `use_llm` which will bypass this code, and we'll rehaul it in a future PR.

granawkins:
> Was running some tests manually and found a couple output bugs w/ tqdm and prompt toolkit:
> 
> * The LLM response is printed before the progress bar is complete. I can see this being intentional tho, especially if we want to show the user that other stuff is happening/processing in the background as the LLM response is being streamed in.
> * After the LLM response is printed, the progress bar is printed again but is never updated visually

I got rid of the `input_request` listener, and instead put all the loading_handler stuff in a try block, with finally: loading_handler.terminate(). This should fix it. There were some other bugs going on which are also fixed.

mentat/code_file_manager.py
```diff
@@ -81,27 +81,10 @@ async def write_changes_to_files(
...
-                for r in file_edit.replacements:
-                    for i in range(r.starting_line, r.ending_line):
-                        if not any(f.contains_line(i) for f in context_features):
-                            missing_lines = True
-                if not context_features or missing_lines:
-                    stream.send(
-                        f"Attempted to edit file {file_edit.file_path} not in context",
-                        color="yellow",
-                    )
-                    continue
```
granawkins, line 104: It often happens with auto-context that only some interval is included in context, and we get edits on that interval. I previously updated this to handle intervals, but forgot to update file_edit, and now see that it's duplicated.

mentat/terminal/loading.py
```diff
@@ -0,0 +1,37 @@
+from tqdm import tqdm
```
PCSwingle, line 1: Would it be best to use PromptToolkit's progress bar? https://python-prompt-toolkit.readthedocs.io/en/master/pages/progress_bars.html I have no experience with this one, but it might mesh better with prompt toolkit and since we already use prompt toolkit I thought we might want to think about it

mentat/terminal/loading.py
```diff
@@ -0,0 +1,37 @@
...
+from mentat.session_stream import StreamMessage
+
+
+class LoadingHandler:
+    def __init__(self):
+        self.pbar = None
+
+    def update(self, message: StreamMessage):
+        """Create or update a loading bar with text and progress value (0-100)"""
+        if message.extra and message.extra.get("terminate"):
```
PCSwingle, line 12: This isn't really related to your PR, but StreamMessage.extra shouldn't be able to be None; I'm actually going to just change it now since it's only 2 lines.

mentat/git_handler.py
```diff
@@ -27,6 +27,8 @@ def get_non_gitignored_files(path: Path) -> set[Path]:
                 text=True,
             ).split("\n"),
         )
+        # windows-safe check if p exists in path
+        if Path(os.path.join(path, p)).exists()
```
PCSwingle, line 31: Is there a reason to use os.path.join instead of path / p?

mentat/feature_filters/llm_feature_filter.py
PCSwingle: I think this could be a great use for the json format mode in the future!

mentat/git_handler.py
granawkins: No - I think copilot wrote it for me 🙈

mentat/terminal/loading.py
granawkins: I wasn't aware of that! Looks like a lot less functionality than tqdm - including some key things I'm using like updating the label, or using it without a context manager. I think it makes sense to stick with tqdm for now.

mentat/terminal/loading.py
granawkins: Ooo pyright didn't like that

jakethekoenig:
The file wasn't changed in this PR so I can't comment inline:
Can you change the `CodeMessageLevel` description strings. "Complete Code" should maybe be "Full file" and "Specific range(s)" should be "Specific ranges".

jakethekoenig:
This pull request breaks autocomplete on commands and conversation history between sessions.

mentat/terminal/client.py
```diff
@@ -56,14 +57,16 @@ async def _cprint_session_stream(self):
...
+
     async def _handle_input_requests(self):
         while True:
             input_request_message = await self.session.stream.recv("input_request")
             # TODO: Make extra kwargs like plain constants
-            if (
-                input_request_message.extra is not None
-                and input_request_message.extra.get("plain")
-            ):
+            if "plain" in input_request_message.extra:
```
jakethekoenig, line 69: Looks like plain is always in `input_request_message` but is usually set to false. I'd revert this change.

jakethekoenig: Just noticed it's @PCSwingle change

jakethekoenig (approved):
Looks good to me outside of those 2 comments

granawkins:
> The file wasn't changed in this PR so I can't comment inline: Can you change the `CodeMessageLevel` description strings. "Complete Code" should maybe be "Full file" and "Specific range(s)" should be "Specific ranges".

I'm changing to "Specific range" (no s) for our update to only-one-interval-per-feature.

PR #268, opened 2023-11-10
Auto Tokens re-added

jakethekoenig:
We should not use the LLMs max context as our limit for auto context as it may be unreasonably large. A setting is readded to limit it. Use LLM is also set to false as it doesn't have guards for context length.

### DIFF
--- a/mentat/code_context.py
+++ b/mentat/code_context.py
@@ -166,7 +166,7 @@ class CodeContext:
             )
         return self._code_message
 
-    use_llm: bool = True
+    use_llm: bool = False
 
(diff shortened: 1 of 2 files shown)


biobootloader:
The original plan was to use gpt-3.5 for feature selection - the idea being even if we used the whole 16k context it would cost just $0.016, and so could easily "pay for itself" by reducing the need to spend more on gpt-4 with a larger context.

As is though I see we are using gpt-4 for the feature selection model. Does gpt-3.5 not work well enough? @granawkins what are your thoughts here?

PR #271, opened 2023-11-11
Fix Git subprocess calls to suppress error output

waydegg:
Silence all subprocess STDERR output from git subprocess calls. If there is an error, an exception will be thrown. We don't rely on STDERR for anything, and atm the Terminal Client gets a bunch of console output for git errors even though we already handle that w/ Exception try-catch blocks.

### DIFF
--- a/mentat/git_handler.py
+++ b/mentat/git_handler.py
@@ -7,14 +7,6 @@ from mentat.errors import UserError
 from mentat.session_context import SESSION_CONTEXT
 
 
-def get_git_diff_for_path(path: Path) -> str:
-    session_context = SESSION_CONTEXT.get()
-    git_root = session_context.git_root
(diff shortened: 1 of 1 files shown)


mentat/git_handler.py
```diff
@@ -10,7 +10,9 @@
 def get_git_diff_for_path(path: Path) -> str:
```
jakethekoenig, line 10: It looks like this function isn't used anywhere. I think it's been deprecated for `get_diff_for_file` since [this commit](https://github.com/AbanteAI/mentat/commit/1f258108bfa1de115acb7db29610e340a94e51b0). I think you should delete it in this PR.

jakethekoenig (approved):
LGTM

PCSwingle (approved):

PR #272, opened 2023-11-11
Add `cwd` parameter to Session

waydegg:
This PR adds the `cwd` parameter to `Session`. The Terminal Client and Python Client use `Path.cwd()` as a default. `cwd` will be used to eventually replace `git_root`.


granawkins:
I think it'd be simpler to just add `Config.cwd`. This will automatically create a `--cwd` command on startup, and won't require a new argument to `Session`.

waydegg:
>Would it work to just add a Config.cwd property?. This would automatically create a --cwd command on startup, and included it in the current Session calls via config.

My thinking was cwd replaces git root, so I put in at the "same level" in the Session Context. I don't see why it couldn't be in the config tho (tbh I feel git root could be too).

granawkins:
> My thinking was cwd replaces git root, so I put in at the "same level" in the Session Context. I don't see why it couldn't be in the config tho (tbh I feel git root could be too).

+1 on git root too. That also automatically gives you a slash command to update it in-session.

PCSwingle:
Hmm, I don't think it really fits in the config; my feeling is that the config should be for user configurable things (like which model they're using, specific files to exclude, etc.) and I don't feel like the cwd fits in there. Not to mention it would make a lot of things a really big pain (we'd have to migrate every single Path to the new cwd). If we ever decide to make a MentatPath object that could automatically handle this stuff we could make it possible to change the cwd mid mentat, but lets not add anything like that for now.

mentat/terminal/client.py
```diff
@@ -222,18 +225,23 @@ def run_cli():
         default=None,
         help="A git tree-ish to diff against the latest common ancestor of",
     )
+    parser.add_argument(
+        "--cwd", default=Path.cwd(), help="The current working directory"
```
PCSwingle, line 229: Having this as a cli option seems a bit confusing; imagine trying to autocomplete some paths but then the cwd changes so those paths don't work. Or you're using a shell like zsh and it auto expands a glob but those files obviously don't fit in the new cwd and things would be a bit of a mess.

PCSwingle: I think we should just keep the cwd as the directory they run it from for now

jakethekoenig: I agree. The point of having cwd in the session is so vscode can initialize a session with the right context right? There doesn't seem to be any reason to have this configurable on the CLI as the user can change the cwd before running mentat.

waydegg: >The point of having cwd in the session is so vscode can initialize a session with the right context right?

Correct.

>There doesn't seem to be any reason to have this configurable on the CLI as the user can change the cwd before running mentat.

If that's how we want people to use Mentat then yeah it wouldn't make sense to have it as a configurable option. My thinking was that users could specify a directory to run mentat in (similar to how you can specify directories to run a ton of command line apps/tools in) without having to change into the directory yourself.

waydegg: > imagine trying to autocomplete some paths but then the cwd changes so those paths don't work

If you're referring to prompt toolkit autocompletions, we can have them based off of the cwd (it's a variable the terminal client can access), and just trigger an autocompletion refresh the instant cwd changes.

I do think that is a separate feature that shouldn't be worked on here, but my thinking is being able to change the cwd from a slash command inside of an active terminal session makes for better UX.

granawkins: You can include paths outside of your cwd when you launch mentat, i.e. `mentat ../path`. Right now our `/include` command doesn't seem to support it, but fixing that looks simple. Is that enough flexibility?

Sticking to one cwd makes sense and feels natural from a ui perspective, to me.

jakethekoenig: Thinking more about it my comment does reflect a lack of imagination. I think it's fine to add this flag

jakethekoenig (approved):
LGTM

granawkins:
I'm fine with this as-is but to circle back, if we have cwd configurable, I think it's simpler to put it in the config.

mentat/terminal/client.py
waydegg: Yeah let's just make it static, can always update to dynamic if it's really wanted/needed later

waydegg: >You can include paths outside of your cwd when you launch mentat, i.e. mentat ../path. Right now our /include command doesn't seem to support it, but fixing that looks simple. Is that enough flexibility?

Yeah we'll be able to do this (this was added in another branch)

>Keeping cwd static makes sense and feels natural from a ui perspective, to me.

Agreed based on your previous point

waydegg: > Thinking more about it my comment does reflect a lack of imagination. I think it's fine to add this flag

After hearing what Paul and Grant said I think we should omit this for now.

PR #273, opened 2023-11-11
Remove `git_root` from Commands

waydegg:


mentat/commands.py
```diff
@@ -279,9 +277,12 @@ async def apply(self, *args: str) -> None:
             return
 
         for i, (feature, score) in enumerate(results, start=1):
-            label = feature.ref()
-            if label.startswith(str(git_root)):
-                label = label[len(str(git_root)) + 1 :]
+            if feature.path.is_relative_to(session_context.cwd):
+                label = f"{feature.path.relative_to(session_context.cwd)}"
```
granawkins, line 281: Could feature.ref have an optional 'cwd' arg? It's quite useful elsewhere for quickly grabbing/comparing the path/interval.

granawkins (changes requested):

PCSwingle (approved):

mentat/commands.py
waydegg: What would the `cwd` arg do in this case?

granawkins:
```
    def ref(self, cwd: Optional[Path] = None):
        path_string = str(self.path) 
        if cwd and self.path.is_relative_to(cwd):
            path_string = self.path.relative_to(cwd)
        if self.level == CodeMessageLevel.INTERVAL:
            interval_string = f"{self.interval.start}-{self.interval.end}"
            return f"{path_string}:{interval_string}"
        return path_string

```

tests/commands_test.py
```diff
@@ -43,6 +43,7 @@ async def test_commit_command(
     assert subprocess.check_output(["git", "status", "-s"], text=True) == ""
 
 
+# TODO: test without git
```
biobootloader, line 46: these test changes coming in another PR?

biobootloader (approved):

tests/commands_test.py
waydegg: Yup! The Session and SessionContext still depend on git_root, so we can't use the non-git testbed in this PR

waydegg: Yup! The Session and SessionContext still depend on git_root, so we can't use the non-git testbed in this PR

jakethekoenig:
Is there a reason this PR isn't merged? Seems better to merge as we go to avoid a bunch of merge conflicts when you're done with the series.

granawkins (approved):

waydegg:
>Is there a reason this PR isn't merged? Seems better to merge as we go to avoid a bunch of merge conflicts when you're done with the series.

Yeah this could have been merged sooner

PR #274, opened 2023-11-12
Remove git root from CodeFileManager, FileEdit

waydegg:
This PR removes `git_root` from `CodeFileManager` and `FileEdit`, making `SESSION_CONTEXT.cwd` the root path to join any relative file paths to.

- `FileEdit.file_path` is now validated during instantiation. The value must be a `Path` type and must be absolute.
- `CodeFileManager.file_lines` now expects all keys to be absolute paths instead of relative paths.


mentat/code_file_manager.py
```diff
@@ -69,10 +67,12 @@ async def write_changes_to_files(
     ):
         session_context = SESSION_CONTEXT.get()
         stream = session_context.stream
-        git_root = session_context.git_root
 
         for file_edit in file_edits:
-            rel_path = Path(os.path.relpath(file_edit.file_path, git_root))
+            rel_path = None
+            if file_edit.file_path.is_relative_to(session_context.cwd):
+                rel_path = file_edit.file_path.relative_to(session_context.cwd)
```
granawkins, line 74:
```suggestion
                rel_path = file_edit.file_path.relative_to(session_context.cwd)
            else:
                rel_path = file_edit.file_path
```
This saves a lot of ink below and makes it a bit clearer what's going on. Since it's only used for displaying, maybe display_path would be better?

mentat/parsers/file_edit.py
```diff
@@ -64,39 +64,53 @@ class FileEdit:
...
     def is_valid(self) -> bool:
         session_context = SESSION_CONTEXT.get()
-        git_root = session_context.git_root
         stream = session_context.stream
         code_context = session_context.code_context
 
-        rel_path = Path(os.path.relpath(self.file_path, git_root))
+        rel_path = None
+        if self.file_path.is_relative_to(session_context.cwd):
+            rel_path = self.file_path.relative_to(session_context.cwd)
```
granawkins, line 81: Same comment re: display_path

tests/code_file_manager_test.py
```diff
@@ -158,6 +158,7 @@ async def test_change_after_creation(
 
 
 @pytest.mark.asyncio
+@pytest.mark.no_git_testbed
```
granawkins, line 161: Nice

granawkins (changes requested):

mentat/code_file_manager.py
```diff
@@ -31,13 +31,11 @@ def __init__(self):
 
     def read_file(self, path: Path) -> list[str]:
         session_context = SESSION_CONTEXT.get()
-        git_root = session_context.git_root
 
-        abs_path = path if path.is_absolute() else Path(git_root / path)
-        rel_path = Path(os.path.relpath(abs_path, git_root))
+        abs_path = path if path.is_absolute() else session_context.cwd.joinpath(path)
```
PCSwingle, line 35: Can we use the / operator instead of joinpath? Also, we don't actually need to check for is_absolute; if you do the / operator and the second argument is absolute already it won't change it at all (although for clarity's sake I'm perfectly fine with keeping the is absolute check)

mentat/code_file_manager.py
PCSwingle: I agree; the rel_path is only used for printing. What if we just call it 'display_path' and use the full path if it isn't relative to cwd?

mentat/code_file_manager.py
```diff
@@ -31,13 +31,11 @@ def __init__(self):
...
         with open(abs_path, "r") as f:
             lines = f.read().split("\n")
-        self.file_lines[rel_path] = lines
+        self.file_lines[abs_path] = lines
```
PCSwingle, line 38: Thanks for making this absolute! I thought it was absolute already, but I guess it must've gotten changed somehow (or I'm misremembering)

mentat/code_file_manager.py
waydegg: Sg, will fix this

mentat/code_file_manager.py
waydegg: > Can we use the / operator instead of joinpath?

Sure

>if you do the / operator and the second argument is absolute already it won't change it at all

Huh I didn't know that was a python "feature" haha. I vote for keeping the explicit check in.

PCSwingle (approved):
LGTM!

granawkins (approved):

PR #279, opened 2023-11-14
Remove git_root from CodeFeature, validate path

waydegg:

### DIFF
--- a/mentat/code_feature.py
+++ b/mentat/code_feature.py
@@ -3,7 +3,6 @@ from __future__ import annotations
 import asyncio
 import logging
 import math
-import os
 from collections import OrderedDict
 from enum import Enum
 from pathlib import Path
@@ -22,7 +21,6 @@ MIN_INTERVAL_LINES = 10
 
 
 def split_file_into_intervals(
-    git_root: Path,
     feature: CodeFeature,
     min_lines: int | None = None,
     user_features: list[CodeFeature] = [],
@@ -35,7 +33,7 @@ def split_file_into_intervals(
         return [feature]
 
     # Get ctags data (name and start line) and determine end line
-    ctags = list(get_ctags(git_root.joinpath(feature.path)))
+    ctags = list(get_ctags(feature.path))
     ctags = sorted(ctags, key=lambda x: int(x[4]))  # type: ignore
     named_intervals = list[tuple[str, int, int]]()  # Name, Start, End
     _last_item = tuple[str, int]()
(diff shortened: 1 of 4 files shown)


mentat/code_feature.py
```diff
@@ -156,36 +158,35 @@ def contains_line(self, line_number: int):
...
-        abs_path = Path(git_root / self.path)
-        rel_path = Path(os.path.relpath(abs_path, git_root))
-        posix_rel_path = Path(rel_path).as_posix()
-        filename = f"{posix_rel_path}"
-        code_message.append(filename)
+        if self.path.is_relative_to(session_context.cwd):
+            code_message_path = self.path.relative_to(session_context.cwd).as_posix()
+        else:
+            code_message_path = self.path.as_posix()
+        code_message.append(str(code_message_path))
```
waydegg, line 170: I'm not sure if this will trip up edit quality from LLMs if we have both relative and absolute paths mixed in. If we wan to keep using relative paths for files outside of the cwd, we can set `walk_up` to `True` (see the [python docs](https://docs.python.org/3/library/pathlib.html#pathlib.PurePath.relative_to) for this) and on any errors just default to absolute paths

PCSwingle: Hmm, yeah I don't think we want to give the LLM absolute paths. I think setting walkup and on error going to absolute path is the best solution.

mentat/code_feature.py
```diff
@@ -132,6 +130,10 @@ def __init__(
                     raise MentatError("CodeFeatures should only have on interval.")
                 self.interval = interval[0]
                 level = CodeMessageLevel.INTERVAL
+
+        if not self.path.is_absolute():
```
PCSwingle, line 134: I think it might be good to name this self.abs_path (just to make extra clear that it's always absolute); not entirely necessary if you don't want to though.

PCSwingle (approved):
This looks great; thanks for putting in so much awesome work on this refactor! Once you make the walk_up change you suggested I think this is ready to merge

mentat/code_feature.py
waydegg: My opinion is that all paths should be assumed to be absolute unless specified otherwise. Like we have `rel_path` and `posix_path` variables for non-absolute paths which I think we should continue doing.

granawkins (approved):

granawkins:
> This looks great; thanks for putting in so much awesome work on this refactor! Once you make the walk_up change you suggested I think this is ready to merge

mentat/code_feature.py
waydegg: So it turns out the `walk_up` argument is not available for python 3.10 (the earliest version we support), so I'll add in a function to do this

PR #282, opened 2023-11-14
fix errors in vscode pyright linter

PCSwingle:
@granawkins I added a new CTAG type; you were previously checking for if line number or name was None; in my own testing I couldn't find anywhere where it would've been None, but I don't have much experience with ctags. Do you know if it can ever be None?

### DIFF
--- a/mentat/code_map.py
+++ b/mentat/code_map.py
@@ -44,7 +44,6 @@ def get_ctags(abs_file_path: Path, exclude_signatures: bool = False):
         line_number: int = tag.get("line")
 
         ctags.add((scope, kind, name, signature, line_number))
-        print((scope, kind, name, signature, line_number))
     return ctags
 
 
--- a/mentat/code_map.py
+++ b/mentat/code_map.py
@@ -6,10 +6,10 @@ from functools import cache
 from pathlib import Path
 from typing import Any
 
+CTAG = tuple[str | None, str, str, str | None, int]
 
-def get_ctags(
-    abs_file_path: Path, exclude_signatures: bool = False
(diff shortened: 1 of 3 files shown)


granawkins:
From the [documentation](https://docs.ctags.io/en/latest/man/ctags.1.html#tags-file-contents-options) and a couple tests, I think you're correct, line number shouldn't ever be None.

mentat/code_map.py
```diff
@@ -6,10 +6,10 @@
 from pathlib import Path
 from typing import Any
 
+CTAG = tuple[str | None, str, str, str | None, int]
```
granawkins, line 9: I like this having its own class. Does it make sense to label these fields, just for clarity - maybe with attr or namedtuple or something:

mentat/code_map.py
```diff
@@ -6,10 +6,10 @@
...
 
-def get_ctags(
-    abs_file_path: Path, exclude_signatures: bool = False
-) -> set[tuple[str | int | None, ...]]:
+
+def get_ctags(abs_file_path: Path, exclude_signatures: bool = False):
```
granawkins, line 12:
```suggestion
def get_ctags(abs_file_path: Path, exclude_signatures: bool = False) -> set[CTAG]:
```

granawkins (approved):
Couple small suggestions, otherwise looks good.

mentat/code_map.py
PCSwingle: I think it makes more sense as a type; it's not like we're doing anything with it other than getting the internals

PR #283, opened 2023-11-14
add conversation tokens and use real token counts in cost tracker

PCSwingle:


granawkins (approved):
LGTM 🚀 

By chance were you able to do an actual comparison of the estimate vs what OpenAI bills you for?

PCSwingle:
> LGTM 🚀
> 
> By chance were you able to do an actual comparison of the estimate vs what OpenAI bills you for?

No, I never did that; I assume it's the same though?

PR #284, opened 2023-11-15
gpt-4-turbo default, rate limit message

PCSwingle:
Rather than add a model 'waterfall', I decided to leave control in the user's hands and just give a warning that they should use /config model <model_name>. I also set gpt-4 turbo as the default model.
image

### DIFF
--- a/mentat/conversation.py
+++ b/mentat/conversation.py
@@ -6,7 +6,7 @@ from enum import Enum
 from timeit import default_timer
 from typing import TYPE_CHECKING
 
-from openai.error import InvalidRequestError
+from openai.error import RateLimitError
 
 from mentat.errors import MentatError
 from mentat.llm_api import (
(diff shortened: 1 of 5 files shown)


mentat/conversation.py
```diff
@@ -188,9 +179,19 @@ async def get_model_response(self) -> list[FileEdit]:
...
-            stream, parser, messages_snapshot
-        )
+        try:
+            parsedLLMResponse, time_elapsed = await self._stream_model_response(
+                stream, parser, messages_snapshot
+            )
+        except RateLimitError:
+            stream.send(
+                "Rate limit recieved from OpenAI's servers using model"
+                f" {config.model}.\nUse /config model <model_name> to switch to a"
```
granawkins, line 189:
```suggestion
                f' {config.model}.\nUse "/config model <model_name>" to switch to a'
```

jakethekoenig (approved):

PR #286, opened 2023-11-15
Remove git_root from `CodeContext`

waydegg:
This PR removes git_root from `CodeContext`, and updates most of the code that includes/excludes `CodeFeature`'s from the code context.

Since the functions that create `CodeFeature`'s from paths (everything in `include_files.py`) are in basically every module, there's a lot of code changes in this PR. I went ahead and commented wherever I think there are noteworthy changes to help make reviewing a bit easier. 

**To-Do before merging**
- [x] Fix for Windows (see pytest results for Windows in GHA)
- [x] Add unit tests for `include_files.py`
- [x] Add unit tests for `CodeContext` (wherever applicable) that remove git from the testbed
- [x] Add docstrings to all functions in `include_files.py`
- [x] Fix how `CodeContext.exclude` handles a directory path


mentat/code_context.py
```diff
@@ -12,80 +12,69 @@
...
-    diff_context: DiffContext
-    features: list[CodeFeature] = []
-    diff: Optional[str] = None
-    pr_diff: Optional[str] = None
-
     def __init__(
         self,
         stream: SessionStream,
-        git_root: Path,
+        git_root: Optional[Path] = None,
```
waydegg, line 39: Will remove this eventually

mentat/code_context.py
```diff
@@ -12,80 +12,69 @@
...
 from mentat.utils import sha256
 
 
 class CodeContext:
-    include_files: dict[Path, list[CodeFeature]]
-    ignore_files: set[Path]
-    diff_context: DiffContext
-    features: list[CodeFeature] = []
-    diff: Optional[str] = None
-    pr_diff: Optional[str] = None
```
waydegg, line 38: Removed this since it's mostly redundant

mentat/code_context.py
```diff
@@ -180,14 +168,20 @@ async def _get_code_message(
...
+                code_message += [
+                    "Diff References:",
+                    f' "-" = {self.diff_context.name}',
+                    ' "+" = Active Changes',
+                    "",
+                ]
+
+                if len(self.include_files) == 0:
+                    for file in self.diff_context.files:
+                        self.include(file)
```
waydegg, line 183: This logic used to be in `CodeContext.set_paths`. Because `set_paths` is was removed, I put it here whenever a new code message needs to be generated.

mentat/code_context.py
```diff
@@ -271,37 +273,133 @@ def _get_all_features(
...
-        # matches files but doesn't match any files in context, we won't know what that glob is
-        # and can't return it as an invalid path
-        paths, invalid_paths = get_include_files([path], [])
-        removed_paths = list[Path]()
-        for new_path in paths.keys():
-            if new_path in self.include_files:
-                removed_paths.append(new_path)
-                del self.include_files[new_path]
-        return removed_paths, invalid_paths
+    def include(
```
waydegg, line 287: Updated the naming of `include_file` to `include` since we are including **Code Feature(s)**

mentat/code_context.py
```diff
@@ -271,37 +273,133 @@ def _get_all_features(
...
+        self, path: Path | str, ignore_patterns: Iterable[Path | str] = []
+    ) -> Set[Path]:
+        """Add code to the context
+
+        '.' is replaced with '*' (recusively search the cwd)
```
waydegg, line 292: I have this logic here since I believe this is what people are meaning when they type "."

mentat/code_context.py
```diff
@@ -271,37 +273,133 @@ def _get_all_features(
...
+
+        return included_paths
+
+    def exclude(self, path: Path | str) -> Set[Path]:
+        """Remove code from the context
+
+        # NOTE: should this be the intended behavior?
+        Code Features are removed from the context and are excluded from any future `include` calls (unless explicity specified).
+        For paths that are files or file intervals, if they don't already exist in the Code Context an Exception is thrown.
+        For file intervals, if the interval doesn't *exactly match* any intervals in the code context, nothing is removed.
```
waydegg, line 335: Would like feedback/thoughts on this

mentat/code_file_manager.py
```diff
@@ -44,20 +44,20 @@ def _create_file(self, code_context: CodeContext, abs_path: Path):
...
 
     def _rename_file(
         self, code_context: CodeContext, abs_path: Path, new_abs_path: Path
     ):
         logging.info(f"Renaming file {abs_path} to {new_abs_path}")
+        code_context.exclude(abs_path)
         os.rename(abs_path, new_abs_path)
-        code_context.include_file(new_abs_path)
-        code_context.exclude_file(abs_path)
+        code_context.include(new_abs_path)
```
waydegg, line 60: Ordering of this matters since `.exclude()` will throw an Exception if `abs_path` does not exist

mentat/include_files.py
```diff
@@ -44,104 +23,225 @@ def is_file_text_encoded(abs_path: Path):
...
 
 
-def get_ignore_files(ignore_paths: list[Path]) -> set[Path]:
-    """Returns a set of files to ignore from a list of ignore paths."""
+def match_path_with_patterns(path: Path, patterns: Set[str]) -> bool:
+    """Check if the given absolute path matches any of the patterns.
 
-    ignore_paths, _ = expand_paths(ignore_paths)
+    TODO: enforce valid glob patterns? Right now we allow glob-like
+    patterns (the one's that .gitignore uses). This feels weird imo.
```
waydegg, line 86: Thoughts on this?

mentat/session.py
```diff
@@ -78,6 +79,8 @@ async def _main(self):
             code_context.display_context()
             await conversation.display_token_count()
         except MentatError as e:
+            # BUG: this is not getting streamed in the Python and Terminal clients.
+            # It makes it look like Mentat is hanging forever.
```
waydegg, line 83: This might just happen when using Pytest. Need to verify.

tests/clients/python_client_test.py
```diff
@@ -24,8 +25,7 @@ async def test_editing_file_auto_accept(mock_call_llm_api, mock_setup_api_key):
         @@code
         # Line 2
         @@end""")])
-
-    python_client = PythonClient(paths=["."])
+    python_client = PythonClient(cwd=Path.cwd(), paths=["."])
```
waydegg, line 28: Need to specify `cwd` here since pathlib/os will get the correct `cwd` that pytest is working out of.

dev-requirements.txt
```diff
@@ -4,6 +4,6 @@ gitpython==3.1.37
 ipdb==0.13.13
 isort==5.12.0
 pip-licenses==4.3.3
-pyright==1.1.329
+pyright==v1.1.335
```
waydegg, line 7: Added since GHA was complaining about the version. This likely already has been updated by the time I rebase w/ the parent branch

mentat/code_feature.py
```diff
@@ -55,10 +55,10 @@ def split_file_into_intervals(
                 )
         else:
             line_number = 1
-        if i == len(ctags) - 1:
-            named_intervals.append((str(key), int(line_number), n_lines))
+        if i == len(ctags) - 1:  # type: ignore
+            named_intervals.append((str(key), int(line_number), n_lines))  # type: ignore
         else:
-            _last_item = (key, int(line_number))
+            _last_item = (key, int(line_number))  # type: ignore
```
waydegg, line 61: These pyright errors might already be fixed in the parent branch

mentat/auto_context.py
waydegg, line 1: ATM there aren't any unit tests for this module (however there might be tests by the time I rebase)

mentat/code_context.py
```diff
@@ -271,37 +273,133 @@ def _get_all_features(
...
+            # directory
+            elif validated_path.is_dir():
+                if validated_path not in self.include_files:
+                    session_context.stream.send(
+                        f"Directory path {validated_path} not in context",
+                        color="light_red",
+                    )
+                    return excluded_paths
+                excluded_paths.add(validated_path)
+                del self.include_files[validated_path]
```
waydegg, line 388: Need to fix this. All paths inside of `validated_path` (a directory) should be removed *instead* of checking if `validated_path` is in `self._include_files` (which is impossible)

Ty for @PCSwingle for finding this earlier in https://github.com/AbanteAI/mentat/pull/246#discussion_r1385344004

mentat/include_files.py
```diff
@@ -44,104 +23,224 @@ def is_file_text_encoded(abs_path: Path):
...
+        _interval_path, interval_str = str(abs_path).rsplit(":", 1)
+        interval_path = Path(_interval_path)
+        if not interval_path.exists():
+            raise PathValidationError(f"File {interval_path} does not exist")
+        if check_for_text and not is_file_text_encoded(interval_path):
+            raise PathValidationError(f"Unable to read file {interval_path}")
+        intervals = parse_intervals(interval_str)
+        if len(intervals) == 0:
+            raise PathValidationError(
+                f"Unable to parse intervals for path {interval_path}"
```
waydegg, line 67: Should we also validate that each interval actually exists in the specified path?

mentat/code_context.py
PCSwingle: Don't we already recursively search directories, so shouldn't "." already work like that? Let's not replace it with "*".

mentat/code_context.py
PCSwingle: I don't like this behavior; I think exclude should just remove the file (or all files in the directory, or all matching files for a glob pattern, etc.) whenever it gets called. Otherwise, we're going to end up with some pretty unintuitive behavior (you /exclude'd a directory? Suddenly nothing is ever being added in when you try to include it later).

mentat/session.py
PCSwingle: This looks like you haven't merged into main for a while; I think I fixed all the error stuff sometime last week.

mentat/include_files.py
```diff
@@ -1,38 +1,17 @@
...
-                intervals = None
-            if Path(p).exists() and intervals:
-                for interval in intervals:
-                    globbed_paths.add(f"{p}:{interval.start}-{interval.end}")
-            else:
-                invalid_paths.add(str(path))
-    return [Path(path).resolve() for path in globbed_paths], list(invalid_paths)
-
-
+# TODO: replace this with something that doesn't load the file into memory
```
PCSwingle, line 14: I don't think there's any other way to definitively check if a file is utf-8 encoded or not

mentat/git_handler.py
```diff
@@ -84,17 +85,20 @@ def _get_git_root_for_path(path: Path) -> Path:
...
+            raise UserError()
+        else:
+            return
 
 
 def get_shared_git_root_for_paths(paths: list[Path]) -> Path:
     git_roots = set[Path]()
     for path in paths:
-        git_root = _get_git_root_for_path(path)
+        git_root = cast(Path, get_git_root_for_path(path))
```
PCSwingle, line 98: ? Why are we casting it here? I assume you're going to change this function in the future (so that len(git_roots) == 0 isn't an error); is this just a placeholder for now? I would rather you change it to only adding the git root into git_roots if it isn't None just to make sure we don't forget in the future (and because this can lead to returning None)

mentat/config.py
```diff
@@ -72,7 +73,7 @@ class Config:
 
     # Context specific settings
     file_exclude_glob_list: list[str] = attr.field(
-        default=[],
+        default=cast(List[str], []),
```
PCSwingle, line 76: I fixed these errors in #282; instead of using default, I used factory (which you need for objects or attr's copies the reference to the empty list), and you can also just do list[str]() instead of casting an empty list.

mentat/include_files.py
```diff
@@ -44,104 +23,224 @@ def is_file_text_encoded(abs_path: Path):
...
+
+    # Validate different path types
+    # File
+    if abs_path.is_file():
+        if not abs_path.exists():
+            raise PathValidationError(f"File {abs_path} does not exist")
+        if check_for_text and not is_file_text_encoded(abs_path):
+            raise PathValidationError(f"Unable to read file {abs_path}")
+    # File interval
+    elif len(str(abs_path).rsplit(":", 1)) > 1:
```
PCSwingle, line 57: This has to be done after checking if it's a directory and/or glob path (since, especially on windows, valid directories might have a :)

mentat/include_files.py
```diff
@@ -44,104 +23,224 @@ def is_file_text_encoded(abs_path: Path):
...
+            raise PathValidationError(f"File {interval_path} does not exist")
+        if check_for_text and not is_file_text_encoded(interval_path):
+            raise PathValidationError(f"Unable to read file {interval_path}")
+        intervals = parse_intervals(interval_str)
+        if len(intervals) == 0:
+            raise PathValidationError(
+                f"Unable to parse intervals for path {interval_path}"
             )
+    # Directory
+    elif abs_path.is_dir():
```
PCSwingle, line 70: Is this a TODO?

mentat/include_files.py
```diff
@@ -44,104 +23,224 @@ def is_file_text_encoded(abs_path: Path):
...
+        pass
+    # Glob pattern
+    elif re.search(r"[\*\?\[\]]", str(path)):
```
PCSwingle, line 73: Why not just check if glob.glob returns anything?

mentat/session.py
```diff
@@ -45,7 +45,7 @@ def __init__(
 
         cost_tracker = CostTracker()
 
-        code_context = CodeContext(stream, git_root, diff, pr_diff)
+        code_context = CodeContext(stream, git_root, diff, pr_diff, ignore_paths)
```
PCSwingle, line 48: Ignore paths is solely for auto context; exclude paths is for the user included context (although soon these 2 things will be merged anyways)

mentat/include_files.py
```diff
@@ -44,104 +23,224 @@ def is_file_text_encoded(abs_path: Path):
...
-    files_from_dirs = [
-        file
-        for file in files_from_dirs
-        if file.path.resolve() not in glob_excluded_files
-    ]
-    files_direct.update(files_from_dirs)
+def get_paths_for_directory(
+    path: Path,
+    include_patterns: Iterable[Path | str] = [],
+    ignore_patterns: Iterable[Path | str] = [],
```
PCSwingle, line 107: This should be exclude_paths

mentat/include_files.py
PCSwingle: I don't think we need this function at all; why aren't we just using glob.glob like we were before? Just match all non-gitignored files in the directory, and then subtract all globbed files.

mentat/include_files.py
```diff
@@ -44,104 +23,224 @@ def is_file_text_encoded(abs_path: Path):
         return False
 
 
-def abs_files_from_list(paths: list[Path], check_for_text: bool = True):
```
PCSwingle, line 47: Why do we need to rewrite all of this logic? As far as I'm aware the only git related part of include_files was the git ignore directory logic

mentat/include_files.py
```diff
@@ -1,38 +1,17 @@
...
-from mentat.code_feature import CodeFeature, parse_intervals
-from mentat.git_handler import get_non_gitignored_files
+from mentat.code_feature import CodeFeature
+from mentat.errors import PathValidationError
+from mentat.git_handler import get_git_root_for_path, get_non_gitignored_files
+from mentat.interval import parse_intervals
 from mentat.session_context import SESSION_CONTEXT
 
 
-def expand_paths(paths: list[Path]) -> tuple[list[Path], list[str]]:
```
PCSwingle, line 13: Same up here; why are we rewriting this?

mentat/include_files.py
```diff
@@ -44,104 +23,224 @@ def is_file_text_encoded(abs_path: Path):
...
+            code_feature = CodeFeature(
+                f"{interval_path}:{interval.start}-{interval.end}"
+            )
+            code_features.add(code_feature)
+    # Glob pattern
+    else:
+        root_parts: List[str] = []
+        pattern: str | None = None
+        for i, part in enumerate(validated_path.parts):
+            if re.search(r"[\*\?\[\]]", str(part)):
```
PCSwingle, line 218: Again, why aren't we just using glob.glob?

mentat/code_context.py
```diff
@@ -271,37 +273,132 @@ def _get_all_features(
...
+        path = Path(path)
+
+        excluded_paths: Set[Path] = set()
+
+        try:
+            validated_path = validate_and_format_path(
+                path, session_context.cwd, check_for_text=False
+            )
+            # file
+            if validated_path.is_file():
```
PCSwingle, line 354: We use this exact set of file, file interval, directory, glob pattern if statements 3 or 4 times; there shouldn't need to be a difference at all. Why don't we do it like we did before where we just call get_include_files on the path and include_files handled everything in one function?

PCSwingle (changes requested):
I feel like we're rewriting a lot of code that doesn't need to be rewritten here. Why don't we keep the original include_files logic, use glob.glob still, and just remove the git ignore part if we aren't in a git project?

mentat/code_context.py
waydegg: "." for git recursively searches, but it does not for glob patterns.

mentat/code_context.py
waydegg: Yeah I agree w/ that. I actually never added the logic that excludes any file/patterns from future `include` calls, so I just won't implement that

waydegg:
>I feel like we're rewriting a lot of code that doesn't need to be rewritten here. Why don't we keep the original include_files logic, use glob.glob still, and just remove the git ignore part if we aren't in a git project?

`glob.glob` will get recursively traverse every folder w/o any regard to what's ignored which can be super inefficient. This wasn't a problem before b/c git repos get an already filtered list of files with `git ls-files`. 

For example, if I did `mentat .` on a non-git directory with a virtual environment, just using `glob.glob` and filtering afterwards (like we had originally) is going to first get every file in the venv then filter out every file. `get_paths_for_directory` traverses the file system in a much smarter/efficient way so that situations like this don't happen.

mentat/include_files.py
waydegg: See https://github.com/AbanteAI/mentat/pull/286#issuecomment-1821996841

mentat/include_files.py
waydegg: See https://github.com/AbanteAI/mentat/pull/286#issuecomment-1821996841

mentat/include_files.py
waydegg: See https://github.com/AbanteAI/mentat/pull/286#issuecomment-1821996841

mentat/include_files.py
waydegg: Because calling `glob.glob` might return a huge amount of files unnecessarily. Here we are just doing a no-op check of `path` containing valid glob pattern characters

waydegg: Actually just remembered `glob.iglob` is a thing. Just added that for more robust validation

mentat/include_files.py
waydegg: See https://github.com/AbanteAI/mentat/pull/286#issuecomment-1821996841

mentat/include_files.py
waydegg: https://github.com/AbanteAI/mentat/pull/286#issuecomment-1821996841

mentat/code_context.py
waydegg: >there shouldn't need to be a difference at all

Could you elaborate on this? Not sure I understand

biobootloader:
@waydegg let me know if you want me to take a look at anything here, or if you're still changing things let me know when you want me to

mentat/code_context.py
PCSwingle: But "." should point to a directory (the cwd) which assuming we didn't change how including directories works should auto include all sub files and subdirectories

mentat/include_files.py
PCSwingle: Is this a problem though? This adds a lot of complexity and room for error compared to what we were doing before for what seems like not much benefit

PCSwingle:
> > I feel like we're rewriting a lot of code that doesn't need to be rewritten here. Why don't we keep the original include_files logic, use glob.glob still, and just remove the git ignore part if we aren't in a git project?
> 
> `glob.glob` will get recursively traverse every folder w/o any regard to what's ignored which can be super inefficient. This wasn't a problem before b/c git repos get an already filtered list of files with `git ls-files`.
> 
> For example, if I did `mentat .` on a non-git directory with a virtual environment, just using `glob.glob` and filtering afterwards (like we had originally) is going to first get every file in the venv then filter out every file. `get_paths_for_directory` traverses the file system in a much smarter/efficient way so that situations like this don't happen. The user can specify "venv" in their list of excluded patterns and `get_paths_for_directory` would see this directory that matches this pattern, and then completely ignore it and not traverse.

How big of a problem is this though? If the user is doing mentat ., then they're putting their whole project into context anyway which makes me think it probably isn't the largest project so this shouldn't be a problem anyways. I think this is all adding a lot of complexity and ways to fail that just using a simple glob.glob like we did before would fix, for not much upside. This is only ever going to be a problem when the user includes a full directory with very few unignored files and a whole lot of ignored files; I see people using mentat more on specific files or subdirectories and eventually just having everything auto included into context by the llm.

waydegg:
>let me know if you want me to take a look at anything here, or if you're still changing things let me know when you want me to

Will do! Knocking out the to-do list items rn and will request a review shortly

mentat/code_context.py
waydegg: Ahh I see what you mean now. Yeah I'll remove this

mentat/include_files.py
waydegg: You can read it in chunks and validate each chunk. Out-of-scope for this PR anways

biobootloader:
@waydegg @PCSwingle responding to related conversations all here:

> TODO: enforce valid glob patterns? Right now we allow glob-like
    patterns (the one's that .gitignore uses). This feels weird imo.

I don't recall the differences between valid glob patterns and glob-like patterns, but we need to support `.gitignore` anyway, so what would enforcing valid glob patterns mean?

> The main need for this would be having folders like venv, node_modules, or __pycache__ folders that people probably would like to have blacklisted globally. That way when ppl are in projects that have these folders that aren't in a git directory, they don't have to think twice about starting a mentat session and can just do mentat . to add everything

I agree with this, I don't want Mentat slowed down in these scenarios. I think the extra complexity is manageable as we can keep the initial file finding separated/independent from the rest of the code. It is important to test it well. Perhaps we can make some automatic tests that compare the paths we find to what git finds and ensures the lists are identical?

waydegg:
>I don't recall the differences between valid glob patterns and glob-like patterns, but we need to support .gitignore anyway, so what would enforcing valid glob patterns mean?

A key difference that's relevant to us is excluding file/folder names at _any_ directory level. You just add names into a .gitignore and have them be ignored recursively, however with GNU glob patters you need to append/prepend `**` to get the same effect.

For example:

`*.py` in a `.gitignore` file finds all python files recursively

`*.py` with python's `glob.glob()` finds all python files in the current working directory. To get the same effect as `.gitignore` you need `glob.glob("**/*.py", recursive=True)`

The (dumb) solution to this is to check the whole path and check each part of the path for matches. It's not very efficient but it works lol

PCSwingle:
> > The main need for this would be having folders like venv, node_modules, or **pycache** folders that people probably would like to have blacklisted globally. That way when ppl are in projects that have these folders that aren't in a git directory, they don't have to think twice about starting a mentat session and can just do mentat . to add everything
> 
> I agree with this, I don't want Mentat slowed down in these scenarios. I think the extra complexity is manageable as we can keep the initial file finding separated/independent from the rest of the code. It is important to test it well. Perhaps we can make some automatic tests that compare the paths we find to what git finds and ensures the lists are identical?

I don't think it would really slow Mentat down at all though; I can do "./**" on mentat right now and it's still the same bootup time as normal. I'm ok with changing though if you still want to.

> > I don't recall the differences between valid glob patterns and glob-like patterns, but we need to support .gitignore anyway, so what would enforcing valid glob patterns mean?
> 
> A key difference that's relevant to us is excluding file/folder names at _any_ directory level. You just add names into a .gitignore and have them be ignored recursively, however with GNU glob patters you need to append/prepend `**` to get the same effect.
> 
> For example:
> 
> `*.py` in a `.gitignore` file finds all python files recursively
> 
> `*.py` with python's `glob.glob()` finds all python files in the current working directory. To get the same effect as `.gitignore` you need `glob.glob("**/*.py", recursive=True)`
> 
> The (dumb) solution to this is to check the whole path and check each part of the path for matches. It's not very efficient but it works lol

I think we want regular GNU glob patterns; it makes more sense to me for *.py to get only py files in one directory, and **/*.py for all py files.

mentat/code_context.py
PCSwingle: We only want to add the diff files if the user actually did --diff or --pr-diff;
```suggestion
                if len(self.include_files) == 0 and (self.diff or self.pr_diff):
                    for file in self.diff_context.files:
                        self.include(file)
```

mentat/code_feature.py
PCSwingle: Since we'll be merging in to main anyways, it might be best to remove the type ignores now so we don't forget later? Up to you

mentat/git_handler.py
```diff
@@ -84,17 +85,23 @@ def _get_git_root_for_path(path: Path) -> Path:
...
     for path in paths:
-        git_root = _get_git_root_for_path(path)
+        git_root = get_git_root_for_path(path)
+        if git_root is None:
+            logging.error(f"File {path} isn't part of a git project.")
+            raise UserError()
         git_roots.add(git_root)
     if not paths:
-        git_root = _get_git_root_for_path(Path(os.getcwd()))
+        git_root = cast(Path, get_git_root_for_path(Path(os.getcwd())))
```
PCSwingle, line 104: Could we get rid of the cast here too, so we don't forget?

mentat/include_files.py
```diff
@@ -44,104 +25,281 @@ def is_file_text_encoded(abs_path: Path):
...
+
+    all_include_patterns = set(str(p) for p in include_patterns)
+    all_exclude_patterns = set(str(p) for p in exclude_patterns)
+
+    for root, dirs, files in os.walk(path, topdown=True):
+        root = Path(root)
+
+        if get_git_root_for_path(root, raise_error=False):
+            dirs[:] = []
+            git_non_gitignored_paths = get_non_gitignored_files(root)
```
PCSwingle, line 198: get_non_git_ignore should be recursive already; I think putting inside of os.walk is going to be really slow. If we have a git root, we shouldn't need to do os.walk (just get_non_gitignored_files).

mentat/session.py
PCSwingle, line 48: I think this is supposed to be exclude_paths (although I could be wrong)

PCSwingle:
@waydegg just went over it on more time and it looks great! Thanks for getting this done; it's definitely a super big project. Left a couple of mostly small comments, but after those are resolved (and we decide if we're going to not allow gitignore glob patterns; I think we shouldn't) I think we can merge this. Thanks!

mentat/include_files.py
waydegg: Implemented this

mentat/session.py
waydegg: Nah it's supposed to be `ignore_paths`. Should update the naming at some point tho

mentat/interval.py
```diff
@@ -23,7 +23,7 @@ class Interval:
     end: int | float = attr.field()
 
     def contains(self, line_number: int):
-        return self.start <= line_number < self.end
```
waydegg, line 26: @PCSwingle I added this in b/c intervals with a single line were failing to be loaded into the CodeFeature's code message

E.g. `some_file.py:1,3-5` would only read lines 3-5 with the old logic

PCSwingle: Hmm, I see the problem. We'll probably need to remake some adjustments to the place where we automatically create intervals, but we can work this out later.

PCSwingle: How about for now we keep it exclusive, but in parse_intervals a few lines up we just add one to the ending interval if it's a single line? Should be on line 12 of this file.

PCSwingle: I think only interval.end should be exclusive; I agree that we should convert in parse_intervals

biobootloader:
> A key difference that's relevant to us is excluding file/folder names at any directory level. You just add names into a .gitignore and have them be ignored recursively, however with GNU glob patters you need to append/prepend ** to get the same effect.

> I think we want regular GNU glob patterns; it makes more sense to me for *.py to get only py files in one directory, and **/*.py for all py files.

ah, ok makes sense! yeah let's just support GNU glob for what the user enters (but still support .gitignores as well)

mentat/include_files.py
waydegg: ^ yup

mentat/code_context.py
```diff
@@ -13,76 +13,76 @@
...
-    pr_diff: Optional[str] = None
-
     def __init__(
         self,
         stream: SessionStream,
-        git_root: Path,
+        git_root: Optional[Path] = None,
         diff: Optional[str] = None,
         pr_diff: Optional[str] = None,
+        exclude_patterns: Iterable[Path | str] = [],
```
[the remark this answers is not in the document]
waydegg: Yeah that's a good point, I'll update this

mentat/code_context.py
```diff
@@ -286,37 +297,181 @@ def _get_all_features(
...
+
+        included_paths: Set[Path] = set()
+        try:
+            code_features = get_code_features_for_path(
+                path=path,
+                cwd=session_context.cwd,
+                exclude_patterns=[
+                    *exclude_patterns,
+                    *self.exclude_patterns,
+                    *session_context.config.file_exclude_glob_list,
```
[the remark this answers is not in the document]
waydegg: That is all correct. I updated `self.exclude_patterns` to be `self.ignore_patterns` based on your earlier comment btw

PCSwingle (approved):
Thanks for getting this done!

PR #287, opened 2023-11-15
add default arguments for diff and pr-diff if no arg given

PCSwingle:


PCSwingle:
@granawkins I set the default for pr-diff to main, but I'm not entirely sure that's the most intuitive default. What do you think?

mentat/diff_context.py
```diff
@@ -133,6 +133,8 @@ def __init__(
 
         meta = get_treeish_metadata(git_root, target)
         name += f'{meta["hexsha"][:8]}: {meta["summary"]}'
+        if target == "HEAD":
+            name = "HEAD (last commit)"
```
granawkins, line 137: If you take this line out, the name would be like `HEAD: 2ab23bc8: add default args for diff`. To me that's preferable to just `HEAD (last commit)` if I've added the `--diff` flag.

granawkins:
> @granawkins I set the default for pr-diff to main, but I'm not entirely sure that's the most intuitive default. What do you think?

I think that's the most intuitive ya, but it'd be `master` on older repos. So we'd have to.. check? make an assumption? maybe pr-diff doesn't need a default like this?

mentat/diff_context.py
PCSwingle: Hmm I see what you're saying, but the name doesn't have HEAD at the start; it's just ```c7b6123b: add conversation tokens and use real token counts in cost tracker (#283) | 1 files | 2 lines``` which I find a lot worse than ```HEAD (last commit)``` (since it doesn't specify that it's HEAD)

PCSwingle:
@biobootloader What do you think about the pr-diff default? I think it could be ok to just base it off of main, but I also see the side of not having a default for pr-diff.

mentat/diff_context.py
jakethekoenig: I vote with Grant. They should know it's diffed with head when they ran `--diff`. But they may not know the file/line counts. Maybe we could split the difference with `name += " (HEAD)"` like it appears in `git log`?

jakethekoenig:
> @biobootloader What do you think about the pr-diff default? I think it could be ok to just base it off of main, but I also see the side of not having a default for pr-diff.

Currently mentat crashes on invalid treeish so I don't think we should risk a default that may be invalid. It looks like there is a way to check which branch is [default](https://stackoverflow.com/a/50056710). But it would be sort of complicated and maybe not worth it.

mentat/diff_context.py
PCSwingle: Oh the file and line counts are always there for both; and they might not know it's diffed with HEAD when they run just --diff (because HEAD is the default). I could be ok with adding (HEAD) to the end, but I still think it makes more sense to not have the commit hash or description there; after all, this is a diff with HEAD, so it's just the current working changes.

PCSwingle:
Hmm... first off, I think we need to fix it crashing on an invalid tree-ish. But I think I agree that maybe pr-diff just shouldn't have a default.

mentat/diff_context.py
jakethekoenig: You've convinced me. I guess I don't feel that strongly one way or the other anyway. One more open would be to write just HEAD since they should no that means it's diffed with the last commit. But all the options seem fine.

jakethekoenig (approved):

mentat/diff_context.py
granawkins: First off, I wasn't happy with how get_code_message was handling this, so in #291 I have this update in CodeContext:
```
        diff = self.diff_context.get_display_context()
        if diff:
            stream.send(f"{prefix}Diff: {diff}", color="green")
```
So if there are no files, the Diff display is omitted altogether.

After reading you guys' convo, I think we should omit the hexsha/description altogether, and just display `<target> | <i> files | <i> lines`. This is concise, and if they want to see details, they can copy and run a git diff in the terminal.

PR #289, opened 2023-11-15
Vision API added

jakethekoenig:
Type `/screenshot path` where path can be a url or local file and mentat will open it with selenium and take a screenshot and put it in the LLMs context. You can see it with `/conversation`.

* A Message class is introduced which holds onto the role, text and optional image path. It handles representing it in the form the llm wants.
* The conversation viewer no longer puts the messages in pre so you don't have to side scroll them anymore. New lines in messages are replaced with \<br\>.
* The new transcript logging format is breaking with the old one so you'll have to clear out your logs before you use this.

Todo:
- [x] Automatically switch the user to `gpt-vision-preview` with a warning when they run screenshot.
- [x] `/screenshot` should work with no arguments and screenshot the open browser.
- [x] Tests!

Long term
* Because we're using selenium we could allow the LLM to interact with the DOM in more complex ways. We could let it request parts of the DOM. We could let it navigate the page and explore.


tests/license_check.py
```diff
@@ -18,6 +18,7 @@
     "Apache 2.0",
     "BSD 3-Clause",
     "ISC License (ISCL)",
+    "Historical Permission Notice and Disclaimer (HPND)",
```
PCSwingle, line 21: I think you'll need to add just BSD (without the 'License' at the end)

mentat/commands.py
```diff
@@ -387,3 +388,31 @@ def argument_names(cls) -> list[str]:
     @classmethod
     def help_message(cls) -> str:
         return "Set a configuration option or omit value to see current value."
+
+
+class ScreenshotCommand(Command, command_name="screenshot"):
+    async def apply(self, *args: str) -> None:
+        if not args:
+            raise UserError("Screenshot command requires a URL or string to capture.")
```
PCSwingle, line 396: Let's not raise a UserError; these should be used to exit mentat with an error message; in this case, we just want to fail the command

mentat/message.py
```diff
@@ -0,0 +1,53 @@
...
+
+
+class MessageRole(Enum):
+    System = "system"
+    User = "user"
+    Assistant = "assistant"
+
+
+@attr.define
+class Message:
```
PCSwingle, line 17: Since none of this will be necessary once I get the OpenAI API 1.0 PR through, I'm wondering if we should hold off on this PR until then?

mentat/vision/vision_manager.py
```diff
@@ -0,0 +1,49 @@
...
+
+from mentat.utils import mentat_dir_path
+
+
+@attr.define
+class VisionManager:
+    driver: webdriver.Chrome | None = attr.field(default=None)
+
+    def init(self) -> None:
+        "This opens a browser with no page open so we don't want to call it until the user actually wants it."
```
PCSwingle, line 18: How about we change the name of this function to _open_browser?

mentat/vision/vision_manager.py
```diff
@@ -0,0 +1,49 @@
...
+        if self.driver is None:
+            service = Service(ChromeDriverManager().install())
+            self.driver = webdriver.Chrome(service=service)
+
+    def open(self, path: str) -> None:
+        self.init()
```
PCSwingle, line 24: Does this open a new browser every time and does it close the browser afterwards or not?

PCSwingle:
Looks great to me! I left a few comments. I'm mostly wondering if we want to delay merging this until OpenAI gets updated; I'm not entirely sure how I feel and I'd be ok with either. The only downside to merging this early is that we'd have to go ahead and remove all of the Message stuff, but that would be doable.

mentat/llm_api.py
```diff
@@ -117,13 +117,14 @@ async def call_llm_api(
             messages=messages,
             temperature=config.temperature,
             stream=True,
+            max_tokens=4000,
```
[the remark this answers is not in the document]
jakethekoenig: Oh yeah it's sort of weird: The open ai api returns a max of 30 tokens by default if you include an image. I spent a long time debugging thinking it must be my problem. Not sure why they added that behavior.

mentat/message.py
jakethekoenig: Yeah I agree :( I should have looked at the API before doing this.

mentat/vision/vision_manager.py
jakethekoenig: It opens in the same browser. The browser is automatically closed when mentat quits (or the user can close it themselves).

jakethekoenig:
> Looks great to me! I left a few comments. I'm mostly wondering if we want to delay merging this until OpenAI gets updated; I'm not entirely sure how I feel and I'd be ok with either. The only downside to merging this early is that we'd have to go ahead and remove all of the Message stuff, but that would be doable.

I'm down to delay until after the OpenAI update. I still have to write tests for this and stuff so it isn't that close to being done. And I don't know how similar my Message class is to the one OpenAI has. I don't want to write code in the conversation_viewer that relies on my abstraction that you'd have to then rewrite.

mentat/llm_api.py
```diff
@@ -153,18 +154,34 @@ def chunk_to_lines(chunk: Any) -> list[str]:
...
+            return count_tokens(message.text, model)
+        else:
+            count = count_tokens(message.text, model)
+            img = Image.open(message.image_path)
+            size = img.size
+            scale = min(1, 2048 / max(size))
+            size = (int(size[0] * scale), int(size[1] * scale))
+            scale = min(1, 768 / min(size))
+            size = (int(size[0] * scale), int(size[1] * scale))
+            count += 85 + 170 * (size[0] // 512) * (size[1] // 512)
```
jakethekoenig, line 182: Note to self: Need to round up not down here.

mentat/resources/templates/conversation_viewer.jinja
```diff
@@ -16,11 +16,31 @@
...
+                            {% elif submessage["type"] == "image_url" %}
+                                <img src="{{ submessage['image_url']['url'] }}">
+                            {% endif %}
+                        {% endfor %}
+                    {% endif %}
                     {% if message["prior_messages"] is not none %}
                         {% for seen_message in message["prior_messages"] %}
                         <div class="message viewpoint-message {{seen_message['role']}} hidden">
-                            {{ seen_message['content'] |e|replace('\n', '<br>')|safe }}
+                            {% if seen_message["content"] is string %}
```
jakethekoenig, line 33: There's some code duplication here. I tried to use mentat to make a macro to fix it and it failed. I tried myself and it seemed like the arguments to the macro were passed as plaintext. I don't think it's that important to fix but if anyone who has experience with jinja macros it'd be nice.

mentat/llm_api_handler.py
```diff
@@ -71,9 +74,25 @@ def conversation_tokens(messages: list[ChatCompletionMessageParam], model: str):
         # every message follows <im_start>{role/name}\n{content}<im_end>\n
         num_tokens += 4
         for key, value in message.items():
-            if not isinstance(value, str):
-                continue
-            num_tokens += len(encoding.encode(value))
+            if isinstance(value, list):
+                for entry in value:  # type: ignore
```
jakethekoenig, line 78: There are a lot of ignores here. I think Pyright should have enough information given the classes to know entry is of type `ChatCompletionContentPartParam`. Maybe it would help if instead of iterating over items we just accessed the content which should be the only source of tokens anyway. For our use case the only attributes should be role and content. If we start using function calls there could be others.

I'm not sure as written it's exactly correct because when we iterate over items one of them will be the role which should already be counted in line 72. In my `llm_api_handler_test` line 33 I add 5 at the end but I think adding 4 would be correct.

jakethekoenig (commented):
This PR is merged to use OpenAI's types now. Thanks again @PCSwingle. It makes the change a lot tighter and hopefully easier to review. Also the jinja refactor means it's no longer necessary to keep track of local image files for screenshots because we can inline the base64 encoded images.

I left two inline comments that would be nice to address but I don't think they're blocking.

I'm also not sure if we want to write an end to end screenshot test that actually fires up selenium and takes a screenshot. Github action runners don't have chrome by default and I'm worried it would slow down our test suite a lot.

mentat/llm_api.py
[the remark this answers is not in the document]
PCSwingle: Huh, that is weird

mentat/commands.py
```diff
@@ -390,3 +391,53 @@ def argument_names(cls) -> list[str]:
...
+        vision_manager = session_context.vision_manager
+        stream = session_context.stream
+        config = session_context.config
+        conversation = session_context.conversation
+        model = config.model
+
+        if "gpt" in model:
+            if "vision" not in model:
+                stream.send(
+                    "Using a version of gpt-4 that doesn't support images. Changing to"
```
PCSwingle, line 408: Maybe just say version of gpt? Could be using gpt-3.5

jakethekoenig: Running the screenshot command doesn't call the model it just adds a screenshot to the conversation. The model will need to be `gpt-4-vision-preview` when the user next enters a prompt (and on all future prompts unless the run `/clear`). This is part of the reason I think this change of the model is safe: If the user doesn't want to use `gpt-4-vision-preview` they'll see the model is changed and they'll have the opportunity to change it back or quit before it is used.

mentat/llm_api_handler.py
PCSwingle: It could be a list of ToolCalls (in ChatCompletionAssistantMessageParam). I just added a commit that casts value as long as the key is "content".

 I think the +4 is for the im_start and im_end, not for the role, and the role is always one token which bumps it up to 5. I could be wrong though; I'll try curling some messages to openai (without streaming, so they'll give me the exact tokens) and running them through our function and see if I'm wrong.

mentat/llm_api_handler.py
```diff
@@ -173,6 +192,7 @@ async def call_llm_api(
                 messages=messages,
                 temperature=config.temperature,
                 stream=stream,
+                max_tokens=4096,  # gpt-4-vision-preview returns a max of 30 by default.
```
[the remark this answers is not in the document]
jakethekoenig: Correct. I believe setting this parameter should have no effect mostly except for vision models where it overrides the max of 30. e.g. there's no concern the model will see this large value and try to fill up the space.

mentat/transcripts.py
```diff
@@ -3,13 +3,13 @@
...
 
-from openai.types.chat import ChatCompletionMessageParam
+from openai.types.chat import ChatCompletionContentPartParam, ChatCompletionMessageParam
 
 from mentat.logging_config import logs_path
 
 
 class UserMessage(TypedDict):
-    message: str
+    message: list[ChatCompletionContentPartParam]
```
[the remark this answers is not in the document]
jakethekoenig: This actually does change how the prior messages will appear in the text transcript. e.g. I see prior messages like: `{"role": "user", "content": [{"type": "text", "text": "Describe the doodle"}]}` whereas previously the content would have been a string. This isn't breaking with previous transcipts because the jinja template handles both. I'm actually not 100% sure openai treats both the same. While you're curling could you double check that that message and `{"role": "user", "content": "Describe the doodle"}` return the same token count?

mentat/resources/templates/conversation_viewer.jinja
PCSwingle: I'll work on this

PCSwingle:
TODO before merging:
Curl and check that our token counts are correct
Check that token counts are the same for list of content vs str content
/screenshot an html file isn't working for me; I think we need an absolute path and/or triple /// after file.

mentat/vision/vision_manager.py
```diff
@@ -29,7 +29,7 @@ def open(self, path: str) -> None:
 
     def screenshot(self, path: Optional[str] = None) -> str:
         if path is not None:
-            expanded = os.path.expanduser(path)
+            expanded = os.path.abspath(path)
```
jakethekoenig, line 32: Good catch! I think we should have both. As is `/screenshot ~/.mentat/conversation_viewer.html` won't work.

PCSwingle: Oh I didn't realize abspath didn't do that, thanks! I'll do that

mentat/resources/templates/conversation_viewer.jinja
```diff
@@ -1,3 +1,17 @@
+{% macro display_message(message, content_key) %}
```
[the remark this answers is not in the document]
PCSwingle: Hmm, maybe? That is weird

PCSwingle (approved):
Looks great! Thanks so much for adding this!

PR #291, opened 2023-11-16
Auto context UI Updates

granawkins:
Hopefully this can be the last PR before launching auto-context, except for maybe a README update.

Included:
- `/context` will now show self.features (instead of self.included_files) AFTER a query, so you can see what auto-context has selected. This takes the place of printing out 'N full files selected, N intervals..' we had before.
- That readout includes interval numbers
- Costs from LLMFeatureFilter and Embeddings are logged to CostTracker. They aren't displayed at the time, but are logged and will contribute to the total cost displayed at end of session.
- If LLMFeatureFilter doesn't return valid json, retry twice, then just go without.
- Allow `included_files` to be > `maximum_context`. Just warn user on startup and sort/truncate.

Not included, but could if someone feels strongly:
- Editing files not in context. This touches a lot, and intersects with the git stuff, so I think it's best to do separately.
- I have a note about use `ask_yes_no` when paths is empty and we include `diff_context.files`. This means set_paths (where it happens) has to be async, and also intersects a lot with the git, so again, separate.

LMK if there's anything I missed, or other stuff you think we should include.

TODO

- [x] Add `auto-context` arg back, set to 4k tokens default, apply only to auto-generated features


mentat/code_context.py
```diff
@@ -72,26 +74,18 @@ def display_context(self):
         stream = session_context.stream
         config = session_context.config
         git_root = session_context.git_root
+        prefix = "  "
 
+        stream.send("")
```
jakethekoenig, line 79: I personally don't like the newline here.

mentat/code_context.py
```diff
@@ -72,26 +74,18 @@ def display_context(self):
...
         stream.send("Code Context:", color="blue")
```
jakethekoenig, line 80: I notice we use three different shades of blue: blue, light_blue and cyan. I wonder if we should consolidate?
image

mentat/code_context.py
```diff
@@ -100,20 +94,27 @@ def display_context(self):
         else:
             stream.send(f"{prefix}Auto-Context: Disabled")
 
-    def display_features(self):
-        """Display a summary of all active features"""
-        session_context = SESSION_CONTEXT.get()
-        stream = session_context.stream
```
jakethekoenig, line 106: I'm not sure if there's a clean way to do this but it might be nice to somehow have the features displayed in similarity order so you see what the model thinks is most likely to be useful first.

mentat/conversation.py
```diff
@@ -92,18 +92,11 @@ async def display_token_count(self):
             )
         else:
             self.max_tokens = context_size
-        if context_size and tokens > context_size:
```
jakethekoenig, line 95: On line 80 (can't comment inline there) `max_tokens` is set to 0. Is there a reason it isn't set to `context_size`? I think it might not have been updated when we changed the meaning of the field from allowed auto_tokens.

mentat/conversation.py
```diff
@@ -92,18 +92,11 @@ async def display_token_count(self):
...
-        elif tokens + 1000 > context_size:
+        if context_size and tokens + 1000 > context_size:
             stream.send(
-                f"Warning: Included files are close to token limit ({tokens} /"
-                f" {context_size}), you may not be able to have a long"
-                " conversation.",
-                color="red",
+                f"Included files exceeds token limit ({tokens} / {context_size}). "
+                "Truncating based on task similarity.",
+                color="yellow",
```
jakethekoenig, line 99: `display_context()` is called in session before this filtering takes place. I'm not sure we want to print out the full tree if they run `mentat .`. Also we do filtering before a request because we want to know the token count before a request to report to the user. But this is wasteful because ordering by similarity with no prompt isn't meaningful.

mentat/conversation.py
```diff
@@ -92,18 +92,11 @@ async def display_token_count(self):
...
-                " number of files."
-            )
-        elif tokens + 1000 > context_size:
+        if context_size and tokens + 1000 > context_size:
             stream.send(
-                f"Warning: Included files are close to token limit ({tokens} /"
-                f" {context_size}), you may not be able to have a long"
-                " conversation.",
-                color="red",
+                f"Included files exceeds token limit ({tokens} / {context_size}). "
```
jakethekoenig, line 97: Tokens here isn't the token count of included files. It's the token count of what the code_message would be plus the system prompt.

jakethekoenig (commented):
You can take or leave the design notes. This mostly looks good to me but I think we need to fix is so `mentat . --maximum-context 8000` correctly says 100000+/8000 tokens. There are two bugs in the way:
1. `code_context.get_code_message()` is given a max_tokens of 0. in `display_token_count` and `get_code_message` is already doing truncation so it no longer returns the included token count.

mentat/code_context.py
granawkins: I tried to update it so that cyan = context/cost, and light_blue = instructions/edits. Here's a full convo for reference.
image

granawkins (commented):
I rewrote part of Conversation.display_token_count, which resolves several different comments.

jakethekoenig (approved):
Looks good to me. Just the one typo. The error I ran into was unrelated to this change. I made an [issue](https://github.com/AbanteAI/mentat/issues/334) for it. I also checked the unchecked box in the description since I think that is done.

PR #294, opened 2023-11-16
Benchmark runner and simple benchmark added

jakethekoenig:
This represents a break from how I was thinking about benchmarks previously:
* Rather than generating many benchmarks it'll likely be more valuable to have a few that we can understand well.
* If we're going to write or understand them by hand I'd much prefer to have them in python than json.
  - It'll be easy to convert the json tests into this format when we are ready for generated tests.
* Similar to how pytest discovers test files I think our benchmark runner should discover benchmarks in the benchmark directory.
* For simple benchmarks we should have a deterministic test to see if gpt succeeded which is what I've done for this one test.
* When I introduce a more complicated benchmark I'll add GPT grading with a hand made context for the task.
* Part of what makes a task challenging is the ambiguity of the prompt so I wrote multiple versions. The first mostly succeeds. The second two maybe 50-50.

Run with:
```
pytest tests/benchmarks/benchmark_runner.py --benchmark -s --retries 2
```

### DIFF
--- a/tests/benchmarks/benchmark_runner.py
+++ b/tests/benchmarks/benchmark_runner.py
@@ -1,0 +1,94 @@
+import importlib.util
+import os
+
+import pytest
+from git import Repo
+
+from mentat.config import Config
+from mentat.python_client.client import PythonClient
+from tests.benchmarks.utils import clone_repo
+
+
+def dynamic_import(path_to_module, module_name):
+    spec = importlib.util.spec_from_file_location(module_name, path_to_module)
+    module = importlib.util.module_from_spec(spec)
+    spec.loader.exec_module(module)
+    return module
+
+
+pytestmark = pytest.mark.benchmark
+
+
+@pytest.fixture
+def retries(request):
+    return int(request.config.getoption("--retries"))
+
+
+@pytest.mark.asyncio
+async def test_benchmark(retries):
+    print("Running benchmarks")
+    benchmarks_dir = f"{os.path.dirname(__file__)}/benchmarks"
+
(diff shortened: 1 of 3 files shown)


tests/benchmarks/benchmarks/mentat/license_update.py
```diff
@@ -0,0 +1,45 @@
...
+minimum_context = ["tests/license_check.py:11-22"]
+
+
+def verify():
+    try:
+        import tests.benchmarks.repos.mentat.tests.license_check as license_check
+
+        importlib.reload(license_check)
+        return set(license_check.accepted_licenses) == set(
+            [
```
PCSwingle, line 31: To stop this from being flaky, why don't we get the list of licenses first instead of hard coding them here?

tests/benchmarks/benchmark_runner.py
```diff
@@ -0,0 +1,94 @@
...
+            for i in range(1, retries + 1):
+                client = PythonClient(
+                    config=Config(
+                        auto_context=True,
+                        maximum_context=8000,
+                    )
+                )
+
+                await client.startup()
+                await client.call_mentat_auto_accept(prompt)
```
PCSwingle, line 67: I love the python client and these functions so much!

tests/benchmarks/benchmarks/mentat/license_update.py
PCSwingle: Oh, I didn't realize that. Nevermind then!

PCSwingle (approved):

PR #295, opened 2023-11-16
Only 100 conversations collected

jakethekoenig:


PCSwingle:
What's the maximum character limit allowed? If the conversations are big this could still overflow pretty easily; we could just keep going until we reach the limit instead

jakethekoenig:
After the jinja rewrite the browser can open all the logs (2190) I've made so far pretty easily. But it feels noticeably laggy both to open and to scroll around so I think we should still have a limit. Given it's about feel and not a hard performance limit of the browser I think limiting by transcript count instead of character count is fine.

PCSwingle (approved):

PR #297, opened 2023-11-17
Conversation Viewer statically built with Jinja

jakethekoenig:
This should be mostly identical to the previous version except:
* new lines replaced with <br> in messages and pre isn't used.
* If you select a message, leave and come back your selection is remembered.
* The browser doesn't barf when you open 100s of transcripts.

I wanted to make a few small QoL improvements to the viewer and thought it would be best to do this first.


jakethekoenig:
Sort of conflicts with my vision PR but I have that on ice until we update our OpenAI version.

PCSwingle (approved):
Thank you so much for adding this! This makes the conversation viewer 10x better! I haven't tested this but as long as you've tested it and it works I'm fine with merging since this is a dev feature

PR #298, opened 2023-11-18
add auto-tokens argument back

granawkins:
Short PR to add the `auto-tokens` arg back, with a default of 4000, which applies only to additional files added by `auto-context`, as discussed.

This is quite necessary now that `gpt-4-1103` is the default model, so I'm doing a tiny PR to fast-track it.

### DIFF
--- a/mentat/config.py
+++ b/mentat/config.py
@@ -84,7 +84,7 @@ class Config:
         converter=converters.optional(converters.to_bool),
     )
     auto_tokens: int = attr.field(
-        default=4000,
+        default=8000,
         metadata={
             "description": "The number of tokens auto-context can add.",
         },
--- a/mentat/config.py
+++ b/mentat/config.py
@@ -83,6 +83,13 @@ class Config:
         },
         converter=converters.optional(converters.to_bool),
     )
+    auto_tokens: int = attr.field(
+        default=4000,
+        metadata={
+            "description": "The number of tokens auto-context can add.",
+        },
+        converter=int,
+    )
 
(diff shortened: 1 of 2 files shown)


mentat/config.py
```diff
@@ -83,6 +83,13 @@ class Config:
         },
         converter=converters.optional(converters.to_bool),
     )
+    auto_tokens: int = attr.field(
+        default=4000,
```
jakethekoenig, line 87: I think the default should be a little higher. Like 8000. The one benchmark I wrote fails when autocontext is this low but succeeds at 8000. Sorry I wish we had more benchmarks already so we could be more scientific.

mentat/config.py
```diff
@@ -83,6 +83,13 @@ class Config:
...
+        metadata={
+            "description": "The number of tokens auto-context can add.",
```
jakethekoenig, line 89: Maybe rephrase this to "will add" since that's the current behavior roughly

jakethekoenig (approved):
Thanks!

mentat/config.py
granawkins: Only if it's turned on though, that was the 'can'

mentat/config.py
granawkins: ya I don't know either. I'll set it to 8k and lets keep an eye on it.

mentat/config.py
granawkins: On second thought I agree, will is clearer.

PR #301, opened 2023-11-18
Add file to context if it really exits but could not be expanded by glob, which is the case of pages/[slug]/index.tsx or simillar pattern, common in nextjs

rogeriochaves:
I was getting errors when trying to include files like `[that].tsx` (yes, this is allowed)

This commit fixes it

### DIFF
--- a/mentat/include_files.py
+++ b/mentat/include_files.py
@@ -18,6 +18,8 @@ def expand_paths(paths: list[Path]) -> tuple[list[Path], list[str]]:
         new_paths = glob.glob(pathname=str(path), recursive=True)
         if new_paths:
             globbed_paths.update(new_paths)
+        elif Path(path).exists():
+            globbed_paths.add(str(path))
         else:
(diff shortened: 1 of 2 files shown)


jakethekoenig:
Thanks for the fix! I think it has a bug for paths with colons though e.g. `test:[example].tsx` won't work. I think it'd be better to have the check right after the `if new_paths`.

That'll also have a bug if the glob has matches and exists. I'm not sure how to handle that case actually because it seems like the user might want any of the three possibilities. Probably adding the glob files and not the actual file is sensible though.

PCSwingle:
Let's make sure we also make sure to add a test for this (maybe even one with a colon as well)? I think we might be changing this logic relatively soon and it would be good to make sure this can't slip through again. I can handle finishing up this PR after I finish what I'm currently working on if nobody else has done it.

PCSwingle (approved):

PR #303, opened 2023-11-20
update openai, add llm_api_handler, fix tests

PCSwingle:
Update's to OpenAI 1.0 API


mentat/code_edit_feedback.py
```diff
@@ -24,36 +23,47 @@ async def get_user_feedback_on_edits(
     match user_response.lower():
         case "y" | "":
             edits_to_apply = file_edits
-            # TODO: All of the 'user messages' here feel like they should be system messages;
-            # if we decide to change this, make sure to change the clear command as well
             conversation.add_message(
-                MessageRole.User, "User chose to apply all your changes."
+                {"role": "system", "content": "User chose to apply all your changes."}
```
biobootloader, line 27: in these and the changes below we're changing these message roles from `user` to `system`. I believe they were originally `system`, and that feels like it makes more sense, but maybe that should be a separate change from this PR? Since it might impact model understanding, conversation viewer display, etc

mentat/conversation.py
```diff
@@ -2,50 +2,46 @@
...
+from mentat.utils import add_newline
 
 if TYPE_CHECKING:
     from mentat.parsers.file_edit import FileEdit
 
 
-class MessageRole(Enum):
-    System = "system"
-    User = "user"
-    Assistant = "assistant"
```
biobootloader, line 29: why remove this Enum? Seems cleaner to me than just using the strings everywhere

mentat/logging_config.py
```diff
@@ -74,19 +76,22 @@ def setup_logging():
...
             transcript = f.readlines()
         if len(transcript) == 0:
             continue
-        else:
-            transcript = json.loads("[" + ", ".join(transcript) + "]")
-        ans.append((timestamp, transcript))  # type: ignore
+
+        # TODO: Test if this still works (despite not being explicitly converted to ChatCompletionMessageParam)
+        transcript = json.loads("[" + ", ".join(transcript) + "]")
+        ans.append((timestamp, transcript))
```
biobootloader, line 96: TODO

mentat/session.py
```diff
@@ -53,13 +53,39 @@ def __init__(
         sentry_init()
         self.id = uuid4()
 
+        try:
+            # This is so that if we set the session context in a test, we don't reset it;
+            # is there a better way to do this?
```
biobootloader, line 58: @waydegg

tests/conftest.py
```diff
@@ -171,31 +163,88 @@ def set_stream_messages(values):
     return async_mock
 
 
-@pytest.fixture(scope="function")
-def mock_setup_api_key(mocker):
-    mocker.patch("mentat.session.setup_api_key")
-    mocker.patch("mentat.conversation.is_model_available")
-    return
+class MockLlmApiHandler:
```
biobootloader, line 166:
```suggestion
class MockLLMApiHandler:
```

tests/diff_context_test.py
```diff
@@ -161,31 +159,3 @@ async def test_diff_context_pr(temp_testbed, git_history, mock_session_context):
     assert diff_context.name.startswith("Merge-base Branch master:")
     assert diff_context.name.endswith(": commit2")  # NOT the latest
     assert diff_context.files == [abs_path]
-
-
-@pytest.mark.asyncio
-async def test_diff_context_end_to_end(
-    temp_testbed, git_history, mock_call_llm_api, mock_setup_api_key
-):
```
biobootloader, line 169: why is this removed?

mentat/code_edit_feedback.py
PCSwingle: Hmm, ok. I do think they make a lot more sense to be system messages though, but I'll try benchmarking them.

mentat/conversation.py
PCSwingle: I did this because OpenAI's typing uses Literal['user', 'system', 'assistant']; while we can go back to Enums, I think we should stick with OpenAI's typing (we're just going to be role.value everywhere with the Enum, and of course pyright will catch any errors because it is a Literal)

mentat/conversation.py
```diff
@@ -140,46 +138,56 @@ def get_messages(self) -> list[dict[str, str]]:
...
         loading_multiplier: float = 0.0,
     ):
-        start_time = default_timer()
         session_context = SESSION_CONTEXT.get()
         stream = session_context.stream
         config = session_context.config
         parser = config.parser
+        llm_api_handler = session_context.llm_api_handler
+
+        start_time = default_timer()
```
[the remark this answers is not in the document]
PCSwingle: Hmm... I would prefer to keep actual llm logic inside of llm_api_handler, but it could be useful; having the outside have to pass in an object that gets filled out seems awkward though. I think our current method is the simplest; let's just stick with it for now, but I can definitely see wrapping a timer generator around the response working in the future.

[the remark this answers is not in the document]
PCSwingle: The main reason I'm hesitating on adding a wrapped generator is we would just have to await the timer object that gets filled out when the generator is finished, which feels weird (since things could somehow go wrong and then we'd just get stuck waiting for the timer to fill out)

mentat/logging_config.py
PCSwingle: Thanks for pointing this out haha, I completely forgot about this

mentat/session.py
PCSwingle: It's not that I want the session context to persist between different tests; it's that we mock the session context in a fixture, and I don't want Session overwriting the mocked sessioncontext. Thinking about it more, I think the best thing to do would actually be to create a Session in a fixture before we mock sessioncontext and then to use that Session fixture in our tests.

tests/conftest.py
PCSwingle: I don't think you normally capitalize abbreviations in PascalCase; otherwise we'd be capitalizing both LLM and API and we'd have MockLLMAPIHandler. I think it's best this way because it's consistent

requirements.txt
```diff
@@ -1,8 +1,9 @@
+attrs==23.1.0
```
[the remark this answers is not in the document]
PCSwingle: Yeah I think the old openai library must have been requiring it because I deleted my env and installed requirements.txt and it suddenly wasn't installed anymore haha

tests/diff_context_test.py
PCSwingle: It would've been an annoying test to port over, and I didn't think it was adding any coverage; we don't really have any other 'e2e' tests like this one (just client tests, but because the Session keeps the test SessionContext they work pretty easily), just unit tests, and in order to add them I'm going to add a marker to not automock session context, add a fixture to mock llm_api_handler, and change a few things. I'm actually thinking I'm going to change  it so we don't mock SessionContext though, since I realized it's a bit of a relic making tests harder though. I probably won't add this test back though, since like I said it doesn't really add any coverage.

mentat/llm_api_handler.py
```diff
@@ -0,0 +1,214 @@
...
+        "PYTEST_CURRENT_TEST" in os.environ
+        and "--benchmark" not in sys.argv
+        and (not bool(benchmarks_running) or benchmarks_running == "false")
+    )
+
+
+def raise_if_in_test_environment():
+    assert (
+        not is_test_environment()
+    ), "OpenAI call attempted in non benchmark test environment!"
```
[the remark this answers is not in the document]
PCSwingle: Hmm, I'm planning on changing how we do mocks, and I think this is a good idea; it'll make sure we're mocking less (just the stuff that matters; openai)

mentat/llm_api_handler.py
```diff
@@ -0,0 +1,214 @@
...
+
+
+# Ensures that each chunk will have at most one newline character
+def chunk_to_lines(chunk: ChatCompletionChunk) -> list[str]:
+    content = chunk.choices[0].delta.content
+    return ("" if content is None else content).splitlines(keepends=True)
+
+
+def count_tokens(message: str, model: str, full_message: bool = False) -> int:
```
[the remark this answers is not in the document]
PCSwingle: We want the user to be able to set any model they want to though (they can set openai base to automatically change the url we request from); I think I'll leave it as a string because of this.

mentat/llm_api_handler.py
```diff
@@ -0,0 +1,214 @@
...
+            if not isinstance(value, str):
+                continue
+            num_tokens += len(encoding.encode(value))
+            if key == "name":  # if there's a name, the role is omitted
+                num_tokens -= 1  # role is always required and always 1 token
+    num_tokens += 2  # every reply is primed with <im_start>assistant
+    return num_tokens
+
+
+# TODO: This function should just get the prompt token count; it shouldn't be warning at all
```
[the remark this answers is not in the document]
PCSwingle: :+1:

mentat/llm_api_handler.py
```diff
@@ -0,0 +1,214 @@
...
+            return (0.003, 0.004)
+        else:
+            return (0.0015, 0.002)
+    elif "ada-002" in model:
+        return (0.0001, 0)
+    else:
+        return None
+
+
+class LlmApiHandler:
```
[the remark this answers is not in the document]
PCSwingle: @biobootloader suggested the same thing; like I said there, I think convention for PascalCase is not to capitalize abbreviations (otherwise we'd have LLMAPIHandler)

mentat/logging_config.py
PCSwingle: Confirmed works; I expected it to because TypedDicts should be equivalent to regular Dicts at runtime, but wanted to check :+1:

tests/diff_context_test.py
PCSwingle: Re-added the test!

mentat/conversation.py
biobootloader: I see, didn't realize it was Literal['user', 'system', 'assistant'] and pyright would still catch. I'm ok with it then

tests/conftest.py
biobootloader: Hmm yeah MockLLMAPIHandler is bad.

I looked at LangChain to see how they do it and they usually capitalize LLM and API in names: https://github.com/search?q=repo%3Alangchain-ai%2Flangchain+API+LLM&type=code

But yeah when they sequential `LLMAPI` is confusing so I guess leave it

biobootloader (approved):

waydegg (approved):

PR #309, opened 2023-11-27
add run command

PCSwingle:
Adds the run command. Will stream both stdout and stderr, and then send it to the model.

### DIFF
--- a/mentat/commands.py
+++ b/mentat/commands.py
@@ -398,8 +398,8 @@ class RunCommand(Command, command_name="run"):
 
     @classmethod
     def argument_names(cls) -> list[str]:
-        return ["command"]
+        return ["command", "args..."]
 
     @classmethod
     def help_message(cls) -> str:
-        return "Run a shell command and send the output the model's conversation."
+        return "Run a shell command and put its output in context."
--- a/mentat/commands.py
+++ b/mentat/commands.py
@@ -44,8 +44,6 @@ class Command(ABC):
     def get_command_completions(cls) -> List[str]:
         return list(map(lambda name: "/" + name, cls.get_command_names()))
 
-    # Although we don't await anything inside an apply method currently, in the future we might
-    # ask ther user or a model something, which would require apply to be async
     @abstractmethod
     async def apply(self, *args: str) -> None:
         pass
@@ -390,3 +388,18 @@ class ConfigCommand(Command, command_name="config"):
     @classmethod
     def help_message(cls) -> str:
         return "Set a configuration option or omit value to see current value."
+
+
+class RunCommand(Command, command_name="run"):
+    async def apply(self, *args: str) -> None:
+        session_context = SESSION_CONTEXT.get()
+        conversation = session_context.conversation
+        await conversation.run_command(list(args))
+
+    @classmethod
+    def argument_names(cls) -> list[str]:
+        return ["command"]
+
+    @classmethod
+    def help_message(cls) -> str:
+        return "Run a shell command and send the output the model's conversation."
--- a/mentat/conversation.py
+++ b/mentat/conversation.py
@@ -205,7 +205,7 @@ class Conversation:
             )
         response = await llm_api_handler.call_llm_api(
             messages,
-            config.model
+            config.model,
             stream=True,
             response_format=parser.response_format(),
         )
@@ -295,7 +295,7 @@ class Conversation:
         if max_context is None:
             return None
 
-        return max_context - conversation_tokens(self.get_messages(), ctx.config.model)
+        return max_context - prompt_tokens(self.get_messages(), ctx.config.model)
 
     def can_add_to_context(self, message: str) -> bool:
         """
--- a/mentat/conversation.py
+++ b/mentat/conversation.py
@@ -1,7 +1,9 @@
 from __future__ import annotations
 
+import asyncio
 import json
 import logging
+import subprocess
 from timeit import default_timer
 from typing import TYPE_CHECKING
 
@@ -90,7 +92,7 @@ class Conversation:
                 f" {context_size}). Please try running again with a reduced"
                 " number of files."
             )
-        elif tokens + 1000 > context_size:
+        elif tokens + config.token_buffer > context_size:
             stream.send(
                 f"Warning: Included files are close to token limit ({tokens} /"
                 f" {context_size}), you may not be able to have a long"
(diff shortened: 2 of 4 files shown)


mentat/conversation.py
```diff
@@ -269,3 +269,76 @@ async def get_model_response(self) -> list[FileEdit]:
...
+            ctx.stream.send(line, end="")
+            # This gives control back to the asyncio event loop so we can actually print what we sent
+            # Unfortunately asyncio.sleep(0) won't work https://stackoverflow.com/a/74505785
+            # Note: if subprocess doesn't flush, output can't and won't be streamed.
+            await asyncio.sleep(0.01)
+        output = "".join(output)
+        message = (
+            f"The user ran this command:\n{' '.join(command)}\nAnd recieved this"
+            f" output:\n{output}"
+        )
```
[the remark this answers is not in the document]
jakethekoenig: For this pr imo `User ran: ${command}\noutput: ${output}` would perform as well, save tokens and look cleaner in the viewer. Doesn't matter much either way though.
image

mentat/conversation.py
```diff
@@ -269,3 +269,76 @@ async def get_model_response(self) -> list[FileEdit]:
...
+            )
+            ctx.stream.send(
+                "Successfully added command output to model context.", color="cyan"
+            )
+            return True
+        else:
+            ctx.stream.send(
+                "Not enough tokens remaining in model's context to add command output"
+                " to model context.",
+                color="light_red",
```
jakethekoenig, line 342: I don't want to add too many knobs but it might be worth thinking about UX for putting a subset into context? Two ideas:
1. When output exceeds 500 tokens prompt: `Command output was X tokens: (A)dd all/add (F)irst 500 tokens/add (L)ast 500 tokens/add (N)one`.
2. When output exceeds 500 tokens prompt: `Command output was X tokens: (A)dd all to context/(D)on't add to context/Or enter a prompt describing what subset to put into context and a summarized version of the output will be put into context`.

For version one I think this behavior is fine.

mentat/commands.py
```diff
@@ -390,3 +388,18 @@ def argument_names(cls) -> list[str]:
...
+
+class RunCommand(Command, command_name="run"):
+    async def apply(self, *args: str) -> None:
+        session_context = SESSION_CONTEXT.get()
+        conversation = session_context.conversation
+        await conversation.run_command(list(args))
+
+    @classmethod
+    def argument_names(cls) -> list[str]:
+        return ["command"]
```
jakethekoenig, line 401: Should make it clear the user can add arguments as they would on the command line itself.

mentat/conversation.py
```diff
@@ -269,3 +269,76 @@ async def get_model_response(self) -> list[FileEdit]:
...
+        remaining_context = self.remaining_context()
+        return (
+            remaining_context is not None
+            and remaining_context
+            - count_tokens(message, ctx.config.model, full_message=True)
+            - ctx.config.token_buffer
+            > 0
+        )
+
+    async def run_command(self, command: list[str]) -> bool:
```
jakethekoenig, line 297: In my opinion this logic should be in the `RunCommand` and it should call `add_system_message` (which doesn't exist but can be added as part of this pull request.

mentat/commands.py
```diff
@@ -390,3 +388,18 @@ def argument_names(cls) -> list[str]:
...
+
+    @classmethod
+    def help_message(cls) -> str:
+        return "Run a shell command and send the output the model's conversation."
```
jakethekoenig, line 405: I'd prefer "Run a shell command and put its output in context.". (Also note there's a typo)

mentat/commands.py
```diff
@@ -390,3 +388,18 @@ def argument_names(cls) -> list[str]:
     @classmethod
     def help_message(cls) -> str:
         return "Set a configuration option or omit value to see current value."
+
+
```
jakethekoenig, line 392: Something I was thinking about when I was thinking about making a run command is it might be nice to have "put command and its output in llm context" be an option that other commands could use. It's difficult though because commands don't really have output they have `stream.send` statements. And looking through the commands that currently exist none of them make that much sense to put in the llms context (I could imagine the model seeing the user ran `/include file` right before a prompt might change what it focused on?).

I would kind of like to have the commands and their output in `_messages` maybe with some metadata that filters them out before sending to the llm so I could see them with `/conversation`

mentat/commands.py
```diff
@@ -390,3 +388,18 @@ def argument_names(cls) -> list[str]:
...
+class RunCommand(Command, command_name="run"):
```
jakethekoenig, line 393: Another thing about our commands: we should probably split this file up and put each command in its own file. But we should do that refactor in a later pr.

jakethekoenig (commented):
Thanks for doing this! My temperature must have been high when reviewing this PR because I had a lot of feature ideas. The only things I think should actually be changed before merging are:
1. Fix the help message.
2. Put the logic that runs the command in `RunCommand` itself.

mentat/conversation.py
PCSwingle: Hmm, maybe but I think this is adding too much complexity for not much gain. We can think about it later either way

mentat/conversation.py
PCSwingle: Hmm, well we're going to want this logic in the future for other things (like agent mode), so I don't think it makes sense to be command specific. The reason we don't have an add_system_message is because the add_model_message and add_user_message are specifically for messages sent by the user/model so that the transcript viewer (on the left) will look like the user expects it to; we could add an add_system_message function, but I think it would obscure the purpose of the model and user message functions (sometimes we still add model and user messages with the add_message function if the user or model didn't actually send them, so we don't want them to appear on the left of the transcript viewer)

mentat/commands.py
PCSwingle: Agreed. I'm planning on doing a major refactor soon that breaks up our main folder into a lot of subfolders.

mentat/conversation.py
jakethekoenig: I think it's more important than it seems because perhaps the most useful thing to do with this is run tests which can often be quite verbose on failures especially if there are many and it should be easy for the user to specify the range with relevant information. But agree we can handle it later.

mentat/conversation.py
jakethekoenig: That makes sense to me. Thanks.

jakethekoenig (approved):

omri123:
As a curious outsider browsing your repo, can I ask why you chose the "system" role for this message?

biobootloader:
> As a curious outsider browsing your repo, can I ask why you chose the "system" role for this message?

We've been using the system role for automatic things - basically anything other than direct user messages in the conversation

omri123:
I have got the impression that system role get higher attention and is used to influence the bot behavior. openai recommend using it for general chat instructions. I didn't check, but I guess that after a few long and informative system messages, the bot might forget its identity from the first system message. Does my guess match your experience?

Since the user chose to run the terminal command, you can phrase the terminal output as a user message.

biobootloader:
>I didn't check, but I guess that after a few long and informative system messages, the bot might forget its identity from the first system message. Does my guess match your experience?

@omri123 we use system messages for several things and it doesn't seem to forget the required response format (which is in the first system message)

PR #313, opened 2023-11-27
add ctrl z pop up /undo hotkey

PCSwingle:
Adds a hotkey to press ctrl-z and have /undo pop up.

### DIFF
--- a/mentat/terminal/prompt_session.py
+++ b/mentat/terminal/prompt_session.py
@@ -93,3 +93,7 @@ class MentatPromptSession(PromptSession[str]):
                 event.current_buffer.reset()
             else:
                 event.app.exit(result="q")
+
+        @self.bindings.add("c-z")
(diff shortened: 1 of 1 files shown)


jakethekoenig:
In my opinion this is a bit weird. Users might expect ctrl-z to suspend the application not undo. I also think it's weird to fill in the command instead of calling undo directly. If we do do it this way maybe we should run `event.current_buffer.validate_and_handle()` after adding `/undo`? Another thing that's sort of weird is if they've already typed something it is replaced with `/undo`. They might have expected it to undo their typing instead of insert the command. Just some thoughts I don't feel that strongly about it. If other people like it this way that's fine too.

PCSwingle:
Ah, good point; I completely forgot about ctrl-z (even though I always use it haha). Right now ctrl-z doesn't actually suspend (something with prompt toolkit I think); I'll add an issue and close this PR.

PR #319, opened 2023-11-28
Keybindings disabled while searching

jakethekoenig:
Currently if a user uses one of our keybindings after entering reverse search with <c-r> they'll have counterintuitive behavior.


PCSwingle (approved):
Thanks for catching this!

PR #320, opened 2023-11-28
refactor commands

PCSwingle:
Refactors each command into a separate file. I don't like that we have to import each command in command/commands/__init__.py; feel free to suggest any better ideas.


jakethekoenig (approved):
One idea to avoid needing to import is to change the strange pattern where the commands register themselves when they're defined. We could do something more similar to what we do for parsers:
* Have an abstract command class with `apply`/`argument_names`/`help_message`. (like we do now without the classmethods)
* Have a dictionary mapping names to commands.

Approving because I think this is directionaly an improvement.

PR #322, opened 2023-11-28
Whisper added

jakethekoenig:
Enter `<c-u>` to enter transcription mode. Feels kind of cool actually. It's sort of far from being mergeable though:
* [x] pyaudio is sort of a heavy dependency to add as it requires platform specific [dependencies](https://pypi.org/project/PyAudio/). I don't want to require users to run `brew install portaudio` before running `pip install mentat`. One option could be making it an optional dependency using `extras_require` and try/catching the pyaudio import.
~* [ ] Cost tracking for whisper~ No longer using api
* [x] The feel of the transcription. It could be a lot snappier. I think it's better to transcribe with the already transcribed portion passed in so it doesn't bounce around as it is re-transcribed slightly differently. Also keeps costs down. See point 2 [here](https://platform.openai.com/docs/guides/speech-to-text/prompting)
* [x] Should probably separate the audio logic and the whisper logic. Someone suggested they wanted this feature accessible from the python_client as well but I'm not sure exactly how that should work or what it would be useful for.
* [x] tests


biobootloader:
any idea why I'm getting this? 

image

jakethekoenig:
> any idea why I'm getting this?
> 
> image

Oh, `input_device_index=3` probably shouldn't be hardcoded. I'll look into having it detect the system default.

waydegg:
Error I get when trying `pip install -e .` on macos:

image

Fixed by installing `portaudio` with `brew`

https://stackoverflow.com/a/33821084

waydegg:
On another note its pretty cool to be talking to my terminal haha

jakethekoenig:
@waydegg I've found them pretty accurate. I have a high quality mic though and I'm trying to talk clearly. I think ideally we'd run the medium or large model but with my amd graphics card I've had trouble getting the model to actually run on the GPU so only the tiny model feels responsive enough. Model size could definitely be in the config (though that leads to another problem that the `/config` command can't easily change client side settings). We could also support the openai api as well which serves the large model. I've found it less well documented than the python libraries though. And it costs money.

And yeah I'm aware of the hoops. There's a different hoop for every os. I was thinking we could have pyaudio as an optional requirement. GPT tells me we can put this in setup:
```
extras_require={
      'voice':  ['pyaudio'],
 }
```
And then users can install voice `pip install .[voice]`. Of course for the actual brew package we could simply include portaudio.

I'm going to look into using sounddevice first though. I think it might be a better library than pyaudio in a few ways and there documentation suggests you don't need anything special to install it. Can you confirm `pip install sounddevice` works on macos without extras?

Another question I had for you: pyright is complaing faster_whisper has no stub files. Are you okay globally ignoring it in the pyright config?

One final question: Can you replace tiny with Large-v2 and report the performance on your m3? I'm curious.

waydegg:
Yeah it might just be my mic (using the built-in mic). If I raise my voice the transcriptions get better lol.

>Another question I had for you: pyright is complaing faster_whisper has no stub files. Are you okay globally ignoring it in the pyright config?

Yup!

>And then users can install voice pip install .[voice]. Of course for the actual brew package we could simply include portaudio.

Nice.

>Can you confirm pip install sounddevice works on macos without extras?

I can install it without any extras (uninstalled `portaudio` before installing `sounddevice` and it worked). Haven't tried actually using `sounddevice` and running it tho

>One final question: Can you replace tiny with Large-v2 and report the performance on your m3? I'm curious.

Yup will try now

waydegg:
So I can run the Large-V3 model, but when we get the results in mentat I'm only getting the first few words of whatever it is I'm saying

waydegg:
>Huh, that's weird. Maybe it's the model being slow? If you wait a bit you don't get more of the transcript?

If I wait for the model to process each word and then I say the next one it works haha

jakethekoenig:
Alright, in my opinion this is ready for review now. The following things would be nice to have but I don't think necessary for merge or the most valuable thing to work on. I'll make an issue for what we merge without:
* [ ] Stop word for whisper. E.g. if the user says "end/stop/period" then whisper ends itself. Ideally user configurable.
* [ ] User configurable model selection
* [ ] User ability to choose openai api instead of local model
* [ ] Post transcription send to gpt for edits. Can get fancy here and try to put give helping context. Like if the user says "pyaudio" whisper will probably guess "pie audio" but if we put an included file in context then gpt may be able to edit it to pyaudio. Just putting all the unique tokens represented in the code context might help a lot and not cost a ton? The user should be able to turn this feature off because it will add cost and latency.
* [ ] I think the current "frozen_timestamp" algorithm works well enough and is pretty simple. But there are a lot of things to try around detecting silence and splitting there and adding the previous transcript as prompt.

biobootloader:
I'm getting pretty poor transcriptions on the default `tiny` model (On M2 macbook air) Trying `base` now and it's still pretty bad, it seems to be overwriting transcriptions? I'm not sure but there may be a bug.

Feels like this will be a really nice feature to have!

Could we display the audio device being used?

tests/clients/voice_test.py
```diff
@@ -0,0 +1,71 @@
...
+    WHISPER_BATCH_SECS,
+    Transcriber,
+)
+
+
+def test_transcriber_with_frozen_timestamp(mocker):
+    try:
+        mocker.patch("mentat.terminal.voice.transcriber.sd")
+    except AttributeError:
+        return  # Test skipped on Ubuntu
```
PCSwingle, line 19: Could we just install portaudio on ubuntu before running pytest?

mentat/terminal/prompt_session.py
```diff
@@ -57,6 +59,23 @@ def prompt_continuation(
             "" if is_soft_wrap else [("class:continuation", " " * (width - 2) + "> ")]
         )
 
+    def _toggle_audio(self):
```
PCSwingle, line 62: Is there a specific reason to make Whisper a terminal only feature? While it does make more sense for a client to be handling the input, we could always have the server run whisper and have a channel that the server can use to send input (rather than output) to the client. That way we could run whisper on any client, although it obviously is a bit weird having the server type stuff in for the user haha.

jakethekoenig: When I started this I thought it made more sense to do it at the input layer and therefore on the client side. I couldn't think of a natural way to expose it in the session for any client to use.

But all my future enhancement ideas would be a lot easier to do on the server side so maybe you're right that I should switch it.

jakethekoenig:
After discussion yesterday I decided to make two major changes:
* We now use the API to transcribe. This turns out to be so fast there's no reason to make sequential calls and pre transcribe the conversation as the user is talking.
* We transcribe on the server side accessible through the `/talk` command
Also recordings are now logged to `~/.mentat/logs/audio` which should hopefully make debugging easier. Luckily the net effect is that the PR is a lot simpler. I feel like I should write some kind of test before merging but I'm not exactly sure what I want to test because it's basically just a wiring up of sounddevice/soundfile to the openai api.

mentat/session_input.py
```diff
@@ -12,8 +12,10 @@
 async def _get_input_request(**kwargs: Any) -> StreamMessage:
     session_context = SESSION_CONTEXT.get()
     stream = session_context.stream
+    default_prompt = session_context.conversation.default_prompt
+    session_context.conversation.default_prompt = ""
```
jakethekoenig, line 16: This is sort of a hacky way to pass the information to the front end. I like it passing as the data field but I don't like temporarily storing it in conversation and accessing it here. One alternative would be to have commands return an `Optional[str]` and if they do pass it into `_get_input_request` [here](https://github.com/AbanteAI/mentat/blob/1f21444d9b35caa0a576ad02c38a1f58350afb27/mentat/session_input.py#L47). Btw reading this code I wasn't entirely sure why commands were intercepted there instead of one level up [here](https://github.com/AbanteAI/mentat/blob/1f21444d9b35caa0a576ad02c38a1f58350afb27/mentat/session.py#L118). It's sort of the same thing either way though.

PCSwingle: I actually do somewhat separate how commands are handled in my agent PR; I made a separate function specifically for intercepting commands

mentat/llm_api_handler.py
```diff
@@ -238,3 +240,13 @@ async def is_model_available(self, model: str) -> bool:
             model.id async for model in self.async_client.models.list()
         ]
         return model in available_models
+
+    async def call_whisper_api(self, audio_path: Path) -> str:
+        raise_if_in_test_environment()
+
+        audio_file = open(audio_path, "rb")
+        transcript = self.client.audio.transcriptions.create(
```
PCSwingle, line 248: Is there a reason why we shouldn't asynchronously send the user's voice instead of using the synchronous client?

mentat/cost_tracker.py
```diff
@@ -44,6 +44,9 @@ def log_api_call_stats(
         costs_logger = logging.getLogger("costs")
         costs_logger.info(speed_and_cost_string)
 
+    def log_whisper_call_stats(self, seconds: float):
```
PCSwingle, line 47: Hmm, this makes me think we should make a separate function for logging vs displaying; definitely not relevant to this PR though!

mentat/session_input.py
PCSwingle: I was actually thinking something completely different when it comes to passing up the input from whisper; I was thinking we could just send the input on a completely different channel (so we wouldn't have to touch input_request at all), and the client could have another task listening on that channel that it would use to just add whatever comes in on that channel to prompt_toolkit's buffer. What do you think about that idea? I think it would be cleaner than this and more adaptable for other use cases

jakethekoenig: A different channel and the client storing it makes sense to me. We can't simply add it to the buffer though because it doesn't exist while the command is running. It's made when the input request signal is sent

jakethekoenig: I made a new stream `default_prompt`. I can imagine using it for other things though all the ideas I've come up with so far seem sort of contrived. (Maybe if the user runs `/commit` with no argument gpt could write the commit message and then the user could see `/commit $WHAT_GPT_WROTE` and have a chance to edit before actually commiting?)

mentat/cost_tracker.py
jakethekoenig: I separated it out because the other one had token counting as a central part of it's logic. If it was a little more expensive I'd want to display after transcription but at less than a penny for a whole minute it doesn't seem worth.

mentat/terminal/client.py
```diff
@@ -57,6 +57,11 @@ async def _cprint_session_stream(self):
         async for message in self.session.stream.listen():
             print_stream_message(message)
 
+    async def _default_prompt_stream(self):
```
PCSwingle, line 60: I like this a lot more; thanks!

PCSwingle:
Just tested it out for the first time and holy cow it's really good and cool! I know this had already taken a while to merge in, but I had one more idea; would it be difficult to stream openai's output? I think it could be pretty quick to add and would really level up the experience for me.

PR #323, opened 2023-11-28
prompt toolkit enable suspend

### LINKED ISSUE #315: Ctrl-z suspend
Suspend Mentat on ctrl-z; right now prompt toolkit swallows the ctrl-z. Maybe look into use_system_bindings with prompt toolkit and make sure it doesn't add any bindings we don't want or mess up ctrl-c?

PCSwingle:
Fixes #315

### DIFF
--- a/mentat/streaming_printer.py
+++ b/mentat/streaming_printer.py
@@ -44,7 +44,11 @@ class StreamingPrinter:
 
         while not self.shutdown:
             if self.strings_to_print:
-                next_string = self.strings_to_print.popleft()
+                if len(self.strings_to_print) > 500:
+                    next_string = "".join(self.strings_to_print)
+                    self.strings_to_print = deque[str]([])
+                else:
+                    next_string = self.strings_to_print.popleft()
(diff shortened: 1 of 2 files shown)


jakethekoenig (commented):
Something sort of strange is if you suspend while gpt-4 is running when you unsuspend it streams out the output when it'd probably be better to be instant (presumably the user suspended to come back to the completed job). I think we should change `streaming_printer` to print everything out without sleeping when finishing is true and it's sufficiently far behind. (Say 500 strings because then it would take a whole second to wrap up)

PCSwingle:
> Something sort of strange is if you suspend while gpt-4 is running when you unsuspend it streams out the output when it'd probably be better to be instant (presumably the user suspended to come back to the completed job). I think we should change `streaming_printer` to print everything out without sleeping when finishing is true and it's sufficiently far behind. (Say 500 strings because then it would take a whole second to wrap up)

Hmm, good point.

jakethekoenig (approved):
Works like a charm. Thanks!

PR #325, opened 2023-11-28
Model sizes included in dict

jakethekoenig:
Importantly vision model added. Less importantly legacy models have their context size explicitly given.

### DIFF
--- a/mentat/llm_api_handler.py
+++ b/mentat/llm_api_handler.py
@@ -104,24 +104,24 @@ def prompt_tokens(messages: list[ChatCompletionMessageParam], model: str):
     return num_tokens
 
 
-# TODO: These two functions should be a dictionary
 def model_context_size(model: str) -> Optional[int]:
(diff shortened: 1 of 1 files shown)


jakethekoenig:
I guess this also fixes a bug in the gpt-3.5 context sizes which were listed as 4097 but are actually 4096. I listed the models in the same order as this [doc](https://platform.openai.com/docs/models/gpt-3-5)

PCSwingle:
Thanks so much for doing this! Do you think you could also do the same for the function below this one (the price function)?

jakethekoenig:
Legacy gpt-3.5 prices aren't listed so I'm not 100% sure they're right. But there's no reason to use them anymore so it seems fine if they're slightly off.

PCSwingle (approved):
Thanks for doing this! Ideally there would be some way for us to fetch the context length and price from openai's api, but we can this is great.

PR #326, opened 2023-11-28
fix gpt-4-vision-preview

PCSwingle:
Fixes the token bug that would cause smaller models (like gpt-3.5.turbo) to return an error as well as fixing the response format bug with gpt-4-vision-preview.


jakethekoenig (approved):

jakethekoenig:
I know we're going to make some user who puts 126k tokens into context and gets an error mad. But it's good enough.

biobootloader:
think they've fixed this yet?

Issue #316, opened 2023-11-27
Mentat says "Changes applied." when changes not applied due to a file being changed by user


biobootloader:
![image](https://github.com/AbanteAI/mentat/assets/128252497/6eb681e2-2fe9-4953-8b45-0a6c25b026aa)

Issue #293, opened 2023-11-16
Lines duplicated where intervals meet


jakethekoenig:
See screenshot. I think there's likely an off by one error where something that should take the last line as exclusive takes it as inclusive.
![image](https://github.com/AbanteAI/mentat/assets/385534/fdf1ec04-eea3-450e-b2f0-7d5ea09faa36)

biobootloader:
@granawkins

PR #328, opened 2023-11-29
Auto context issues

granawkins:
- #316 Calculate and stream `{n_applied}/{n_total} changes applied.` instead.
- #293 Make intervals inclusive-start, exclusive-end.*

\*This changes the interface (you'll see the test) so you need to specify e.g. `myfile.py:1-2` in order to include just line 1. We already do some preprocessing of these command-line args (splitting intervals into different features), so we could add a transformation to keep the cli-arg end-inclusive like they were. Thoughts?

### DIFF
--- a/mentat/code_edit_feedback.py
+++ b/mentat/code_edit_feedback.py
@@ -22,6 +22,7 @@ async def get_user_feedback_on_edits(
     user_response = user_response_message.data
 
     need_user_request = True
+    total_changes = sum([len(file_edit.replacements) for file_edit in file_edits])
     match user_response.lower():
         case "y" | "":
             edits_to_apply = file_edits
@@ -70,8 +71,15 @@ async def get_user_feedback_on_edits(
         file_edit.resolve_conflicts()
 
     if edits_to_apply:
-        await code_file_manager.write_changes_to_files(edits_to_apply, code_context)
-        stream.send("Changes applied.", color="light_blue")
+        applied_edits = await code_file_manager.write_changes_to_files(
+            edits_to_apply, code_context
+        )
+        applied_changes = sum(
+            [len(file_edit.replacements) or 1 for file_edit in applied_edits]
+        )
+        stream.send(
+            f"{applied_changes}/{total_changes} changes applied.", color="light_blue"
+        )
     else:
         stream.send("No changes applied.", color="light_blue")
 
--- a/mentat/code_edit_feedback.py
+++ b/mentat/code_edit_feedback.py
@@ -22,7 +22,6 @@ async def get_user_feedback_on_edits(
     user_response = user_response_message.data
 
     need_user_request = True
-    total_changes = sum([len(file_edit.replacements) or 1 for file_edit in file_edits])
     match user_response.lower():
         case "y" | "":
             edits_to_apply = file_edits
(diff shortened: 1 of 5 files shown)


mentat/code_edit_feedback.py
```diff
@@ -22,6 +22,7 @@ async def get_user_feedback_on_edits(
     user_response = user_response_message.data
 
     need_user_request = True
+    total_changes = sum([len(file_edit.replacements) or 1 for file_edit in file_edits])
```
PCSwingle, line 25: I'm not sure how much sense this makes; some formats don't really have set replacements (like unified diff format), not to mention this doesn't take into account creation/deletion/renaming changes. I think we should just keep the regular Changes applied. text and only print it if we do end up applying an edit.

PCSwingle:
> * [Mentat says "Changes applied." when changes not applied due to a file being changed by user #316](https://github.com/AbanteAI/mentat/issues/316) Calculate and stream `{n_applied}/{n_total} changes applied.` instead.
> * [Lines duplicated where intervals meet #293](https://github.com/AbanteAI/mentat/issues/293) Make intervals inclusive-start, exclusive-end.*
> 
> *This changes the interface (you'll see the test) so you need to specify e.g. `myfile.py:1-2` in order to include just line 1. We already do some preprocessing of these command-line args (splitting intervals into different features), so we could add a transformation to keep the cli-arg end-inclusive like they were. Thoughts?

I think it makes sense for intervals to exclusive end; additionally, it's not like the interval interface is super cli friendly to begin with. It's mainly just there for the auto context. In light of that, I think we should just keep it simple and have the cli intervals be the same as the regular intervals (like they are now), and maybe add some documentation (if we even have documentation about intervals already).

jakethekoenig:
> > * [Mentat says "Changes applied." when changes not applied due to a file being changed by user #316](https://github.com/AbanteAI/mentat/issues/316) Calculate and stream `{n_applied}/{n_total} changes applied.` instead.
> > * [Lines duplicated where intervals meet #293](https://github.com/AbanteAI/mentat/issues/293) Make intervals inclusive-start, exclusive-end.*
> > 
> > *This changes the interface (you'll see the test) so you need to specify e.g. `myfile.py:1-2` in order to include just line 1. We already do some preprocessing of these command-line args (splitting intervals into different features), so we could add a transformation to keep the cli-arg end-inclusive like they were. Thoughts?
> 
> I think it makes sense for intervals to exclusive end; additionally, it's not like the interval interface is super cli friendly to begin with. It's mainly just there for the auto context. In light of that, I think we should just keep it simple and have the cli intervals be the same as the regular intervals (like they are now), and maybe add some documentation (if we even have documentation about intervals already).

Maybe I'm weird but I actually use the intervals all the time for adding large files where I only think a bit of it needs to be in context.

jakethekoenig:
I don't want to overcomplicate this PR but to me it's sort of weird that the only way to initialize a Code feature as an interval is with a string encoding the intervals and as a result we sometimes create a string to pass in. Might be better to make another constructor and remove that. And then we could change parse_interval to add 1 to the end and keep inclusive there. Which I think makes more sense for the user facing feature.

jakethekoenig (approved):

PCSwingle (approved):
Looks good to me! I agree with @jakethekoenig that it would be good to have intervals in the constructor, but definitely not needed for this PR.

PR #335, opened 2023-11-30
Remove CodeMap

biobootloader:


biobootloader:
@granawkins can you take a look at the changes to `split_file_into_intervals` in `mentat/code_feature.py`? There are a couple differences in behavior - names are now joined when intervals are merged, and if the last interval is less than the minimum it gets merged.

granawkins (approved):
Looks good to me!

PR #336, opened 2023-11-30
Initialize client not mocked in benchmarks

jakethekoenig:
If we're in a benchmark we need to initialize the client. While I was there I fixed a typo.

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -237,8 +237,9 @@ def mock_model_available(mocker):
 
 
 @pytest.fixture(autouse=True, scope="function")
-def mock_initizalize_client(mocker):
-    mocker.patch.object(LlmApiHandler, "initizalize_client")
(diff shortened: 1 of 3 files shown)


tests/conftest.py
```diff
@@ -237,8 +237,9 @@ def mock_model_available(mocker):
 
 
 @pytest.fixture(autouse=True, scope="function")
-def mock_initizalize_client(mocker):
-    mocker.patch.object(LlmApiHandler, "initizalize_client")
+def mock_initizalize_client(mocker, request):
```
PCSwingle, line 240: Typo is still here :upside_down_face:

PCSwingle (approved):
Feel free to merge after typo is fixed

tests/conftest.py
jakethekoenig: I've really got to make a mentat github PR reviewer.

PR #343, opened 2023-12-01
add basic agent mode

PCSwingle:
Add basic agent mode
TODO:

- [x] Allow easily undoing everything agent did
- [x] Improve prompt (especially command selection; it often chooses commands that try to use some non-existent test file, or it doesn't run commands that actually test its code)
- [x] Faster/better way to control control to user; ctrl-c should automatically go back to user.


mentat/agent_handler.py
```diff
@@ -0,0 +1,155 @@
...
+        )
+        ctx.conversation.add_transcript_message(
+            ModelMessage(message=content, prior_messages=messages)
+        )
+
+    async def _determine_commands(self):
+        ctx = SESSION_CONTEXT.get()
+
+        model = ctx.config.model
+        # TODO: Should we give prompt/tokens?
```
PCSwingle, line 92: @granawkins Should we make get_code_message prompt and tokens arguments be optional? In cases like this it doesn't make much sense to send them. The prompt is just used for embeddings and the tokens for auto context, right?

mentat/agent_handler.py
```diff
@@ -0,0 +1,155 @@
...
+        ctx.cost_tracker.log_api_call_stats(
+            prompt_tokens(messages, model),
+            count_tokens(content, model, full_message=False),
+            model,
+        )
+
+        messages.append(
+            ChatCompletionAssistantMessageParam(role="assistant", content=content)
+        )
+        ctx.conversation.add_transcript_message(
```
PCSwingle, line 84: I think we should change the color of this message (since it's a different conversation and only really related to agent mode)

mentat/agent_handler.py
```diff
@@ -0,0 +1,155 @@
...
+        return self._agent_enabled
+
+    def disable_agent_mode(self):
+        self._agent_enabled = False
+
+    async def enable_agent_mode(self):
+        ctx = SESSION_CONTEXT.get()
+
+        self._agent_enabled = True
+        ctx.stream.send(
```
PCSwingle, line 42: A loading circle here would be nice haha

mentat/session.py
```diff
@@ -113,21 +119,51 @@ async def _main(self):
...
+                        file_edits
+                    )
+                    stream.send(
+                        "Changes applied." if applied_edits else "No changes applied.",
+                        color="light_blue",
+                    )
+
+                    if agent_handler.agent_enabled:
+                        if parsed_llm_response.interrupted:
+                            need_user_request = True
```
PCSwingle, line 164: If the user ctrl-c's out, we give them back control. Right now, we still automatically apply changes; I think this makes sense because we're still in 'agent mode', and the user can just /undo, but I can also definitely see wanting it not to apply them (after all, the user interrupted this response). What do y'all think?

granawkins: I think the way you've done it is good. The user already had the chance to interrupt those edits and didn't, and we tell them about `/undo`.

granawkins (approved):
I just kicked the tires a bit. Look good and performs well!

mentat/agent_handler.py
PCSwingle: Well we don't know how long it's going to take, so I'm not sure it's too applicable

mentat/agent_handler.py
```diff
@@ -0,0 +1,160 @@
...
+        messages = ctx.conversation.get_messages() + [
+            ChatCompletionSystemMessageParam(
+                role="system", content=self.agent_command_prompt
+            ),
+            ChatCompletionSystemMessageParam(role="system", content=code_message),
+            ChatCompletionSystemMessageParam(
+                role="system", content=self.agent_file_message
+            ),
+        ]
+        # TODO: Should this even be a separate call or should we collect commands in the edit call?
```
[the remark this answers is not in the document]
PCSwingle: I don't think so because we need a preset list of functions we want to call; the whole point of this call is that we don't know what commands we can run and the llm is figuring it out.

PR #344, opened 2023-12-01
add pre tags so whitespace isnt collapsed

### LINKED ISSUE #342: Whitespace in conversation viewer
Whitespace is collapsed by default in HTML; we can set white-space: pre to stop this, but then a bunch of random newlines and tabs show up in all of our divs. I suspect this has something to do with jinja (https://stackoverflow.com/questions/35775207/remove-unnecessary-whitespace-from-jinja-rendered-template).

PCSwingle:
Fixed #342. Use \<pre\> tags to display text so whitespace appears normally; this always means we don't need to replace newlines with \<br\> tags.

### DIFF
--- a/mentat/resources/templates/conversation_viewer.css
+++ b/mentat/resources/templates/conversation_viewer.css
@@ -17,6 +17,7 @@ h1 {
 
 pre {
     margin: 0;
+    white-space: pre-wrap;
 }
 
 .container {
--- a/mentat/resources/templates/conversation_viewer.css
+++ b/mentat/resources/templates/conversation_viewer.css
@@ -15,6 +15,10 @@ h1 {
     margin: 10 0px;
 }
 
+pre {
+    margin: 0;
+}
+
 .container {
     position: relative;
     width: 100%;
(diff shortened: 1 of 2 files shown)


PCSwingle:
Oh hmm, I do want it to still wrap. I think there is a way to make it still wrap (as long as there's a breaking character); let me fix that. I do think we definitely want this though, since seeing the whitespace is important (otherwise it could be easy to miss a bug relating to whitespace)

granawkins:
I'm getting these errors from the conversation_viewer when I click on a message to see context. It only happens on this branch but I can't see how it's related. Anyone else get this?

image

PCSwingle:
@granawkins could you send me the exact transcript that happens on? I'm not getting this error

jakethekoenig (approved):
Thanks for fixing the wrap! Looks good to me. I can't imagine Grant's error is related though we should get to the bottom of it.

granawkins:
I think the problem is related to the size of the diff - I was using a massive one, but a smaller one works fine. Good to go!

PR #345, opened 2023-12-03
Bugfix: completions don't work when Mentat run in subdirectory

### LINKED ISSUE #317: autocomplete of class/function names seems to be broken?
It works in one project but not another... will try to reproduce more cleanly

biobootloader:
When you ran in a subdirectory, completions (for function/class names) didn't pop up. I found this was due to using paths relative to the git root. When the mis-pathed files couldn't be opened we were calling `logging.debug(f"Skipping {file_path}. Reason: file not found")`

Fixes #317


biobootloader:
@waydegg I'm merging now but just want to make sure you're aware as this probably changes in your git root PRs, don't want to bug to be reintroduced

PR #346, opened 2023-12-04
Redo command

granawkins:


mentat/edit_history.py
```diff
@@ -1,71 +1,88 @@
...
 @attr.define()
 class RenameAction:
     old_file_name: Path = attr.field()
     cur_file_name: Path = attr.field()
 
-    def undo(self) -> str | None:
+    def undo(self) -> FileEdit:
         if self.old_file_name.exists():
-            return colored(
+            raise MentatError(
```
PCSwingle, line 21: We don't want to end the session when an undo fails, which is what throwing a MentatError will do

PCSwingle, line 21: I would prefer to use a new error (rather than MentatError); maybe we could call in UndoError?

mentat/edit_history.py
```diff
@@ -85,21 +103,37 @@ def push_edits(self):
...
+        if not self.undone_edits:
+            return colored("No edits available to redo", color="light_red")
+
+        session_context = SESSION_CONTEXT.get()
+        code_file_manager = session_context.code_file_manager
+        code_context = session_context.code_context
+
+        edits_to_redo = self.undone_edits.pop()
+        edits_to_redo.reverse()
+        await code_file_manager.write_changes_to_files(edits_to_redo, code_context)
```
PCSwingle, line 135: Hmm, this will check to see if the files are in context and won't apply the edit if they aren't. This is probably a pretty minor thing, but if somebody did and undo and then excluded files and then a redo it wouldn't apply. I think this is fine though because it's a pretty contrived series of events and it would only lead to model work being wasted (as opposed to messing up their files somehow).

PCSwingle: It's also going to play out pretty weird if somebody does /undo, then makes changes, and then tries to redo; I think we might want to clear the redo list after changes are made.

mentat/edit_history.py
```diff
@@ -1,71 +1,88 @@
...
-                f"File {self.cur_file_name} does not exist; unable to undo edit",
-                color="light_red",
+            raise MentatError(
+                f"File {self.cur_file_name} does not exist; unable to undo edit"
             )
         else:
+            new_file_lines = self.cur_file_name.read_text().split("\n")
             with open(self.cur_file_name, "w") as f:
                 f.write("\n".join(self.old_file_lines))
+            return FileEdit(
```
PCSwingle, line 76: It is a bit weird that we turn a fileedit into a historyaction and then back again; maybe there's a better way to do it, but I think this is fine as long as undo and redo are just for a couple of commands. Just commenting so we can think about it in case we want to improve it later.

mentat/edit_history.py
biobootloader: why not use MentatError? I'm fine with us adding new errors but the main reason in my mind would be if they need to be caught at different places

PCSwingle: We do catch this error right here in this function; mentaterror is automatically caught outside the main session loop (used to exit with an error message to the user but without a traceback). I would prefer we use a different error to make it clear that we're catching it here in edit_history.

mentat/edit_history.py
granawkins: I think write_changes_to_files will display the proper messages in these scenarios.

If delete then undo/redo, file_edit.is_valid() is never called, so it doesn't actually check if the file is in context. When I've tried, it just works.

If they make a change, then we compare with `stored_lines` and tell the user to revert.

mentat/edit_history.py
granawkins: Agreed. I we get rid of the line numbers when switching from fileedit to historyaction, which is counter-productive. I think I'll need to address this for the `sample-command` PR, so I'll make a note to take care of it there.

mentat/edit_history.py
PCSwingle: Hmm, ok. I'm ok with this for now but let's keep it in mind in case we ever want to change it.

PCSwingle (approved):
Looks great!

PR #347, opened 2023-12-04
Remove git root from parsers

waydegg:
*This PR is part of a series of PRs to remove git_root from Mentat, and allow Sessions to be run anywhere on the filesystem.*


tests/code_file_manager_test.py
```diff
@@ -99,7 +99,9 @@ async def test_run_from_subdirectory(
         # Hello
         @@end""")])
 
-    session = Session(cwd=Path.cwd(), paths=[Path("calculator.py"), Path("../scripts")])
+    session = Session(
+        cwd=temp_testbed, paths=["multifile_calculator/calculator.py", "scripts"]
```
waydegg, line 103: `paths` should be relative to `cwd`

mentat/session.py
```diff
@@ -81,7 +80,7 @@ def __init__(
             stream,
             llm_api_handler,
             cost_tracker,
-            git_root,
+            git_root,  # pyright: ignore
```
waydegg, line 83: Addressing this in another PR on this chain of branches

mentat/parsers/block_parser.py
```diff
@@ -167,12 +167,12 @@ def _special_block(
         if deserialized_json.action == _BlockParserAction.Delete:
             replacements.append(Replacement(starting_line, ending_line, []))
         file_edit = FileEdit(
-            git_root / deserialized_json.file,
```
PCSwingle, line 170: So the model is going to be outputting paths relative to the cwd instead of git_root now, right? That sounds good to me, just wanted to confirm.

mentat/session.py
```diff
@@ -57,8 +57,7 @@ def __init__(
 
         # Since we can't set the session_context until after all of the singletons are created,
         # any singletons used in the constructor of another singleton must be passed in
-        # TODO: An error is thrown in this function; once git root is removed, the error will be removed
-        git_root = get_shared_git_root_for_paths([Path(path) for path in paths])
```
PCSwingle, line 61: In config.py on line 150 we use this function to get the git root so that we can check for the project config; maybe you were already planning to do this in another PR, but the get_shared_git_root function will currently crash. I think we could change it to the get_git_root_for_path function and we could first check for a project config in the git root and then in the cwd if there is not git root.

mentat/parsers/block_parser.py
waydegg: Yup!

mentat/session.py
waydegg: > I think we could change it to the get_git_root_for_path function and we could first check for a project config in the git root and then in the cwd if there is not git root.

Agree on this being the behavior for determining the config to use. I'll add this in another PR!

PR #348, opened 2023-12-04
Maximum context checked

jakethekoenig:
Currently maximum context is determined at conversation initialization. Now it is checked before calling the llm as well. This allows users to change it mid session as well as preventing possible errors in long conversations.

### DIFF
--- a/mentat/conversation.py
+++ b/mentat/conversation.py
@@ -39,7 +39,6 @@ class Conversation:
 
         context_size = model_context_size(config.model)
         maximum_context = config.maximum_context
-        print(maximum_context, context_size)
         if maximum_context is not None:
             if context_size:
                 return min(context_size, maximum_context)
@@ -187,7 +186,7 @@ class Conversation:
                     content=prompt,
                 )
             )
-            return self._messages.copy() + [prompt_message]
+            return [prompt_message] + self._messages.copy()
 
     def clear_messages(self) -> None:
         """Clears the messages in the conversation"""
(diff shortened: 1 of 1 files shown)


mentat/conversation.py
```diff
@@ -27,14 +27,27 @@
...
         # This contains a list of messages used for transcripts
         self.literal_messages = list[TranscriptMessage]()
 
+    def _get_max_tokens(self) -> Optional[int]:
+        session_context = SESSION_CONTEXT.get()
+        config = session_context.config
+
+        context_size = model_context_size(config.model)
+        maximum_context = config.maximum_context
+        print(maximum_context, context_size)
```
PCSwingle, line 42: Debug statement?

mentat/conversation.py
```diff
@@ -86,14 +99,13 @@ async def display_token_count(self):
...
 
+        context_size = self._get_max_tokens()
         if not context_size:
             raise MentatError(
                 f"Context size for {config.model} is not known. Please set"
                 " maximum-context with `/config maximum_context value`."
             )
-        else:
-            self.max_tokens = context_size
-        if tokens + 1000 > context_size:
```
PCSwingle, line 96: Thanks for catching this! I thought I replaces all the token_buffer spots already but I must have missed this!

mentat/conversation.py
```diff
@@ -254,14 +266,42 @@ async def get_model_response(self) -> list[FileEdit]:
                     p.get("text", "") for p in prompt if p.get("type") == "text"
                 ]
                 prompt = " ".join(text_prompts)
+            max_tokens = self._get_max_tokens()
+            if max_tokens is None:
```
PCSwingle, line 270: Could we move this if statement outside the try catch and to the top of the function?

mentat/conversation.py
```diff
@@ -254,14 +266,42 @@ async def get_model_response(self) -> list[FileEdit]:
...
+                        f"The context size is limited to {max_tokens} tokens and"
+                        f" previous messages plus system prompts use {tokens} tokens."
+                        " Please use `/clear` to reset or restart the session.",
+                        color="light_red",
+                    )
+                else:
+                    stream.send(
+                        f"The context size is limited to {max_tokens} tokens and"
+                        f" previous messages plus system prompts use {tokens} tokens,"
+                        " leaving insufficent tokens for a response. Please use"
```
PCSwingle, line 290: Maybe mention the token buffer here? Otherwise I can imagine this being pretty confusing if you don't know what token_buffer is

mentat/conversation.py
```diff
@@ -175,7 +187,7 @@ def get_messages(self) -> list[ChatCompletionMessageParam]:
                     content=prompt,
                 )
             )
-            return [prompt_message] + self._messages.copy()
+            return self._messages.copy() + [prompt_message]
```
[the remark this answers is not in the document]
jakethekoenig: Thanks for catching. Reverting. I think this does improve performance (reduces off by one errors) when the code context is large. I need to do some benchmarking though.

mentat/conversation.py
jakethekoenig: I think I'll just remove the `/config` idea from the warning. They probably don't want to make it smaller than 1000 anyway and I don't want the warning to get too long and confusing.

PCSwingle (approved):
Looks great!

PR #349, opened 2023-12-04
Remove `git_root` from git handler functions

waydegg:
Some git handler functions relied on `SESSION_CONTEXT.git_root`, so this PR removes that dependency.

*This PR is part of a series of PRs to remove git_root from Mentat, and allow Sessions to be run anywhere on the filesystem.*


PCSwingle (approved):
Looks good to me. I'm glad we're moving away from having paths be centered at the git root.

PR #354, opened 2023-12-05
Move repos to project root

jakethekoenig:
Instead of being in `tests/benchmarks/repos` repos cloned for benchmarking purposes are put under `benchmark_repos`.


PCSwingle (approved):
Thanks, this is great!

granawkins (approved):

PR #358, opened 2023-12-07
Add auto_context_llm as a config argument

granawkins:
We removed this from the standard auto-context because it didn't seem very useful, but the code is still there so I've added it as an option:

`mentat -a --llm-feature-filter --maximum-context 8000`

I haven't gotten it to work yet, because instead of returning json it return `'```json<some json>'` instead? Didn't used to do that. Also don't forget to set a lower maximum_context, otherwise it will send the full 128k tokens on the LLMFeatureFilter.filter. Ideas for a better way to handle this, other than another arg called `auto_context_llm_tokens`?


granawkins:
> `mentat -a --maximum-context 8000 --llm-feature-filter-size 32000`

Just to clarify, `--maximum-context` is the global maximum and I think for the above you'd want to use `--auto-tokens`.

Another brainstorm (also don't know how much I like this) is to allow one or two arguments for `--auto-tokens`:

- `mentat -a --auto-tokens 8000`: Select 8k using embeddings
- `mentat -a --auto-tokens 320000 8000`: First select 32k using embeddings, then get max 8k from llm

biobootloader:
> Just to clarify, --maximum-context is the global maximum and I think for the above you'd want to use --auto-tokens.

ah yes that's what I meant 👍 

> I like it because having 32k 8k next to each other reflects how they're used.

I like that too!

PCSwingle:
> > instead of returning json it return '```json' instead? Didn't used to do that.
> 
> @PCSwingle do you know why? I remember you encountered some weird API issue with return type

I have encountered some bugs with the vision model relating to stuff like that, but that's it. I think Grant means the model is just returning the markdown text ```json; it thinks it's displaying it to the user and doesn't realize that it actually needs to output json. @granawkins have you tried the Json response format yet? The call_llm_api function should have a kwarg that you can set to json super easily.

granawkins:
Note from #401 - Add at test to make sure get_code_message can be called for LLMFeatureFilter

granawkins:
> Note from #401 - Add at test to make sure get_code_message can be called for LLMFeatureFilter

Actually the LLMFeatureFilter doesn't use the CodeContext, just Features, so this isn't a problem.

granawkins:
This is functioning fine now - slow, but consistently works for simple-enough tasks.

I add a config variable `llm_feature_filter: int = 0` and the `-l` flag. To get it working I also
- Have it return only files, not lines
- Return a dict of {path: reason} instead of a list of [path]. One reason is that GPT4 seems to only return dicts not lists. Also, providing a reason will let us inject more intelligence with fine-tuning.

jakethekoenig (approved):
Works for me. Thanks!

PR #359, opened 2023-12-07
Github action to run and upload benchmarks

jakethekoenig:
A daily github action that:
- Runs one small benchmark (we should expand after we've validated this)
- Uploads the results.html to a public s3 bucket.
    - Secrets to upload to the bucket are already added.
    - I think it's fine if the benchmarks are public since all the code to run them is public anyway. If we clean them up a bit and make a central dashboard I could even see hosting the bucket as a static site at benchmarks.abante.ai
- Makes a slack message in the benchmarking channel.
    - The webhook url is an already added secret. It's not that important to keep this secret but in theory people could spam us with it.

### DIFF
--- a/.github/workflows/benchmarks.yml
+++ b/.github/workflows/benchmarks.yml
@@ -5,6 +5,7 @@ name: Benchmark
 on:
   schedule:
     - cron: '0 0 * * *'  # Runs at midnight UTC every day
+  workflow_dispatch:
 
 jobs:
   run-benchmarks:
--- a/.github/workflows/benchmarks.yml
+++ b/.github/workflows/benchmarks.yml
@@ -1,0 +1,28 @@
+# This workflow runs our benchmarks uploads the results to S3 and sends a notification to
+# slack when it's done.
+name: Benchmark
+
+on:
+  schedule:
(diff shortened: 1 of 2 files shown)


.github/workflows/benchmarks.yml
```diff
@@ -0,0 +1,29 @@
+# This workflow runs our benchmarks uploads the results to S3 and sends a notification to
+# slack when it's done.
+name: Benchmark
+
+on:
+  schedule:
+    - cron: '0 0 * * *'  # Runs at midnight UTC every day
+  workflow_dispatch:
```
PCSwingle, line 8: Does this make it only run on main?

PCSwingle (approved):
Looks good to me, as long as it actually only runs once a day; I'm not exactly sure how cron or workflow_dispatch work, but I assume it just runs on the main branch

.github/workflows/benchmarks.yml
jakethekoenig: `workflow dispatch` makes it so that there's a button you can click in the actions tab to manually trigger the action. I want it for testing. Could also come in handy if we want to run it for releases as you can choose the branch to run it on.

PR #368, opened 2023-12-08
More benchmarks

jakethekoenig:
Two new benchmarks
* A benchmark that tests Mentat's ability to write an exercise runner for the Clojure language
* A benchmark that tests Mentat's ability to duplicate the white space pr from earlier this week

And Three new forms of grading:
* syntax judged by looking at the diff
* response judged for tone/style
* comparison grading verse human diff if applicable

Misc changes:
- All benchmarks ran for retries iterations.
- Benchmarks make their own configs (so we can iterate on `maximum_context`)


tests/benchmarks/benchmark_runner.py
```diff
@@ -55,14 +66,11 @@ async def test_benchmark(retries):
...
+            }
             for i in range(1, retries + 1):
-                client = PythonClient(
-                    config=Config(
-                        auto_context=True,
-                        maximum_context=8000,
-                    )
-                )
-
+                client = PythonClient(cwd=codebase, config=benchmark.config)
```
jakethekoenig, line 73: Previously I didn't need to specify the cwd but now I do. Not sure exactly why because I change the current working directory before each benchmark and PythonClient should use the current working directory if an argument isn't given.

PCSwingle (approved):
Looks great, thanks!

tests/benchmarks/benchmark_runner.py
```diff
@@ -24,8 +26,82 @@ def retries(request):
     return int(request.config.getoption("--retries"))
 
 
+async def grade(to_grade, prompt):
```
granawkins, line 29: I got this error testing locally on the Clojure one:
image

Maybe here we should check/truncate before sending to gpt4.

granawkins: Related - might be a good idea to put a try/except block around each individual test, so if one of them fails, you still get the results of the rest printed to the screen.

PR #371, opened 2023-12-08
fix display context double and white directory names

PCSwingle:
@granawkins @waydegg  is there a reason the included files was duplicated? I assume it was the result of a merge conflict with one of Wayde's big PR's, but not sure and wanted to make sure I didn't miss anything.


waydegg:
Yeah include files shouldn't have been duplicated

waydegg (approved):

PR #373, opened 2023-12-08
expand home paths

### LINKED ISSUE #355: Expand home (~) in /include /exclude

PCSwingle:
Fixes #355.


mentat/include_files.py
```diff
@@ -126,13 +126,16 @@ def validate_and_format_path(
     """
     path = Path(path)
 
+    # Resolve ~
+    path = path.expanduser()
```
granawkins, line 130:
```suggestion
    try:
        path = path.expanduser()
    except RuntimeError as e:
        raise PathValidationError(f"Error expanding path {path}.")
```
I was getting familiar with `expanduser` and see it might raise a RuntimeError. It's a rare case, but we're not handling them.

granawkins (approved):
Small suggestion, otherwise looks great

mentat/include_files.py
waydegg: +1

waydegg (approved):
LGTM, good to merge after addressing Grant's comment on handling `path.expanduser` exceptions

PR #374, opened 2023-12-09
Sample command

granawkins:
This adds a new command `/sample`, which bundles the last interaction together with active code (after you correct it) into a generic 'sample'. See mentat/sampler/README.md for more info.

This is a rebasing of #352. 

Steps to complete:

- [x] Add git commands to create required diffs/hexsha
- [x] Add a sampler class, which builds itself from context, with tests
- [x] Add a `/sample` command and an integration test of the Sample class
- [x] Add `Sample.evaluate` method
- [x] Add script for model-benchmarking with samples.


mentat/parsers/git_parser.py
```diff
@@ -72,6 +77,9 @@ def parse_string(self, git_diff: str) -> ParsedLLMResponse:
                         end_line = start_line + 1
                     else:
                         end_line = start_line + int(a_b[1])
+                    # TODO: This breaks some of the parser tests, but is required to make
+                    # an insertion diff valid? Need to investigate.
+                    end_line += 1  # FileEdits use exclusive end_line
```
granawkins, line 82: @jakethekoenig I had to add this to get my sample_eval's diff (insert 2 lines) to work. I wonder if we adjusted file_edits to use exclusive end_line and this class was missed? Or maybe I need to dig more 😄

mentat/command/commands/sample.py
```diff
@@ -0,0 +1,29 @@
...
+        sample.save(str(fpath))
+        SESSION_CONTEXT.get().stream.send(f"Sample saved to {fname}.", color="green")
+
+    @classmethod
+    def argument_names(cls) -> list[str]:
+        return ["path?"]
+
+    @classmethod
+    def help_message(cls) -> str:
+        return "Undo the last change made by Mentat"
```
PCSwingle, line 29: ^

mentat/sampler/README.md
```diff
@@ -0,0 +1,23 @@
...
+| title            | `str`         | plaintext by creator |
+| description      | `str`         | plaintext by creator |
+| id               | `uuid`        |  |
+| parent_id        | `uuid`        | id of sample immediately before this |
+| repo             | `str`         | a url to download the code |
+| merge_base       | `str`         | the latest permanent commit |
+| diff_merge_base  | `str`         | between merge_base and latest commit |
+| diff_active      | `str`         | between latest commit and active (pre-edit) code |
+| messages         | `list[dict]`  | user and assistant messages. |
+| args             | `list[str]`   | list of `<relative_path>[:<start_line>,<end_line>]` |
```
PCSwingle, line 15:
```suggestion
| args             | `list[str]`   | list of `<relative_path>[:<start_line>-<end_line>]` |
```

PCSwingle:
I'm a bit confused on this; so the sample creates a json file detailing the changes a model made to an exact commit on a repo? Kind of like a commit except it's in a json file and has extra information (like user and assistant messages)? Also, what happens if there are already diffs when you /sample; do they get lumped in with the model changes?

mentat/sampler/README.md
```diff
@@ -0,0 +1,23 @@
+# Mentat Sampler API
+The Sampler API is an open-source standard to captures interactions between a developer and an LLM-based AI Assistant. It's intended to facilitate sharing of benchmarks and fine-tuning data throughout the open-source community and industry.
```
PCSwingle, line 2: I was a bit confused by this and I work on mentat; I think it would be good if the description was a bit clearer. Maybe we could give a description of what it actually is? (A sample contains the git diff of an edit made by an LLM with the user and assistant messages leading up to that edit)

PCSwingle: Also, I think it would be a lot simpler to just point to a git commit on github with the specified diff instead of having diff_edit, merge_base, diff_merge_base, and other arguments. Is there a reason that wouldn't work as well?

mentat/sampler/README.md
```diff
@@ -0,0 +1,23 @@
...
+
+| Field            | Type          | Description |
+|------------------|---------------|-------------|
+| title            | `str`         | plaintext by creator |
+| description      | `str`         | plaintext by creator |
+| id               | `uuid`        |  |
+| parent_id        | `uuid`        | id of sample immediately before this |
```
jakethekoenig, line 9: `parent_id` is set if you make multiple samples in the same conversation? What's the point of tracking that information?

mentat/sampler/README.md
```diff
@@ -0,0 +1,23 @@
+# Mentat Sampler API
+The Sampler API is an open-source standard to captures interactions between a developer and an LLM-based AI Assistant. It's intended to facilitate sharing of benchmarks and fine-tuning data throughout the open-source community and industry.
+
+| Field            | Type          | Description |
+|------------------|---------------|-------------|
+| title            | `str`         | plaintext by creator |
+| description      | `str`         | plaintext by creator |
+| id               | `uuid`        |  |
```
jakethekoenig, line 8: Why do we need an `id` at all? Isn't a title enough? I guess we plan to make so many (eventually) that we won't have time to name them?

mentat/sampler/README.md
granawkins: I'm mostly concerned with being able to generate samples as I work, rather than having to sit down and 'make samples'. In practice I'm hardly ever working directly from a permanent commit because we squash-merge everything. They certainly can be blank, and maybe they usually will be.

mentat/config.py
```diff
@@ -100,6 +100,20 @@ class Config:
         converter=int,
     )
 
+    # Sample specific settings
+    sample_repo: str | None = attr.field(
+        default=None,
+        metadata={
+            "description": "A public url for a cloneable git repository to sample from."
```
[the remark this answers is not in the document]
granawkins: Nice, didn't think of that! I'm going to leave this optional available, and if it's missing, I'll grab the repo url and ask the user to confirm or enter a new one.

mentat/git_handler.py
```diff
@@ -199,3 +214,81 @@ def get_default_branch() -> str:
     except subprocess.CalledProcessError:
         # Handle error if needed or raise an exception
         raise Exception("Unable to determine the default branch.")
+
+
+def get_merge_base() -> str | None:
```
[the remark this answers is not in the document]
granawkins: Aah ya these are not coordinated. Good catch.

mentat/code_file_manager.py
```diff
@@ -72,6 +72,9 @@ async def write_changes_to_files(
         if not file_edits:
             return []
 
+        # Set pre-edit context in case /sample is called later
+        self.history.set_sample_diffs()
```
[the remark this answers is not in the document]
granawkins: I really it really only needs to be the `diff_active`. It's here so that, later, we can determine what changes were made in response to the last interaction. It requires a bit of diff algebra, which I also need to solve so I'll write it out here.

We want to do:
1. Set `diff_active` here, i.e. `git diff HEAD`
2. Mentat updates the code
3. You update the code
4. Set `diff_edit` to the diff between the current code and `diff_active`

In order to do #4, we need to call `git diff HEAD` again, and 'subtract' the outputs from #1.

mentat/sampler/sample.py
```diff
@@ -0,0 +1,248 @@
...
+            diff_active=code_file_manager.history.diff_active or "",
+            messages=messages,
+            args=args,
+            diff_edit=get_diff_active() or "",  # TODO: subtract diff_active
+            test_command=test_command,
+        )
+
+    def save(self, fname: str) -> None:
+        with open(fname, "w") as f:
+            json.dump(attr.asdict(self), f, indent=4)
```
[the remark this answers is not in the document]
granawkins: Ya we should somehow. Or setup a public repo that anyone can contribute to, and curate lists of IDs for different purposes.

mentat/sampler/README.md
granawkins: I'm imagining if you have 3 interactions in a row, you could use it to benchmark or file-tune an agent LLM.

mentat/sampler/README.md
granawkins: I wouldn't want to rely on names being unique - esp if it's public.

mentat/sampler/README.md
PCSwingle: I'm pretty sure the commits will still be there (even if they are squash-merged). Git squashes all the commits into one commit and puts it on main, but the commits are still valid and can still be accessed via their hash.

PCSwingle: > Still though if you're working from a branch or commit that _could_ get abandoned, and you want to create a sample, it'd be nice to have this option to make sure it'll always be valid.

Maybe, but as far as I know, git will only ever delete a commit if the user does it (via a command), which I don't think we should worry about. If the commit is taken down off of it's origin (like github), then odds are all the other commits also were, in which case this entire sample is moot anyways.

mentat/sampler/sample.py
```diff
@@ -0,0 +1,268 @@
...
+
+        # Run the test command
+        test_result: str = ""
+        if self.test_command:
+            test_result = subprocess.check_output(
+                self.test_command,
+                text=True,
+            )
+        # TODO: Run the LLM evaluations from benchmark
+        # TODO: Save the results
```
jakethekoenig, line 263: I assume these TODOs are for another PR? Doesn't seem necessary to nail down how samples will turned into benchmarks in this one.

mentat/sampler/README.md
granawkins: By abandoned I mean never pushed to GH/remote in the first place. Then it'd only available on my machine, right?

PCSwingle: Oh, I see, yeah.

jakethekoenig:
I get the following exception when I run the evaluate samples script with a dirty git repo:
image

mentat/sampler/sampler.py
```diff
@@ -0,0 +1,206 @@
...
+def parse_message(message: ChatCompletionMessageParam) -> dict[str, str]:
+    ctx = SESSION_CONTEXT.get()
+    content = message.get("content")
+    text, code = "", ""
+    if isinstance(content, str):
+        if message.get("role") != "assistant":
+            text = content
+        output = list[str]()
+        in_special = False
+        for line in content.splitlines():
```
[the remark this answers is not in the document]
granawkins: Agreed - I nearly took a crack at this.

mentat/sampler/sampler.py
```diff
@@ -0,0 +1,206 @@
...
+                " URL."
+            )
+            response = (await collect_user_input()).data.strip()
+            if response == "y":
+                repo = remote_url
+            else:
+                repo = response
+            config.sample_repo = repo
+
+        stream.send("Sample Title:")
```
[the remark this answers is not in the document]
granawkins: what don't you like about it?

mentat/sampler/sampler.py
```diff
@@ -0,0 +1,206 @@
...
+from mentat.sampler.utils import get_active_snapshot_commit
+from mentat.session_context import SESSION_CONTEXT
+from mentat.session_input import collect_user_input
+from mentat.utils import get_relative_path
+
+
+def parse_message(message: ChatCompletionMessageParam) -> dict[str, str]:
+    ctx = SESSION_CONTEXT.get()
+    content = message.get("content")
+    text, code = "", ""
```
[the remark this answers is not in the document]
granawkins: This was intentional - I mean to convert code messages to diff format, but we don't have an easy way of doing that (per your comment below), so it's a placeholder for now.

mentat/sampler/sampler.py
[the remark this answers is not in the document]
PCSwingle: I think a parse string method would be good; we'd only need to implement in the parser superclass so it wouldn't any overhead to new parsers which would be nice. I agree with jake that we shouldn't really be using _starts_special and _ends_special; those are really just meant to assist with streaming. I agree with jake that the best method would be to add a parse_string method, call that, and then extract the code from the output fileedits. We also might need to add a switch to turn off streaming for the stream and parse function if you don't want to stream the output.

granawkins:
> I get the following exception when I run the evaluate samples script with a dirty git repo: image

Im checking for text files now, so hopefully this won't happen.

jakethekoenig (approved):
Given this is for our own use I guess it's good enough. I'd really like to see the `parse_message` function improved though. Another idea would be to have the parser return the non code message separately and have the conversation hold onto that data.

Maybe we should hide the sample command and the config settings related to sampling? I think they'd confuse users.

PCSwingle:
I definitely think that until we change it to a parse_string function we shouldn't have this command be visible to users (since it's going to be super buggy if we use _starts_special and a few other parser specific functions in here which aren't meant to be used outside the parser).

PR #375, opened 2023-12-09
Fix crash if there's no internet

### LINKED ISSUE #178: Gracefully handle openai.error.APIConnectionError
If the user talks to mentat without an internet connection they get a `openai.error.APIConnectionError` and mentat crashes. It would be better to catch it and print a message like "Received openai.error.APIConnectionError. Are you connected to the internet?"

biobootloader:
Let me know if you think the `api_guard` decorator is a good idea or not. Also if there's a better name!

Fixes #178

### DIFF
--- a/mentat/llm_api_handler.py
+++ b/mentat/llm_api_handler.py
@@ -5,7 +5,7 @@ import io
 import os
 import sys
 from pathlib import Path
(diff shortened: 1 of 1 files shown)


granawkins (approved):
LGTM

PR #376, opened 2023-12-09
Add helper func to split interval string from path

granawkins:
It's been pointed out that the way we check for intervals `path_str.rsplit(":")` won't work on a Windows absolute path. This fixes that by doing it with `re` in a helper func.


PCSwingle (approved):
Thanks for adding this, this is super useful!

PR #377, opened 2023-12-10
Fix NameError: name 'FileEdit' is not defined on Rate Error crash

nilshamerlinck:
Tried to run `mentat -a` on mentat codebase itself and faced this NameError issue.


biobootloader (approved):
Thank you for the fix!

PR #378, opened 2023-12-11
Two exercism fixes

jakethekoenig:
* Grading wasn't updated to new openai version.
* Paths needed to be absolute after cwd change.


PCSwingle (approved):
:rocket:

PR #379, opened 2023-12-11
Exercism results viewer refactored to use jinja

jakethekoenig:
I let mentat write it almost entirely so the css looks a little different. I mostly prefer it.

Note this PR is diffed with my previous fixes pr.


tests/benchmarks/exercism_practice.py
```diff
@@ -177,7 +178,7 @@ def run_exercise_sync(problem_dir, language="python", max_iterations=2):
     return result
 
 
-def summarize_results(results):
+def tqdm_summary(results):
```
jakethekoenig, line 181: I renamed this to make it clear what it's purpose was and how it is distinct from the other summary method.

PCSwingle (approved):
Thanks for doing this! The new viewer also looks pretty sick!

PR #380, opened 2023-12-11
Transcript added to exercism results viewer

jakethekoenig:
* get_conversation now available from the python_client.
* transcript viewer logic from log viewer broken into an independent jinja template.


mentat/resources/templates/exercism_benchmark.jinja
```diff
@@ -1,3 +1,7 @@
+<<<<<<< Updated upstream
```
PCSwingle, line 1: This supposed to be here? Haha

jakethekoenig: woops

granawkins (approved):
Seems to work as advertised. Small suggestion, otherwise looks good.

PR #385, opened 2023-12-12
Add command completions and move completions to the server

PCSwingle:
TODO:
- [x] Finish completion lists for all commands.
- [x] Add completions for everything other than commands (function/class names, filenames, etc.)
- [x] Change command arguments to [IBM Command Syntax](https://www.ibm.com/docs/en/spectrum-archive-ee/1.2.6.0?topic=reference-command-syntax)

### DIFF
--- a/mentat/auto_completer.py
+++ b/mentat/auto_completer.py
@@ -16,6 +16,8 @@ from mentat.utils import get_relative_path
 
 SECONDS_BETWEEN_REFRESH = 5
 
+whitespace = " \t\r\n"
+
 
 class Completion(TypedDict):
     """
@@ -98,7 +100,9 @@ class AutoCompleter:
             for completion in filtered_completions
         ]
 
-    def _partial_shlex_split(self, argument_buffer: str) -> tuple[List[str], bool]:
+    def _partial_shlex_split(
+        self, argument_buffer: str
+    ) -> tuple[List[str], bool, bool]:
         """
         Handles unescaped backslashes and unclosed quotation marks
         """
@@ -116,10 +120,13 @@ class AutoCompleter:
                 in_quote = True
             elif "escaped" in str(e):
                 # Remove the offending backslash and pretend it isn't there
-                return self._partial_shlex_split(argument_buffer[:-1])
+                split_buffer, in_quote, _ = self._partial_shlex_split(
+                    argument_buffer[:-1]
+                )
+                return split_buffer, in_quote, True
             else:
                 raise ValueError(f"shlex.split raised unexpected error: {e}")
-        return split_buffer, in_quote
+        return split_buffer, in_quote, False
 
     def _find_shlex_last_word_position(
         self, argument_buffer: str, num_words: int
@@ -135,8 +142,6 @@ class AutoCompleter:
         return 0 if not remaining else -len(remaining[0])
 
     def _command_argument_completion(self, buffer: str) -> List[Completion]:
-        whitespace = " \t\r\n"
-
         if not any(space in buffer for space in whitespace):
             return self._replace_last_word(
                 buffer, [(name, name) for name in Command.get_command_names()]
@@ -152,10 +157,16 @@ class AutoCompleter:
             # shlex fails if there are uncompleted quotations or backslashes not escaping anything;
             # we check for this and try to complete the quotations/backslashes, so that we can get the
             # actual escaped last_word argument to compare against our possible completions.
-            split_buffer, in_quote = self._partial_shlex_split(argument_buffer)
+            split_buffer, in_quote, unescaped = self._partial_shlex_split(
+                argument_buffer
+            )
             last_word_position = self._find_shlex_last_word_position(
                 argument_buffer, len(split_buffer)
             )
+
+            # If we removed an ending \ and before it was whitespace, we need to move on one argument
+            if unescaped and buffer[-2] in whitespace:
+                split_buffer.append("")
 
             # shlex.split doesn't count the ending space
             if buffer[-1] in whitespace and not in_quote:
--- a/mentat/auto_completer.py
+++ b/mentat/auto_completer.py
@@ -1,0 +1,259 @@
+import shlex
+from typing import List, TypedDict
+
+from mentat.command.command import Command
+
+
+class Completion(TypedDict):
+    """
+    Represents a single unserialized auto-completion suggestion
+    """
+
+    # The completion to add to the buffer
+    content: str
+    # Number of characters from the cursor the completion should be placed (negative means backwards)
+    position: int
+    # The display name; if not given, will default to content
+    display: str | None
+
+
+whitespace = " \t\r\n"
+
+
+def _replace_last_word(
+    last_word: str, completions: List[tuple[str, str]], position: int | None = None
+) -> List[Completion]:
+    """
+    Takes a list of string completions along with their displays, and filters to match only completions
+    whose displays start with the last word, and then returns a list of Completions that will replace that word
+    """
+    filtered_completions = [
+        completion for completion in completions if completion[1].startswith(last_word)
+    ]
+    return [
+        Completion(
+            content=completion[0],
+            position=-len(last_word) if position is None else position,
+            display=completion[1],
+        )
+        for completion in filtered_completions
+    ]
+
+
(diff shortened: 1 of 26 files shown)


PCSwingle:
There are a whole lot of edge cases when it comes to the commands autocomplete (because we use shlex.split to split arguments); I've done a bunch of testing already, but very possibly missed something. If you want to smoke test the command auto complete using ', ", and \ (single and double quotes behave differently; in single quotes, backslashes aren't escape characters, but they are in double quotes) that would be great!

jakethekoenig:
Thanks so much! Looking at the code now. Some things I've noticed:
* When you tab complete a directory terminals add a trailing slash. It would be nice if we could too.
* When you tab complete a string with a quote the quote is removed. E.g. `"test` and `t"es` complete to `tests`.
* File tab completion should work in the `/run` command. It would be really great if you could get what the underlying shell would complete for a given token to also autocomplete things like pytest. Not sure if that's possible.

mentat/command/commands/search.py
```diff
@@ -36,10 +41,19 @@ async def apply(self, *args: str) -> None:
...
+        return [CommandArgument("required", "query")]
+
+    @override
     @classmethod
-    def argument_names(cls) -> list[str]:
-        return ["search_query"]
+    def argument_autocompletions(
+        cls, arguments: list[str], argument_position: int
+    ) -> list[str]:
+        return []
```
jakethekoenig, line 54: Should search have the same autocompletes as normal prompt mode?

mentat/command/command.py
```diff
@@ -37,18 +49,24 @@ def get_command_names(cls) -> list[str]:
...
     @abstractmethod
     async def apply(self, *args: str) -> None:
         pass
 
-    # TODO: make more robust way to specify arguments for commands
     @classmethod
     @abstractmethod
-    def argument_names(cls) -> list[str]:
+    def arguments(cls) -> List[CommandArgument]:
+        pass
```
jakethekoenig, line 59: Since a super majority of commands have no arguments I think it makes sense to make a default implementation for these two methods that returns `[]`.

mentat/llm_api_handler.py
```diff
@@ -122,45 +122,39 @@ def prompt_tokens(messages: list[ChatCompletionMessageParam], model: str):
     return num_tokens
 
 
+known_models = {
+    "gpt-4-1106-preview": (128000, 0.01, 0.03),
```
jakethekoenig, line 126: I'd prefer this tuple to be a dict with `context_size` and `pricing`. We may want to add more properties later.

mentat/session.py
```diff
@@ -100,6 +102,18 @@ def __init__(
         for path in paths:
             code_context.include(path, exclude_patterns=exclude_paths)
 
+    def _create_task(self, coro: Coroutine[None, None, Any]):
```
jakethekoenig, line 105: This is the same as the one in the clients right? I wonder if it makes sense to make them inherit it from an abstract `TaskManager` class? Might make it more confusing rather than less though.

mentat/session_input.py
```diff
@@ -57,9 +57,10 @@ async def collect_input_with_commands() -> StreamMessage:
     response = await collect_user_input()
     while isinstance(response.data, str) and response.data.startswith("/"):
         try:
-            arguments = shlex.split(response.data[1:])
-            command = Command.create_command(arguments[0])
-            await command.apply(*arguments[1:])
+            # We only use shlex to split the arguments, not the command itself
+            arguments = shlex.split(" ".join(response.data.split(" ")[1:]))
```
jakethekoenig, line 61: `split` has a maximum number arg which you can use instead of splitting and rejoining. e.g. `response.data.split(" ", 1)[1]` is equivalent to this argument.

jakethekoenig:
Another nice things shells do which I think you should as well: If there's only one possible completion and the user hits tab after inserting the completion add a space.

jakethekoenig (commented):
Thanks so much for doing this! It feels really great.

PCSwingle:
> Thanks so much! Looking at the code now. Some things I've noticed:
> 
> * When you tab complete a directory terminals add a trailing slash. It would be nice if we could too.
Oh, I meant to do this and completely forgot; thanks for bringing this up!
> * When you tab complete a string with a quote the quote is removed. E.g. `"test` and `t"es` complete to `tests`.
This is because we're doing shlex and quotes are treated as quotes; so they aren't actually part of the text (unless you surround it in ' or escape it with \). The reason autocomplete still works with t"es is because you could end it t"es"t which would evaluate to test so it's still technically could evaluate to test. 
> * File tab completion should work in the `/run` command. It would be really great if you could get what the underlying shell would complete for a given token to also autocomplete things like pytest. Not sure if that's possible.
I like this idea, let me add it! I don't know how shells do that, I might look into that and see if it's possible, but probably not even worth the effort given it's only one command (even though it would be very cool)

PCSwingle:
> Another nice things shells do which I think you should as well: If there's only one possible completion and the user hits tab after inserting the completion add a space.

Unfortunately prompt toolkits autocomplete is a bit weird and it won't recheck for new completions after pressing tab (so then you'll still have to delete the space and press it again to reactivate completions for the next argument). This actually makes the new trailing / that I added to directories a bit weird as well; I now add a trailing slash, but you can't jump into autocompleting the next part of the path (because of this annoying prompt toolkit quirk). I actually might have preferred it before the trailing slash; let me know what you think @jakethekoenig

mentat/auto_completer.py
```diff
@@ -0,0 +1,276 @@
...
+from pygments.token import Token
+from pygments.util import ClassNotFound
+
+from mentat.command.command import Command
+from mentat.session_context import SESSION_CONTEXT
+from mentat.utils import get_relative_path
+
+SECONDS_BETWEEN_REFRESH = 5
+
+whitespace = " \t\r\n"
```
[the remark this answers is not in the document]
PCSwingle: The reason I use this is actually because it's the same whitespace string that shlex uses; I could use .isspace but I think I'll keep this just for slight consistency with shlex.

mentat/command/commands/search.py
PCSwingle: Hmm good point; one thing that makes it a bit weird is that search is still a terminal command, so it's arguments are actually split up by space and processed with shlex (so quotes and backslashes will be treated weird). We could change it so only the search command just receives it's argument as the raw text it's user input, but that would be weird making an exception for only one command. Either way I like this idea and I think I'll implement it

jakethekoenig:
> > Another nice things shells do which I think you should as well: If there's only one possible completion and the user hits tab after inserting the completion add a space.
> 
> Unfortunately prompt toolkits autocomplete is a bit weird and it won't recheck for new completions after pressing tab (so then you'll still have to delete the space and press it again to reactivate completions for the next argument). This actually makes the new trailing / that I added to directories a bit weird as well; I now add a trailing slash, but you can't jump into autocompleting the next part of the path (because of this annoying prompt toolkit quirk). I actually might have preferred it before the trailing slash; let me know what you think @jakethekoenig

Ugh yeah that is annoying. It is better the other way I guess so they can type `/` and see the completions for the  subdirectory

mentat/command/command.py
PCSwingle: I was thinking about this, but I think it's best to keep it explicit because otherwise I could see somebody adding a command and not realizing that they can add these functions. This way, it barely takes any extra time when adding a command, and you have to at least think about whether or not you want auto completions and arguments for this command.

mentat/session.py
PCSwingle: If we have more than 2 places we need this functionality we can make a superclass haha

mentat/session_input.py
PCSwingle: I was using that but then I have to check if the list is empty (since [1] returns index out of range error) and I eventually decided that this was cleaner

PCSwingle:
> Ugh yeah that is annoying. It is better the other way I guess so they can type `/` and see the completions for the subdirectory

Unfortunately agree :upside_down_face: 

TODO:
- [ ] Change back to no trailing slash
- [ ] Change model context size and price to TypedDict or attr
- [ ] Change /exclude to only autocomplete files in include_files? this might be too much work for a small thing, but would be a really nice touch

jakethekoenig (approved):

PR #386, opened 2023-12-13
insert code_message before conversation, right after system_prompt

granawkins:


biobootloader:
I'm going to run a few comparisons before approving, but expect this to be an improvement! Especially for longer context scenarios.

biobootloader (approved):
I ran several times both on this commit and the previous commit with high amounts of context (`mentat -a --auto-tokens 64000`) and it does seem to be significantly better. Still struggles sometimes with that much context though.

I also had a longer conversation with it on a smaller context setting to make sure it held up, and it seemed to.

PR #388, opened 2023-12-13
relax constraints for jsonschema package

rupurt:


PCSwingle (approved):
Just checked and this version should work fine; is there a specific reason you want a lower version of jsonschema? Does it conflict with a different package you're using?

rupurt:
Yeah. It conflicts with a mainframe package https://github.com/zowe/zowe-client-python-sdk

PCSwingle (approved):

PR #389, opened 2023-12-13
BenchmarkResult class introduced

jakethekoenig:
Miscellaneous Changes
* `session_context` added to `session` and used to access transcripts and costs.
* cost added to benchmark dashboard.

Todo:
* [ ] Use same BenchmarkResult in benchmark_runner.
* [ ] Make BenchmarkResultSummary class.
* [ ] Make dashboard for benchmark_runner.

### DIFF
--- a/tests/benchmarks/benchmark_result.py
+++ b/tests/benchmarks/benchmark_result.py
@@ -1,0 +1,29 @@
+import json
+from typing import Optional
+
+import attr
+
+from mentat.transcripts import Transcript
+
+
+@attr.define
+class BenchmarkResult:
+    iterations: int
+    passed: bool
+    name: str
+    cost: float
+    tokens: int
+    transcript: Optional[Transcript] = attr.ib(default=None)
+    instructions: Optional[str] = attr.ib(default=None)
+    code: Optional[str] = attr.ib(default=None)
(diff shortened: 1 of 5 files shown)


tests/benchmarks/benchmark_result.py
```diff
@@ -0,0 +1,29 @@
+import json
+from typing import Optional
+
+import attr
+
+from mentat.transcripts import Transcript
+
+
+@attr.define
+class BenchmarkResult:
```
PCSwingle, line 10: I would've preferred to make this a TypedDict if 3.11 was our minimum version (NotRequired was added), but for 3.10 support I think this (with to_json and from_json) is definitely the best call.

PCSwingle (approved):
Thanks for adding this!

jakethekoenig:
> Thanks for adding this!

I was going to add the todos to this PR but since you've approved it and this change sort of stands alone I'll go ahead and merge.

PCSwingle:
> > Thanks for adding this!
> 
> I was going to add the todos to this PR but since you've approved it and this change sort of stands alone I'll go ahead and merge.

Oh yeah I somehow completely missed the todo's when I reviewed this; either way, this is a good modular change so I'm happy with having a small PR like this.

PR #390, opened 2023-12-13
Benchmark Runner results shown in viewer

jakethekoenig:
The BenchmarkResult class and it's viewer are made more general with the following changes:
* Attributes of a benchmark result can be marked aggregateable and if they are the summary will compute and display the appropriate statistic for them: average, percent, sum. The property is not displayed if it is none for all benchmarks. If only a a subset of the benchmarks define it the percent is given in parenthesis next to the aggregated number.
* Attributes of a benchmark can specify how they should be displayed: text, code, transcript or json.
Now if you introduce a new benchmark with a new property you can add the property to the BenchmarkResults class along with the appropriate metadata. This won't break previous benchmarks and if appropriate it is easy to set it for them.

misc:
* previously the data was rewritten to txt on multiple runs. This is fixed.
* Rendering the summary is done by the summary.

todo:
* [x] actually use it in the benchmark runner.
* [x] Aggregate indentation/syntax errors.
* [x] Have the jinja template iterate over non-aggregated display properties instead hard coding.
* [x] Group exercises in the selector. So exercises that differ only by prompt or are just repeat iterations are together.
* [x] When a field represents an aggregation of a subset denote that with parenthesis.

### DIFF
--- a/mentat/resources/templates/benchmark.jinja
+++ b/mentat/resources/templates/benchmark.jinja
@@ -69,7 +69,7 @@
                                 {% elif section.type == "code" %}
                                     <pre><code>{{ section.content }}</code></pre>
                                 {% elif section.type == "json" %}
-                                    <pre><code>{{ section.content |tojson(indent=4)|safe}}</code></pre>
+                                    <pre>{{ section.content |tojson(indent=4)|safe}}</pre>
                                 {% elif section.type == "transcript" %}
                                     {{ transcript_container(section.content) }}
                                 {% endif %}
--- a/mentat/resources/templates/benchmark.jinja
+++ b/mentat/resources/templates/benchmark.jinja
@@ -28,6 +28,14 @@
             });
         </script>
         <script>
+            function toggleVisibility(family) {
+                const familyButtons = document.querySelectorAll('.' + family + '-child');
+                familyButtons.forEach(button => {
+                    button.classList.toggle('hidden');
+                });
+            }
+        </script>
+        <script>
             {% include 'js/transcript.js' %}
         </script>
     </head>
@@ -39,8 +47,15 @@
         </div>
         <div id="container">
             <div id="selector">
-                {% for test, result in summary.results_map.items() %}
-                    <button class="{{ result.display_color() }}" onclick="window.location.hash='{{ result.escaped_name() }}'">{{ test }}</button>
+                {% for family, results in summary.result_groups.items() %}
+                    {% if results|length == 1 %}
+                        <button class="{{ results[0].display_color() }}" onclick="window.location.hash='{{ results[0].name }}'">{{ results[0].name }}</button>
+                    {% else %}
+                        <button class="{{ results[0].display_color() }}" onclick="toggleVisibility('{{family}}')">{{family}}</button>
+                        {% for result in results %}
+                            <button class="hidden {{ family }}-child family-child {{ result.display_color() }}" onclick="window.location.hash='{{ result.name }}'">{{ result.name }}</button>
+                        {% endfor %}
+                    {% endif %}
                 {% endfor %}
             </div>
             <div id="viewer">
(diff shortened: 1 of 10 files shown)


jakethekoenig:
Here's a screenshot of what it looks like with on the new benchmarks:
image

You can reproduce this with the command:
```
pytest tests/benchmarks/benchmark_runner.py --benchmark -s --retries 2
```
It takes a couple of minutes to run, the following is faster:
```
pytest tests/benchmarks/benchmark_runner.py --benchmark -s --benchmarks clojure
```

granawkins:
It seems not all of the code is appearing in the code box? 
image

jakethekoenig:
> It seems not all of the code is appearing in the code box? 
> image
>

Oops, git diffs don't display untracked files. I think I'll run `git add .` and `git diff --staged`.

tests/benchmarks/benchmark_runner.py
```diff
@@ -154,43 +156,66 @@ async def test_benchmark(retries, benchmarks):
...
                 await client.shutdown()
 
                 if hasattr(benchmark, "verify"):
-                    if benchmark.verify():
-                        results[title][prompt]["Verified"] += 1
+                    result.verify = benchmark.verify()
 
-                diff = repo.git.diff(start_commit)
+                # Set syntax and response grade information
+                diff = repo.git.diff()
```
granawkins, line 189: afaik this only saves edits, not new/untracked files. I ran into the same thing in #374 and added a helper func in git_handler. (I think this is why the code was missing in my earlier comment)

granawkins (commented):
I like it, should help us understand the shoggoth a bit better.

tests/benchmarks/benchmark_runner.py
granawkins: Now i see you already caught it. I think stage/diff/unstage is the right approach too.

mentat/resources/templates/benchmark.jinja
```diff
@@ -0,0 +1,83 @@
...
+<html>
+    <head>
+        <title>Benchmark Results</title>
+        <style>
+            {% include 'css/benchmark.css' %}
+            {% include 'css/transcript.css' %}
+        </style>
+        <link rel="stylesheet" href="https://cdnjs.cloudflare.com/ajax/libs/highlight.js/11.9.0/styles/default.min.css">
+        <script src="https://cdnjs.cloudflare.com/ajax/libs/highlight.js/11.9.0/highlight.min.js"></script>
+        <script>hljs.initHighlightingOnLoad();</script>
```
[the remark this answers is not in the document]
jakethekoenig: Oh yeah, that does sort of seem bad in the json code blocks where true mostly means error detected. I'll check. If it's hard I'll just turn off syntax highlighting for json since it doesn't add much anyway

biobootloader:
Very nice! I really like the top bar.

Instead of showing a percentage in parenthesis when only a subset of benchmarks define an attribute, how about we show the actual counts? I.e. below it could show: 87.5% (8/14). That'd be useful even when all benchmarks do define the attribute, so you can instantly see how significant the percentage is (i.e. 100% 2/2 vs 100% 20/20)
image

biobootloader:
No code showing for this one either, but this seems different as this file shouldn't be untracked?

image

jakethekoenig:
> No code showing for this one either, but this seems different as this file shouldn't be untracked?
> 
> image

I think this is an example of mentat making a mistake. You can see in the code block that the line it's replacing already says `white-space: pre-wrap`. I think no diff is actually correct for that example.

jakethekoenig:
After this PR I'll edit the cron job to run the new benchmarks as well as all the javascript, python and clojure exercisms. When I do that I'll change the Benchmark summary to be saved as json, get a summary message from it and send it along with the link to slack. And also sync that json to its own bucket so it will be easy to parse old data and make graphs in the future.

granawkins (approved):

granawkins:
LGTM

Issue #327, opened 2023-11-29
Add support for OpenAI API-compatible local LLM servers (e.g. Ollama, LMStudio, GPT4All)


dvega-flexion:

jakethekoenig:
They already are supported. If you specify the OPENAI_API_BASE it should work. See [here](https://github.com/AbanteAI/mentat/blob/main/docs/configuration.md#-alternative-models).

rupurt:
I've tried getting `mentat` running with `ollama` and it seems like there are a couple of pieces missing besides just setting `OPENAI_API_BASE`.

I also had to set `OPENAI_API_KEY=anything`.

However it fails to respond to any requests as it can't use any embeddings. The list of embedding models appears [hard coded](https://github.com/AbanteAI/mentat/blob/2414ba4fceb9cb8613bf67d602fea7e2bc3a50fa/mentat/llm_api_handler.py#L125) and I can't figure out how to set something like `from langchain.embeddings import OllamaEmbeddings`

rupurt:
If I try to use an OpenAI embedding model like `gpt-3.5-turbo` while having the `OPENAI_API_BASE` it fails when using `ollama` and `litellm` because `litellm` doesn't have that embedding model and I can't use one of their supported embedding models because it's not in the [hardcoded](https://github.com/AbanteAI/mentat/blob/2414ba4fceb9cb8613bf67d602fea7e2bc3a50fa/mentat/llm_api_handler.py#L125) list of `mentat`.

rupurt:
Thinking about this some more. I think it can be solved by adding an extra cli option. When you run `mentat` with an unknown `--model ...` it asks you to the the `--maximum-context ...`. I will try and figure out how I need to add an option like `--maximum-embedding-context ...` use that instead of [raising an error](https://github.com/AbanteAI/mentat/blob/2414ba4fceb9cb8613bf67d602fea7e2bc3a50fa/mentat/embeddings.py#L120)

biobootloader:
@rupurt which embedding model are you trying to use?

The auto context feature (embeddings / RAG) is new and we haven't tried embedding models other than `text-embedding-ada-002` yet, but we are prioritizing adding support for other models.

You're only getting this error when running with auto-context, right? (i.e. `mentat -a`?)

rupurt:
[...]

PR #397, opened 2023-12-14
Embeddings not used if auto_context false

jakethekoenig:


biobootloader (approved):
💯

rupurt:
@jakethekoenig I'm not sure if this fixes the `--embedding-model` issue in #327. `mentat` hits the `POST /embeddings` endpoint of `OPENAI_API_BASE` and most local models don't support that endpoint. You need to route it to a model that does support it. Do you have any suggestions on how I can achieve that?

PR #398, opened 2023-12-14
add the length of include_files to auto-tokens so that auto-selected …

granawkins:
…files are in addition to included.


PCSwingle (approved):
I think this is a good fix for now; I'm about to start working on persistent context (probably after this PR gets merged) and I plan on cleaning up a lot of this code, including separating features and include_files.

PR #399, opened 2023-12-15
Feedback button added

jakethekoenig:
If you click the button and enter optional feedback your transcript is backed up to an s3 button and we are notified on slack via a lambda.

There's definitely a lot to improve:
* The UX of the dialog.
* There should be a `/feedback` command.
* The lambda should probably be in version control in this repo and deployed via github action.
* The s3 bucket should sit behind a url like transcripts.mentat.ai
* As our jinja templates get more complicated we should eventually introduce some automated linting/testing for them.

But I'd rather not spend too much time polishing. I think we'll get a lot of value from this as it is.

### DIFF
--- a/mentat/resources/templates/css/transcript.css
+++ b/mentat/resources/templates/css/transcript.css
@@ -93,6 +93,26 @@ pre {
     cursor: pointer;
 }
 
+.feedback_button {
+    position: absolute;
+    bottom: 20px;
+    right: 70px;
+    cursor: pointer;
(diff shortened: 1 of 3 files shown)


PCSwingle (approved):
I agree that it's pretty basic, but definitely good for now! Definitely think a /feedback command would be good. One other thing I would do is make the feedback button a bit larger; it's super easy to miss as it is in the bottom right of the screen. Thanks for adding this!

PR #402, opened 2023-12-16
remove extra ask_yes_no for deleting files

granawkins:


PCSwingle (approved):
Thanks for doing this!

PR #404, opened 2023-12-18
Implement Poetry for Dependency Management

use-the-fork:
#### Description

This PR represents a significant update in how we handle dependency management in our project. We have removed the traditional `dev-requirements.txt` and `requirements.txt` files and have transitioned to using Poetry. This change aims to simplify the declaration and management of dependencies and ensure consistent installs across various environments.

#### Changes

- Deleted `dev-requirements.txt`
- Deleted `requirements.txt`
- Added `poetry.lock` file
- Moved all dependencies previously listed in `requirements.txt` and `dev-requirements.txt` to the new Poetry lock file

#### Rationale

Poetry provides several advantages over traditional requirements files:

1. **Simplified Dependency Declaration:** Poetry uses the `pyproject.toml` file to declare dependencies, making it more straightforward and readable.
2. **Dependency Locking:** With the `poetry.lock` file, we can lock the versions of our dependencies, ensuring that every developer and environment uses the same versions, leading to fewer "it works on my machine" issues.
3. **Streamlined Dependency Resolution:** Poetry resolves dependencies in a more sophisticated way, reducing the chances of version conflicts.

#### How to Test

1. Ensure Poetry is installed on your system.
2. Run `poetry install` to install dependencies from the new `poetry.lock` file.
3. Run your usual test suite to ensure everything works as expected.

#### Additional Notes

- Developers should now use Poetry commands to manage dependencies.
- The CI/CD pipeline will need to be updated to reflect this change in dependency management.


biobootloader:
Thanks @use-the-fork! I haven't used Poetry before but will look into it so I better understand the benefits here.

gregory-lifhits-assaabloy:
> Thanks @use-the-fork! I haven't used Poetry before but will look into it so I better understand the benefits here.

no problem! There are a bunch of benefits but the big one for me is the fact that it automatically creates a venv for Mentat to run in. Dependency management is better too. Let me know what else you may need from me on this. I am working on getting the above tests working. 

Great work on this project btw its been a real game changer for me!

PR #409, opened 2023-12-19
Fix for #408 (explicitly use AsyncAzureOpenAI when required)

### LINKED ISSUE #408: Azure OpenAI no longer after updating to openai 1.x
In #303, the version of the `openai` library was updated. This has caused Azure OpenAI to no longer work with mentat by simply setting the `OPENAI_API_BASE` environment variable. Crucially, support for setting the required `api_version` parameter was removed from the "regular" `OpenAI` class, and is now only available on the new, separate `AzureOpenAI` class:

https://github.com/openai/openai-python/blob/main/README.md#microsoft-azure-openai
https://github.com/openai/openai-python/discussions/742

Fortunately, other than that, the Azure-specific client is mostly a drop-in replacement for the prior generic one. I have a proposed fix that I will open a PR for shortly.

laszlovandenhoek:
Fixes #408

### DIFF
--- a/README.md
+++ b/README.md
@@ -66,6 +66,13 @@ You'll need to have API access to GPT-4 to run Mentat. There are a few options t
 2. Run `export OPENAI_API_KEY=<your key here>` prior to running Mentat
 3. Place the previous command in your `.bashrc` or `.zshrc` to export your key on every terminal startup
 
+### Azure OpenAI
+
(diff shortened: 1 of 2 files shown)


PCSwingle (approved):
Thanks for fixing this! It is really weird that Azure has a different model name than the normal OpenAI API haha

laszlovandenhoek:
Yeah, that was a real head scratcher. Hence the "🙄"... but thanks for linting my code!

PR #410, opened 2023-12-19
Files outside cwd can be edited

jakethekoenig:
The code_file_manager uses absolute paths as keys but paths passed in may not be resolved leading to keyerrors. The cwd passed to sessions are also now resolved.

To be concrete this fixed two bugs:
* If you started mentat with `mentat --cwd mentat` (from outside the mentat dir) you could not make file edits.
* If you added a file outside the cwd such as `mentat ~/.zshrc` (assuming you're not in home) you could not edit it.

### DIFF
--- a/mentat/code_file_manager.py
+++ b/mentat/code_file_manager.py
@@ -102,7 +102,7 @@ class CodeFileManager:
                 continue
 
             if not file_edit.is_creation:
-                stored_lines = self.file_lines[file_edit.file_path]
+                stored_lines = self.get_file_lines(file_edit)
                 if stored_lines != self.read_file(file_edit.file_path):
                     logging.info(
                         f"File '{file_edit.file_path}' changed while generating changes"
@@ -152,3 +152,7 @@ class CodeFileManager:
             ]
             text = "\n".join(filtered_lines)
         return sha256(text)
+
+    def get_file_lines(self, file_edit: FileEdit) -> list[str]:
+        path = file_edit.file_path.resolve()
+        return self.file_lines[path]
--- a/mentat/code_file_manager.py
+++ b/mentat/code_file_manager.py
@@ -102,7 +102,7 @@ class CodeFileManager:
                 continue
 
             if not file_edit.is_creation:
-                stored_lines = self.get_file_lines(file_edit)
+                stored_lines = self.file_lines[file_edit.file_path]
                 if stored_lines != self.read_file(file_edit.file_path):
                     logging.info(
                         f"File '{file_edit.file_path}' changed while generating changes"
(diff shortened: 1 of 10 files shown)


mentat/python_client/client.py
```diff
@@ -20,7 +20,7 @@ def __init__(
         pr_diff: str | None = None,
         config: Config = Config(),
     ):
-        self.cwd = cwd
+        self.cwd = cwd.resolve()
```
PCSwingle, line 23: We should probably also expand home

mentat/terminal/client.py
```diff
@@ -244,7 +244,7 @@ def run_cli():
     Config.add_fields_to_argparse(parser)
     args = parser.parse_args()
 
-    cwd = args.cwd
+    cwd = Path(args.cwd).resolve()
```
PCSwingle, line 247: Same here

mentat/code_file_manager.py
```diff
@@ -102,7 +102,7 @@ async def write_changes_to_files(
                 continue
 
             if not file_edit.is_creation:
-                stored_lines = self.file_lines[file_edit.file_path]
+                stored_lines = self.get_file_lines(file_edit)
```
PCSwingle, line 105: Hmm, file edit's aren't supposed to have relative paths (it should always be absolute); how do I reproduce the problem? I think we should fix the actual problem (file edit not having absolute paths) rather than throwing this bandaid on it

jakethekoenig: I updated the description with two ways to reproduce. They are absolute. But they also have '..''s in them.

jakethekoenig: e.g. I get errors like: `KeyError: PosixPath('/home/jake/Development/mentat/../../dotfiles/zsh/zshrc')`

jakethekoenig: Oh yeah, I'll do that. It'll probably fix another bug I've noticed that isn't fixed by this: the line numbers aren't displayed correctly while streaming:
image

jakethekoenig: Display info also fixed now.

mentat/parsers/block_parser.py
```diff
@@ -149,9 +149,12 @@ def _special_block(
...
             raise ModelError("Error: Model output malformed edit.")
 
-        file_lines = self._get_file_lines(
-            code_file_manager, rename_map, deserialized_json.file
+        full_path = (cwd / deserialized_json.file).resolve()
+        rename_file_path = (
+            (cwd / deserialized_json.name).resolve() if deserialized_json.name else None
         )
+
+        file_lines = self._get_file_lines(code_file_manager, rename_map, full_path)
```
PCSwingle, line 157: This is going to lead to a bug when renaming files; _get_file_lines expects a relative path; I'm going to see if I can change the rename map to use absolute paths instead.

PCSwingle (approved):
Thanks for finding and fixing this!

PR #415, opened 2023-12-19
Persistent context

PCSwingle:
Add's persistent context. Dependent on #401

### DIFF
--- a/mentat/command/commands/clear.py
+++ b/mentat/command/commands/clear.py
@@ -35,5 +35,5 @@ class ClearCommand(Command, command_name="clear"):
     @classmethod
     def help_message(cls) -> str:
         return (
-            "Clear the current message history and auto included features from context."
+            "Clear the current message history and auto included code features from context."
         )
--- a/mentat/command/commands/clear.py
+++ b/mentat/command/commands/clear.py
@@ -12,8 +12,10 @@ class ClearCommand(Command, command_name="clear"):
         session_context = SESSION_CONTEXT.get()
         stream = session_context.stream
         conversation = session_context.conversation
+        code_context = session_context.code_context
 
         conversation.clear_messages()
+        code_context.clear_auto_context()
         message = "Message history cleared"
         stream.send(message, color="green")
 
@@ -32,4 +34,6 @@ class ClearCommand(Command, command_name="clear"):
     @override
     @classmethod
     def help_message(cls) -> str:
-        return "Clear the current message history."
+        return (
+            "Clear the current message history and auto included features from context."
+        )
--- a/mentat/command/commands/clear.py
+++ b/mentat/command/commands/clear.py
@@ -35,5 +35,6 @@ class ClearCommand(Command, command_name="clear"):
     @classmethod
     def help_message(cls) -> str:
         return (
-            "Clear the current message history and auto included code features from context."
+            "Clear the current message history and auto included code features from"
+            " context."
         )
--- a/mentat/config.py
+++ b/mentat/config.py
@@ -111,7 +111,7 @@ class Config:
                 " context. Adds this many tokens to context each request."
             ),
             "abbreviation": "a",
-            "const": 2000,
+            "const": 5000,
         },
         converter=int,
         validator=validators.ge(0),  # pyright: ignore
--- a/mentat/config.py
+++ b/mentat/config.py
@@ -111,6 +111,7 @@ class Config:
                 " context. Adds this many tokens to context each request."
             ),
             "abbreviation": "a",
+            "const": 2000,
         },
         converter=int,
         validator=validators.ge(0),  # pyright: ignore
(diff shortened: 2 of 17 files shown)


mentat/config.py
```diff
@@ -103,21 +103,17 @@ class Config:
         factory=list,
         metadata={"description": "List of glob patterns to exclude from context"},
     )
-    auto_context: bool = attr.field(
```
granawkins, line 106: Could we somehow continue to support running `mentat -a` without having to write a token amount?

granawkins: I should've finished #358 by now, but I'd planned to support any of:
`mentat -a`              default auto-context (8k)
`mentat -a 9000`         9k of auto-context
`mentat -a 32000 8000`   32k to LLMFeatureFilter, 8k max auto_features

Thoughts?

biobootloader: > mentat -a default auto-context (8k)
mentat -a 9000 9k of auto-context
mentat -a 32000 8000 32k to LLMFeatureFilter, 8k max auto_features

this is how I think it should be called as well

biobootloader:
should we warn users if the context is getting large? like 32k or so? Obviously the model can go further than that but I think it hurts performance badly enough to warn. Like a suggestion to run `/clear` if the earlier context is no longer relevant

mentat/config.py
PCSwingle: Just pushed a commit that sets the default to 2000 if no value is given (but still disabled if -a isn't given, and still takes a user value if a number is given after). I changed the default to 2000 because most of the time whenever I use auto context the vast majority of the tokens are useless, and now with persistent context I think it makes more sense than ever to have the default be lower, but we can move it back up if you want to.

PCSwingle: > I should've finished #358 by now, but I'd planned to support any of: `mentat -a` default auto-context (8k) `mentat -a 9000` 9k of auto-context `mentat -a 32000 8000` 32k to LLMFeatureFilter, 8k max auto_features
> 
> Thoughts?

Why don't we have a different flag for llm feature filter max compared to auto features? It doesn't make much sense to me to include them both in a single flag, and especially not to swap the numbers when you add a second one.

PCSwingle:
> should we warn users if the context is getting large? like 32k or so? Obviously the model can go further than that but I think it hurts performance badly enough to warn. Like a suggestion to run `/clear` if the earlier context is no longer relevant

I think this is a good idea, I'll add it.

mentat/config.py
granawkins: It's concise, does everything we need, and kinda mirrors the actual process - first 32k, then 8k.

Think of it like Python's `range` function; by default the first kwarg (start) is 0, and if called with a single arg it binds to the second kwarg (end).

granawkins: should we add support for 'k'?

PCSwingle: > It's concise, does everything we need, and kinda mirrors the actual process - first 32k, then 8k.
> 
> Think of it like Python's `range(start, stop, step)` function; by default the first kwarg (start) is 0, and if called with a single arg it binds to the second kwarg (stop).

I was thinking of this function exactly, and I really dislike it haha. Multiple times I've written a for loop with just the end and then realized I needed a step later on, and didn't realize that because I only added one number it still didn't add the step. It is the only function to change the position of it's arguments based on how many you have. I think it's bad design, and I also think it would be bad design for us to change the order of arguments based on how many we have. Sorry for the rant; I really dislike how the range function works.

granawkins: How about
`mentat -a`
`mentat -a 8000`
`mentat -a 8000 `-lff 32000` or `--llm-feature-filter`

PCSwingle: That sounds good to me (although I would prefer abbreviation to just be -l); this doesn't relate to my PR at all though, right? Should I add this in this PR or are we just discussing this for future reference.

biobootloader: Paul I see your point about the range function: you just want to add a step but now you have to add two more args. I don't think that applies here though as we only have 2 arguments, not 3.

I'm ok with an extra `-lff` or `-l` though, this is just an experimental feature for now.

mentat/command/commands/clear.py
```diff
@@ -32,4 +34,6 @@ def argument_autocompletions(
     @override
     @classmethod
     def help_message(cls) -> str:
-        return "Clear the current message history."
+        return (
+            "Clear the current message history and auto included features from context."
```
biobootloader, line 38: "code features" instead of "features" as users won't know what a feature is

```suggestion
            "Clear the current message history and auto included code features from context."
```

mentat/config.py
biobootloader: > should we add support for 'k'?

I like this but not in this PR

biobootloader:
Maybe instead of just "Total token count" we should now display a breakdown of user included vs auto context tokens? We can just make a new issue, this PR is good to merge as is

image

biobootloader (approved):

mentat/config.py
granawkins: Ya, originally I just wanted to keep a default value of `-a` and we've got that so I'm good for this PR.

PR #419, opened 2023-12-21
add generic Parser.parse_llm_response to parse without printing.

granawkins:


mentat/parsers/parser.py
```diff
@@ -86,8 +88,12 @@ async def stream_and_parse_llm_response(
         stream = session_context.stream
         code_file_manager = session_context.code_file_manager
 
-        printer = StreamingPrinter()
-        printer_task = asyncio.create_task(printer.print_lines())
+        if self._silence_printer:
+            printer = DummyPrinter()
```
PCSwingle, line 92: Instead of making a dummy printer, we could just not start printer_task (keep it None); I think this would be better but let me know what you think.

granawkins: I don't love it either but there's like 20 `printer.x` statements in stream_and_parse_llm_response, and the alternative seems like either a ton of if statements, or a refactor (which we wanted to avoid). Right?

granawkins: Aah I see what you mean now. Ya that's much simpler.

biobootloader (approved):

PR #420, opened 2023-12-21
Fix: user message is string unless image included

jakethekoenig:
Certain openai "compatible" endpoints want the content to be a string and fail if type is specified. This change makes us compatible with llama.cpp and litellm's openai compatible endpoints.


PCSwingle (approved):
Does this work in transcript viewer? Just wanted to confirm before merging. Thanks for getting this done!

jakethekoenig:
> Does this work in transcript viewer? Just wanted to confirm before merging. Thanks for getting this done!

Yep, I checked.

PR #421, opened 2023-12-21
Fix: Python client throws exception with session

jakethekoenig:
Currently if the session fails the python client will continue to wait for input. This change interrupts the call_mentat function when the session fails and prints out the session's exception.

### DIFF
--- a/mentat/session.py
+++ b/mentat/session.py
@@ -102,6 +102,7 @@ class Session:
         )
         self.ctx = session_context
         SESSION_CONTEXT.set(session_context)
+        self.error = None
 
         # Functions that require session_context
         check_version()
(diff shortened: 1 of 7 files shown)


mentat/session.py
```diff
@@ -245,6 +246,7 @@ async def run_main():
                 # Helps us handle errors in tests
                 if is_test_environment():
                     print(error)
+                self.error = error
```
jakethekoenig, line 249: I'd like the python client to print the session's error when it fails for easier debugging. I'm not sure this is actually the best way to accomplish that. One alternative would be to have the the above check evaluate to true for benchmark environments which would give us most of the benefit.

granawkins (approved):
LGTM

PR #423, opened 2023-12-22
Update embeddings batching based on new understanding of limits

granawkins:
The API actually allows 8192 tokes *per input*, and up to 2048 inputs, so we'll hardly ever need to send more than one batch. I opened a new repo and ran `mentat -a`, and saw it added 10MB to the embeddings database in ~3 seconds.

image


jakethekoenig (approved):
Merry Christmas!

PR #426, opened 2023-12-22
Disable sampler on error

granawkins:


jakethekoenig (approved):
Thanks for fixing this!

PR #427, opened 2023-12-27
Strip leading spaces when checking for commands

A-Joshi:
If we have a leading space on the prompt command line - e.g. ">>>      /run ls" the command fails while ">>>/run ls" works.
The reason is that the code looks for a "/" as the first character but the first character is a whitespace. Trimming the whitespace fixes the issue. 

Have only trimmed it on the left in case the commands being run need trailing whitespace  (things like xargs would use them).


jakethekoenig (approved):
Thanks for fixing this!

PR #430, opened 2023-12-28
Fixes #429 (ignore chunks with empty choices array)

### LINKED ISSUE #429: Don't choke on Azure OpenAI content filtering metadata
On Azure OpenAI, when streaming responses, the first chat completion chunk may contain an empty `choices[]` array, and instead only contains information about filtered categories:

```
{"id":"","choices":[],"created":0,"model":"","object":"","system_fingerprint":null,"prompt_filter_results":[{"prompt_index":0,"content_filter_results":{"hate":{"filtered":false,"severity":"safe"},"self_harm":{"filtered":false,"severity":"safe"},"sexual":{"filtered":false,"severity":"safe"},"violence":{"filtered":false,"severity":"safe"}}}]}
```

This is the same issue as https://github.com/microsoft/semantic-kernel/issues/3650.

Directly indexing into `chunk.choices[0]`, as the Mentat code currently does, naturally causes an `IndexError: list index out of range`.

I'll submit a PR to rectify this shortly.

laszlovandenhoek:
As described in #429, `chunk.choices` is sometimes an empty array, so we should guard accordingly. This only occurs when streaming responses, so I have left other occurrences of unguarded List indexing `choices[0]` as-is.

The documentation change is not related to this particular issue, but it does pertain to Azure OpenAI in general so I decided to include it in the PR.

### DIFF
--- a/README.md
+++ b/README.md
@@ -70,6 +70,8 @@ You'll need to have API access to GPT-4 to run Mentat. There are a few options t
 
 Mentat also works with the Azure OpenAI API. To use the Azure API, provide the `AZURE_OPENAI_ENDPOINT` (`https://<your-instance-name>.openai.azure.com/`) and `AZURE_OPENAI_KEY` environment variables instead of `OPENAI_API_KEY`.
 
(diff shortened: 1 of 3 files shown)


jakethekoenig (approved):
Thanks for fixing this! Glad we have a user using the Azure endpoint.

PR #435, opened 2024-01-01
Add descriptions and show defaults for all command line parameters

A-Joshi:
Not all the commandline options had descriptions. 
Also the defaults were not showing up in the documentation - fix that.

### DIFF
--- a/mentat/config.py
+++ b/mentat/config.py
@@ -35,15 +35,22 @@ class Config:
     # Model specific settings
     model: str = attr.field(
         default="gpt-4-1106-preview",
-        metadata={"auto_completions": list(known_models.keys()),
-                  "description" : "Choose from one of the supported models: " + ",".join(list(known_models.keys()))
-                  },
+        metadata={
+            "auto_completions": list(known_models.keys()),
+            "description": "Choose from one of the supported models: " + ",".join(
+                list(known_models.keys())
+            ),
+        },
     )
     feature_selection_model: str = attr.field(
         default="gpt-4-1106-preview",
-        metadata={"auto_completions": list(known_models.keys()),
-                  "description" : "Choose from one of the supported feature selection models: " + ",".join(list(known_models.keys()))
-                  },
+        metadata={
+            "auto_completions": list(known_models.keys()),
+            "description": (
+                "Choose from one of the supported feature selection models: "
+                + ",".join(list(known_models.keys()))
+            ),
+        },
     )
     embedding_model: str = attr.field(
         default="text-embedding-ada-002",
@@ -51,14 +58,29 @@ class Config:
             "auto_completions": [
                 model.name for model in known_models.values() if model.embedding_model
             ],
-                  "description" : "Choose from one of the supported embedding models: " + ",".join([model.name for model in known_models.values() if model.embedding_model])
+            "description": (
+                "Choose from one of the supported embedding models: "
+                + ",".join(
+                    [
+                        model.name
+                        for model in known_models.values()
+                        if model.embedding_model
+                    ]
+                )
+            ),
         },
     )
     temperature: float = attr.field(
-        default=0.2, converter=float, validator=[validators.le(1), validators.ge(0)],
+        default=0.2,
+        converter=float,
+        validator=[validators.le(1), validators.ge(0)],
         metadata={
-            "description": "Temprature controls the model variability from the context: higher value implies more variability. Teemprature is a floating point number between 0 and 1. "
-        }
+            "description": (
+                "Temprature controls the model variability from the context: higher"
+                " value implies more variability. Teemprature is a floating point"
+                " number between 0 and 1. "
+            )
+        },
     )
 
     maximum_context: int | None = attr.field(
--- a/mentat/config.py
+++ b/mentat/config.py
@@ -78,9 +78,8 @@ class Config:
         validator=[validators.le(1), validators.ge(0)],
         metadata={
             "description": (
-                "The temperature setting used by model. Temprature is a floating point"
-                " number that controls the model variability from the context: higher"
-                " value implies more variability. number between 0 and 1. "
+                "The model's temperature. Temprature is a number between 0 to 1"
+                "  that controls the model variability."
             )
         },
     )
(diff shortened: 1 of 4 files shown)


mentat/config.py
```diff
@@ -35,22 +35,52 @@ class Config:
     # Model specific settings
     model: str = attr.field(
         default="gpt-4-1106-preview",
-        metadata={"auto_completions": list(known_models.keys())},
+        metadata={
+            "auto_completions": list(known_models.keys()),
+            "description": "Choose from one of the supported models: " + ",".join(
+                list(known_models.keys())
+            ),
```
jakethekoenig, line 42: I'd prefer not to put the models in the description. This description will make sense in the command line but not in the sphinx documentation. I prefer "The LLM used to respond to user requests and make edits."

mentat/config.py
```diff
@@ -35,22 +35,52 @@ class Config:
...
+        },
     )
     feature_selection_model: str = attr.field(
         default="gpt-4-1106-preview",
-        metadata={"auto_completions": list(known_models.keys())},
+        metadata={
+            "auto_completions": list(known_models.keys()),
+            "description": (
+                "Choose from one of the supported feature selection models: "
```
jakethekoenig, line 50: I prefer "The model used to prune the features before sending to the main model for edits. Only used if `llm_feature_filter` is greater than 0."

mentat/config.py
```diff
@@ -35,22 +35,52 @@ class Config:
...
     )
     embedding_model: str = attr.field(
         default="text-embedding-ada-002",
         metadata={
             "auto_completions": [
                 model.name for model in known_models.values() if model.embedding_model
-            ]
+            ],
+            "description": (
+                "Choose from one of the supported embedding models: "
```
jakethekoenig, line 62: I prefer "The model used to create embeddings for auto context".

mentat/config.py
```diff
@@ -35,22 +35,52 @@ class Config:
...
         },
     )
     temperature: float = attr.field(
-        default=0.2, converter=float, validator=[validators.le(1), validators.ge(0)]
+        default=0.2,
+        converter=float,
+        validator=[validators.le(1), validators.ge(0)],
+        metadata={
+            "description": (
+                "Temprature controls the model variability from the context: higher"
```
jakethekoenig, line 79: I prefer "The temperature setting used by `model`".

jakethekoenig (commented):
I have some thoughts on what the descriptions should be. But it mostly looks good.

mentat/config.py
A-Joshi: Description sounds good. Models, I feel, can be in the description since they are hard-coded in the code rather than being driven by a configuration.

mentat/config.py
A-Joshi: Will change the language

mentat/config.py
PCSwingle: I agree with @jakethekoenig on this one; I would prefer not to put the list of models in the description. If argument parser has a way to add autocomplete, I would be ok with that.

mentat/config.py
```diff
@@ -35,22 +35,54 @@ class Config:
...
     feature_selection_model: str = attr.field(
         default="gpt-4-1106-preview",
-        metadata={"auto_completions": list(known_models.keys())},
+        metadata={
+            "auto_completions": list(known_models.keys()),
+            "description": (
+                "The model used to prune the features before sending to the main model"
+                " for edits. Only used if llm_feature_filter is greater than 0."
+                " Possible models are: "
+                + ", ".join(list(known_models.keys()))
```
PCSwingle, line 53: Again, would prefer not to have model names in description.

mentat/config.py
```diff
@@ -35,22 +35,54 @@ class Config:
...
     temperature: float = attr.field(
-        default=0.2, converter=float, validator=[validators.le(1), validators.ge(0)]
+        default=0.2,
+        converter=float,
+        validator=[validators.le(1), validators.ge(0)],
+        metadata={
+            "description": (
+                "The temperature setting used by model. Temprature is a floating point"
+                " number that controls the model variability from the context: higher"
+                " value implies more variability. number between 0 and 1. "
```
PCSwingle, line 83:
```suggestion
                "The model's temperature. Temperature is a number between 0 to 1"
                " that controls the model variability."
```

A-Joshi: Like the re-phrasing will change.

mentat/config.py
A-Joshi: Curious why the preference against listing the supported models? Its a common approach to list the supported "items" - e.g. pandoc documentation lists the supported file formats, etc. It also helps when using the options on the commandline and selecting a model without having to run the app first or read the code.

Perhaps better formatting on the list of models? The list is exactly the same as the autocomplete. In the case of generated documentation there is typically no autocomplete (at most you might have some generated javascript with the list hidden and then displayed for html version of the documentation).

PCSwingle: For one, there are going to be a lot of available models and I think it would bloat the description way too much. I'm working on switching us to a new library right now that can support models from any provider easily, and I don't want 10 or 20 lines of models in the descriptions.

mentat/config.py
```diff
@@ -35,22 +35,53 @@ class Config:
...
         },
     )
     temperature: float = attr.field(
-        default=0.2, converter=float, validator=[validators.le(1), validators.ge(0)]
+        default=0.2,
+        converter=float,
+        validator=[validators.le(1), validators.ge(0)],
+        metadata={
+            "description": (
+                "The model's temperature. Temprature is a number between 0 to 1"
```
PCSwingle, line 81:
```suggestion
                "The model's temperature. Temperature is a number between 0 to 1"
```

mentat/config.py
A-Joshi: ah missed this one - sounds good.

mentat/config.py
A-Joshi: I was going to suggest moving to a lib like llmware that had a lot of other support (like vetors, local models etc.). I would change the description from the hardcoded list (which will hopefully e gone with the change) to one referencing the lib at that point.

PR #441, opened 2024-01-02
Fix autocomplete crash bug

### LINKED ISSUE #424: Autocomplete exception
image
Steps to reproduce: type "/" and then a space. Only gives an exception one time. Autocompletes don't work for the rest of the session.

PCSwingle:
Fixes #424


biobootloader (approved):

jakethekoenig (approved):
Works for me. We should probably make some tests for the auto_completer

PCSwingle:
> Works for me. We should probably make some tests for the auto_completer

Agreed; I'll make an issue for this and probably work on it soon.

PR #444, opened 2024-01-02
UI test fix

jakethekoenig:


PCSwingle (approved):
Thanks! This was an idea I had early on for a test we could manually run every once in a while that would check all sorts of scenarios to make sure it looks good (i.e., make sure we didn't mess up how interactive mode looked, or commands, etc.) but I never really got around to doing much with it. If we decide to put some time into it I can see it being more useful in the future, but we'll see.

PR #446, opened 2024-01-03
First part of theme support

A-Joshi:
The PR setups the ability to handle themes with different color palettes.
Currently bulk of the messages use the stream output with hardcoded colors. The prompt text coloring is handled separately. There is also a small set of files that use colors for output using a function (colored) with colors directly.
The PR add a new theme option and adds support for using the theme for messages using the stream output. The other two cases (prompt and colored functions) are not handled by the PR but will come later.

Split the work to make reviewing easier - this PR affects a lot of files but the change is relatively benign and small.


mentat/terminal/output.py
```diff
@@ -29,13 +31,22 @@ def print_stream_message(message: StreamMessage):
     end = "\n"
     color = None
     flush = False
+    session_context = SESSION_CONTEXT.get()
```
jakethekoenig, line 34: I think we want to avoid using `SESSION_CONTEXT` on the clients. The terminal holds the config which has the theme. Can you get it from there instead please?

mentat/terminal/output.py
```diff
@@ -29,13 +31,22 @@ def print_stream_message(message: StreamMessage):
...
+    if session_context.config.theme is not None:
```
jakethekoenig, line 35: I prefer `Themes.get(config.theme)` to the if statement. It will be `None` if `config.theme` is anyway.

mentat/terminal/output.py
```diff
@@ -29,13 +31,22 @@ def print_stream_message(message: StreamMessage):
...
+    else:
+        theme = None
     if message.extra:
         if isinstance(message.extra.get("end"), str):
             end = message.extra["end"]
         if isinstance(message.extra.get("color"), str):
             color = message.extra["color"]
         if isinstance(message.extra.get("flush"), bool):
             flush = message.extra["flush"]
+        if isinstance(message.extra.get("style"), str):
```
jakethekoenig, line 46: Maybe delete the if statement where we check if color is included as it is no longer sent.

jakethekoenig (commented):
It looks pretty good to me outside the comments. Definitely an improvement. Thanks so much!

mentat/terminal/output.py
PCSwingle: Well we might end up sending it in the future; I think it still makes sense to keep.

PCSwingle (commented):
Just looked over this and it looks great! Agree with Jake on almost all of his comments, but once those are resolved this should be perfect.

mentat/config.py
```diff
@@ -146,6 +146,15 @@ class Config:
         },
     )
 
+    theme: str | None = attr.field(
+        default="light",
+        metadata={
+            "description": (
+                "Theme for interaction possible choices are light, dark or none."
+            )
+        },
```
granawkins, line 155:
```suggestion
            ),
            validator=validators.in_(["light", "dark"]),
        },
```
You might import Themes here and use Themes.keys() instead, but idk if there's a circular dependency..

mentat/terminal/output.py
jakethekoenig: I guess it's fine to keep. I kind of think it will be a mistake for the session to ever have an opinion about the colors the client sends after this PR though.

mentat/streaming_printer.py
```diff
@@ -13,7 +14,22 @@ def __init__(self):
...
+    ):
+        if style is not None:
+            session_context = SESSION_CONTEXT.get()
+            if session_context.config.theme is not None:
+                theme = Themes.get(session_context.config.theme)
+            else:
+                theme = None
+            if theme is not None:
+                color = theme[style]
+
```
granawkins, line 32: Could we just pass on the style here instead of converting to the color?

Looks like these strings end up being passed to `stream.send` eventually, which should take care of this.

granawkins (commented):
Good idea! I've thought of this before. Glad to have it.

mentat/terminal/themes.py
```diff
@@ -0,0 +1,36 @@
...
+#        "input": "light_blue",
+#        "success": "light_green",
+#        "error": "light_red",
+#        "warning": "light_yellow",
+#        "error": "red",
+#        "warning": "yellow",
+
+CurrentTheme = "dark"
+
+Theme = Themes[CurrentTheme]
```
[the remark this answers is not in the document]
A-Joshi: Removed temp code.

mentat/terminal/output.py
A-Joshi: Redid getting the theme based on the suggestion and added a validation check on the theme name.

mentat/terminal/output.py
A-Joshi: I have left it there in case I missed a spot. Can remove it once we have confirmed that all is well.

mentat/streaming_printer.py
A-Joshi: We could but the next function (in the same file) seems to be very low level tied to colors.

mentat/config.py
A-Joshi: Thought about it - was lazy. There should be no dependency - Themes are a simple set of dicts.

mentat/terminal/output.py
A-Joshi: removed

A-Joshi: sorry for the wrong one.

mentat/config.py
```diff
@@ -146,13 +147,14 @@ class Config:
...
 
-    theme: str | None = attr.field(
+    theme: str | None = attr.field(  # pyright: ignore
         default="light",
         metadata={
             "description": (
                 "Theme for interaction possible choices are light, dark or none."
             )
         },
+        validator=validators.in_(Themes.keys.append(None)),  # pyright: ignore
```
PCSwingle, line 157: Why is there an append None? And I don't think append returns a value, so I don't think this works.

mentat/terminal/themes.py
```diff
@@ -1,4 +1,4 @@
-Themes = {
+Themes: dict[str | None, dict[str, str]] = {
```
PCSwingle, line 1: I would prefer if this was lowercase; I keep expecting Themes to be a class

mentat/config.py
A-Joshi: fixed.

mentat/terminal/themes.py
A-Joshi: changed

PCSwingle (approved):
LGTM :rocket:

PR #448, opened 2024-01-03
Viewer share upgrades

jakethekoenig:
There is now a feedback button, share button and download button. Each with helpful tooltips. The feedback button no longer copies the s3 link to your clipboard. They have a consistent style in the upper right.


granawkins (approved):
Look fine, seems to be working 👍 

I tried submitting some feedback but I'm not sure if/how I can view the s3 bucket, so I haven't double-checked but I'm sure its fine.

jakethekoenig:
> Look fine, seems to be working 👍
> 
> I tried submitting some feedback but I'm not sure if/how I can view the s3 bucket, so I haven't double-checked but I'm sure its fine.

Oops there was a problem with the feedback button. It's fixed now. You should be able to see the s3 bucket in the AWS console. When a user clicks feedback we should also get pinged in the dogfooding slack channel.

PR #450, opened 2024-01-04
Add litellm

PCSwingle:
TODO: Add documentation saying to check litellm to see how to use other models.

## Pull Request Checklist

### DIFF
--- a/README.md
+++ b/README.md
@@ -58,9 +58,13 @@ cd mentat
 pip install -e .
 ```
 
-## Add your OpenAI API Key
+## Selecting which LLM Model to use
 
-You'll need to have API access to GPT-4 to run Mentat. There are a few options to provide Mentat with your OpenAI API key:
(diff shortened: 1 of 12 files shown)


jakethekoenig:
I'm having some trouble getting it to work with ollama. We need to set `litellm.drop_params` to true because the ollama api doesn't support `response_format`. And we also need to pass in an `api_base` to the litellm completions. After I did that though I got the following exception which may be a bug with litellm itself? Not sure.
```
  File "/Users/jakekoenig/mentat/.venv/lib/python3.11/site-packages/litellm/llms/ollama.py", line 260, in ollama_async_streaming
    status_code=response.status_code, message=response.text
                                              ^^^^^^^^^^^^^
  File "/Users/jakekoenig/mentat/.venv/lib/python3.11/site-packages/httpx/_models.py", line 574, in text
    content = self.content
              ^^^^^^^^^^^^
  File "/Users/jakekoenig/mentat/.venv/lib/python3.11/site-packages/httpx/_models.py", line 568, in content
    raise ResponseNotRead()
httpx.ResponseNotRead: Attempted to access streaming response content, without having called `read()`.
```

PCSwingle:
Replaced by #451.

PR #451, opened 2024-01-04
Add documentation for using other models, make it easier to use non OpenAI models

PCSwingle:
Replaces #450. After discussing with @jakethekoenig, we feel this is a better alternative than using litellm directly in the codebase.

## Pull Request Checklist

### DIFF
--- a/README.md
+++ b/README.md
@@ -58,7 +58,11 @@ cd mentat
 pip install -e .
 ```
 
-## Add your OpenAI API Key
+## Selecting which LLM Model to use
+
+We highly recommend using the default model, `gpt-4-1106-preview`, as it performs vastly better than any other model benchmarked so far. However, if you wish to use a different model, jump [here](#alternative-models).
+
+### Add your OpenAI API Key
 
 You'll need to have API access to GPT-4 to run Mentat. There are a few options to provide Mentat with your OpenAI API key:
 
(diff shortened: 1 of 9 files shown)


mentat/code_context.py
```diff
@@ -148,12 +148,13 @@ async def get_code_message(
             "\n".join(include_files_message), model, full_message=False
         )
 
-        tokens_used = (
-            prompt_tokens + meta_tokens + include_files_tokens + config.token_buffer
-        )
+        tokens_used = prompt_tokens + meta_tokens + include_files_tokens
         if not is_context_sufficient(tokens_used):
-            raise ContextSizeInsufficient()
```
jakethekoenig, line 155: Should we print something here so the user knows what the problem is?

README.md
```diff
@@ -70,9 +74,35 @@ You'll need to have API access to GPT-4 to run Mentat. There are a few options t
 
 Mentat also works with the Azure OpenAI API. To use the Azure API, provide the `AZURE_OPENAI_ENDPOINT` (`https://<your-instance-name>.openai.azure.com/`) and `AZURE_OPENAI_KEY` environment variables instead of `OPENAI_API_KEY`.
 
-In addition, Mentat uses the `gpt-4-1106-preview` by default. On Azure, this model is available under a different name: `gpt-4-1106-Preview` (with a capital P). To use it, override the default model as described in [configuration.md](docs/configuration.md).
+In addition, Mentat uses the `gpt-4-1106-preview` model by default. When using Azure, you will have to set the model as described in [configuration.md](docs/configuration.md) to the name you gave your Azure model.
```
jakethekoenig, line 77: We should duplicate this to the user guide and remove it from the user guide when that's hosted. This is good for now though.

jakethekoenig (approved):
Looks good to me. Thanks!

mentat/code_context.py
PCSwingle: We do print something in is_context_sufficient; maybe we should move the warning out of the function, but we do use it in multiple places.

PR #453, opened 2024-01-05
change default format to replacement and also remove unneeded part in…

PCSwingle:
… block format prompt

## Pull Request Checklist


jakethekoenig:
Triggered the benchmark action for this branch: https://github.com/AbanteAI/mentat/actions/runs/7424919093

PCSwingle:
Not going to merge yet; would like more feedback from using Mentat on real projects (not benchmarks) to make sure this format performs just as well as the block format.

PCSwingle:
I've been working with replacement format on a bit over the past week or 2 and I've found that it feels a lot worse than the block format (even if the benchmark scores are similar), so I'm closing this.

PR #454, opened 2024-01-05
Mutate sample

granawkins:
This PR refactors and adds to the Sampler utility scripts. Now we have:

- validate: check repo/context/diffs each item in your samples dir
- evaluate: run mentat on the sample and return the resulting diff
- add-context: create n duplicate but with lots of extra (useless) context
- remove-context: create n duplicates but missing some context, and with a rejection message (interactive)
- finetune: generate finetuning examples from Samples.

Run them with e.g. `python scripts/sampler -v`, `python scripts/sampler --add-context -n 10`


scripts/sampler/__main__.py
```diff
@@ -0,0 +1,163 @@
...
+                logs.append({"id": new_sample.id, "prototype_id": sample.id})
+            except Exception as e:
+                warn(f"Error adding extra context to sample {sample.id}: {e}")
+        elif args.remove_context:
+            if not sample.context or len(sample.context) == 1:
+                warn(f"Skipping {sample.id}: no context to remove.")
+                continue
+            try:
+                new_sample = await remove_context(sample)
+                new_sample.save(SAMPLES_DIR / f"sample_{new_sample.id}.json")
```
jakethekoenig, line 122: Same comment here and above. Can we print the files generated?

scripts/sampler/__main__.py
```diff
@@ -0,0 +1,163 @@
...
+
+from add_context import add_context
+from evaluate import evaluate_sample
+from finetune_gpt import generate_finetune_gpt
+from remove_context import remove_context
+from validate import validate_sample
+
+from mentat.sampler.sample import Sample
+from mentat.utils import mentat_dir_path
+from tests.benchmarks.benchmark_runner import (
```
jakethekoenig, line 18: I couldn't run this until I added `tests` to `find_packages` in setup.py. Btw what do you think about moving the benchmark code out of the tests directory into its own top level directory benchmarking?

scripts/sampler/remove_context.py
```diff
@@ -0,0 +1,134 @@
...
+
+    # Setup the repo and load context files
+    repo = setup_repo(
+        url=sample.repo,
+        commit=sample.merge_base,
+        diff_merge_base=sample.diff_merge_base,
+        diff_active=sample.diff_active,
+    )
+    cwd = Path(repo.working_dir)
+    python_client = PythonClient(cwd=Path("."), paths=[])
```
jakethekoenig, line 30: The python client needs to be shut down.

scripts/sampler/__main__.py
```diff
@@ -0,0 +1,163 @@
...
+            try:
+                new_sample = await remove_context(sample)
+                new_sample.save(SAMPLES_DIR / f"sample_{new_sample.id}.json")
+                logs.append({"id": new_sample.id, "prototype_id": sample.id})
+            except Exception as e:
+                warn(f"Error removing context from sample {sample.id}: {e}")
+        else:
+            print(f"Evaluating sample {sample.id[:8]}")
+            print(f"  Prompt: {sample.message_prompt}")
+            diff_eval = await evaluate_sample(sample)
```
jakethekoenig, line 129: To me `evaluate_sample` would imply running and grading the sample. Do you think `run_sample` or `execute_sample` would be better?

scripts/sampler/evaluate.py
```diff
@@ -0,0 +1,52 @@
...
+)
+
+from mentat.errors import SampleError
+from mentat.git_handler import get_git_diff
+from mentat.python_client.client import PythonClient
+from mentat.sampler.utils import get_active_snapshot_commit, setup_repo
+from mentat.session_context import SESSION_CONTEXT
+
+
+async def evaluate_sample(sample, cwd: Path | str | None = None):
```
jakethekoenig, line 15: What do you think about making this one an instance method of samples? I'd like to use it in the benchmark runner itself and it'd be easy to get there.

scripts/sampler/__main__.py
```diff
@@ -0,0 +1,163 @@
...
+            )
+            print(f"[{sample.id[:8]}] {sample.title}: {status}")
+            logs.append({"id": sample.id, "is_valid": is_valid, "reason": reason})
+        elif args.finetune:
+            print(f"Generating fine-tuning example for sample {sample.id[:8]}")
+            try:
+                example = await generate_finetune_gpt(sample)
+                example_file = FINETUNE_DIR / f"finetune_{sample.id}.json"
+                with open(example_file, "w") as f:
+                    json.dump(example, f, indent=4)
```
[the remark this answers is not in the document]
granawkins: I've just finally read the actual difference between `.json` and `.jsonl` - 

I think your interpretation was actually correct for jsonl: each line should be independently json-serializable. I guess theoretically you'd be able to read-in and deserialize one at a time, making it better for extremely large runs where the whole thing is too big to read into memory.

I guess I'm more used to working with `.json` and don't expect we'll hit that limit soon, right? Any other thoughts?

scripts/sampler/__main__.py
```diff
@@ -0,0 +1,163 @@
...
+                if is_valid
+                else f"\033[91mFAILED: {reason}\033[0m"
+            )
+            print(f"[{sample.id[:8]}] {sample.title}: {status}")
+            logs.append({"id": sample.id, "is_valid": is_valid, "reason": reason})
+        elif args.finetune:
+            print(f"Generating fine-tuning example for sample {sample.id[:8]}")
+            try:
+                example = await generate_finetune_gpt(sample)
+                example_file = FINETUNE_DIR / f"finetune_{sample.id}.json"
```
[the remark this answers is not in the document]
granawkins: I'll print a the FINETUNE_DIR path at the end with the results.

scripts/sampler/__main__.py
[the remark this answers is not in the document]
jakethekoenig: No other thoughts.

scripts/sampler/__main__.py
granawkins: Ah makes sense, I'll add that to setup.py.

Ya I like the idea of moving benchmarks to their own directory, mostly because I always forget the `--benchmark` flag.

PR #455, opened 2024-01-05
Samples can be run by the benchmark runner

jakethekoenig:
If you put a sample in the tests/benchmarks/benchmarks directory it will be picked up and evaluated by the benchmark runner. Test with:
```
pytest tests/benchmarks/benchmark_runner.py --benchmark -s --benchmarks \
Dummy clojure
```
Based off PR #454 to make use of setup_repo. There's some duplication of sample evaluation.

## Pull Request Checklist

### DIFF
--- a/tests/benchmarks/benchmark_runner.py
+++ b/tests/benchmarks/benchmark_runner.py
@@ -5,14 +5,22 @@ import re
 from pathlib import Path
 
 import pytest
+from openai.types.chat import (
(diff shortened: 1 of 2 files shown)


granawkins (approved):
LGTM, the test worked fine locally.

Still some overlap as you say, but this is a step in the right direction.

jakethekoenig:
Sorry for the messy history. Decided it was best to cherry-pick the one commit and force push

PR #460, opened 2024-01-08
Use ChromaDB for our embeddings database and similarity search

granawkins:


mentat/embeddings.py
```diff
@@ -1,77 +1,107 @@
...
-database = EmbeddingsDatabase()
+    def __init__(self, embedding_model: str):
+        api_key = os.getenv("OPENAI_API_KEY")
+        # src: https://cookbook.openai.com/examples/vector_databases/chroma/using_chroma_for_embeddings_search
+        embedding_function = OpenAIEmbeddingFunction(
+            api_key=api_key, model_name=embedding_model
+        )
+        self._collection = client.get_or_create_collection(
+            name=f"mentat-{embedding_model}",
+            embedding_function=embedding_function,  # type: ignore
```
granawkins, line 28: The typing issue here and below is related to the OpenAIEmbeddingFunction, which is their in-house recommended approach but isn't setup to handle images (which the rest of them are).

mentat/feature_filters/embedding_similarity_filter.py
```diff
@@ -19,7 +19,7 @@ async def score(
             self.query, features, self.loading_multiplier
         )
         features_scored = zip(features, sim_scores)
-        return sorted(features_scored, key=lambda x: x[1], reverse=True)
+        return sorted(features_scored, key=lambda x: x[1])
```
granawkins, line 22: We were using cosine_similarity, which is larger-is-better, but Chroma's default similarity func is 'l2', which is lower-is-better.

tests/conftest.py
```diff
@@ -220,17 +220,6 @@ def set_unstreamed_values(value):
...
-def mock_call_embedding_api(mocker):
-    embedding_mock = mocker.patch.object(LlmApiHandler, "call_embedding_api")
-
-    def set_embedding_values(value):
-        embedding_mock.return_value = value
-
-    embedding_mock.set_embedding_values = set_embedding_values
-    return embedding_mock
-
-
```
granawkins, line 233: I've removed this and embeddings_test.py, because the behavior we were actually testing (properly measuring similarity) is now handled by chroma.

tests/license_check.py
```diff
@@ -9,6 +9,12 @@
     "tiktoken",
     # openai as Apache 2.0; for some reason, after updating to 1.0, pip-licenses thinks it's UNKNOWN
     "openai",
+    # UNKNOWN
+    "chroma-hnswlib",
+    "pyinstaller-hooks-contrib",
+    # "Other/Proprietary License"
+    "pinecone-client",  # part of chromadb - we don't use directly, but potentially problematic
+    "pygls",
```
granawkins, line 17: Kindof a mess. Thoughts?

PCSwingle: As long as we're compatible with all of their licenses it's fine; have you checked the licenses on their github? We mainly just need to make sure none of them have copyleft licenses like GPL

tests/license_check.py
```diff
@@ -26,6 +32,19 @@
     "GNU General Public License (GPL)",
     "Public Domain",
     "The Unlicense (Unlicense)",
+    "Apache License, Version 2.0",
+    "Apache License v2.0",
+    "Apache License 2.0",
+    "GNU Lesser General Public License v2 or later (LGPLv2+)",
```
PCSwingle, line 38: Which library has this license? @biobootloader are we ok with using this library?

mentat/embeddings.py
PCSwingle, line 28: I don't know how chromadb works, but will it be calling this embedding function? I also don't like that we're setting the API key in here. I would prefer to leave all of this stuff to the llm api handler; how about we just pass in the call_embedding_api function as the embedding function (that way we get the api_guard decorator and everything is handled in the right place)

tests/conftest.py
PCSwingle: We're still gonna need this mock, because otherwise we can never use call_embedding_api in tests (since it has an api_guard on it)

mentat/embeddings.py
granawkins: Aha, good idea ya I think I can do that.

tests/license_check.py
granawkins: [chardlet](https://github.com/chardet/chardet) is the library.

tests/conftest.py
granawkins: As it's setup we wouldn't ever need to, so that func could be removed as well.

But in light of your comment below this will change

granawkins:
Now we're passing our `llm_api_handler.call_embeddings_api` directly to ChromaDB. Only complication was that it takes a non-async func.

tests/license_check.py
biobootloader: looks like it should be fine!

mentat/embeddings.py
```diff
@@ -1,77 +1,124 @@
...
+        """Temporary helper function to migrate sqlite3 to chromadb
+
+        Prior to January 2024, embeddings were fetched directly from the OpenAI API in
+        batches and saved to a db. We're currently using the same embeddings (ada-2) with
+        ChromaDB, so we might as well save the effort of re-fetching them. One drawback
+        is that ChromaDB saves the actual text, while our old schema did not, so migrated
+        records will have an empty documents field. This shouldn't be a problem. If it is,
+        we can just update the 'exists' method to require a non-empty "document" field.
+
+        TODO: erase this method/call after a few months
```
biobootloader, line 90: this is a nice optimization, but I'm not sure the extra complexity is worth saving users a couple cents to rebuild. It might be unlikely, but some bug from migrated records being slightly different could be a big headache and not worth the risk

mentat/llm_api_handler.py
```diff
@@ -356,13 +383,14 @@ async def call_llm_api(
         return response
 
     @api_guard
-    async def call_embedding_api(
+    def call_embedding_api(
```
biobootloader, line 386: hmm a little too bad Chroma doesn't support async. It's probably not an issue now but in the future if were doing embeddings in the background it could make mentat unresponsive

tests/license_check.py
```diff
@@ -26,6 +32,19 @@
...
+    "ISC",
+    "LGPL-2.1-only",
+    "GNU General Public License v3 (GPLv3)",
```
biobootloader, line 41: hmm could be an issue? 
image

mentat/embeddings.py
granawkins: Ok make sense. I was just looking at the size of my sqlite file but (1) mine is probably bigger than anyone's and (2) it is indeed really fast and cheap to reload.

granawkins:
I messed up - I was running license_check locally outside a venv, so it was scanning my entire pip library. Chroma only requires is 1 skip/UNKNOWN, and some different wordings of the Apache license.

mentat/llm_api_handler.py
granawkins: Look like some other people on Discord also want async, but not dice yet: https://discord.com/channels/1073293645303795742/1073293646083919946/1191885640120414218

biobootloader (approved):

PR #463, opened 2024-01-09
Config settings documented

jakethekoenig:
How to change config settings and the available settings are added to the sphinx documentation. A new page on how to use alternative models is added.

## Pull Request Checklist


PCSwingle (approved):
:rocket:

PR #465, opened 2024-01-10
Readme shorted. Docs referenced

jakethekoenig:
## Pull Request Checklist

### DIFF
--- a/README.md
+++ b/README.md
@@ -61,7 +61,7 @@ pip install -e .
 
 ### Add your OpenAI API Key
 
-You'll need to have API access to GPT-4 to run Mentat. There are a few options to provide Mentat with your OpenAI API key:
+There are a few options to provide Mentat with your OpenAI API key:
 
 1. Create a `.env` file with the line `OPENAI_API_KEY=<your-api-key>` in the directory you plan to run mentat in or in `~/.mentat/.env`
 2. Run `export OPENAI_API_KEY=<your key here>` prior to running Mentat
--- a/README.md
+++ b/README.md
@@ -1,5 +1,6 @@
 [![Twitter Follow](https://img.shields.io/twitter/follow/bio_bootloader?style=social)](https://twitter.com/bio_bootloader)
 [![Discord Follow](https://dcbadge.vercel.app/api/server/XbPdxAMJte?style=flat)](https://discord.gg/zbvd9qx9Pb)
+[![Documentation Status](https://readthedocs.org/projects/mentat-ai/badge/?version=latest)](https://docs.mentat.ai/en/latest/?badge=latest)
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -173,4 +89,4 @@ The goal for Mentat is to become the best way to write code. Some big improvemen
 
 If this is something you'd like to work on, jump right in! And if you want to join the team to work on this full time, message [@bio_bootloader](https://twitter.com/bio_bootloader) on twitter.
 
-To find something specific to work on, take a look at [open issues](https://github.com/AbanteAI/mentat/issues) and/or check this Github Project: [Mentat Development](https://github.com/users/biobootloader/projects/2)
+To find something specific to work on, take a look at [open issues](https://github.com/AbanteAI/mentat/issues).
```
jakethekoenig, line 92: It doesn't seem like we really use the project so I thought it was better to remove the link.

README.md
```diff
@@ -58,10 +59,6 @@ cd mentat
 pip install -e .
 ```
 
-## Selecting which LLM Model to use
-
-We highly recommend using the default model, `gpt-4-1106-preview`, as it performs vastly better than any other model benchmarked so far. However, if you wish to use a different model, jump [here](#alternative-models).
-
 ### Add your OpenAI API Key
 
 You'll need to have API access to GPT-4 to run Mentat. There are a few options to provide Mentat with your OpenAI API key:
```
PCSwingle, line 64: Maybe this should be changed; everybody has gpt-4 access now, and it isn't even necessary anymore.

README.md
```diff
@@ -70,66 +67,6 @@ You'll need to have API access to GPT-4 to run Mentat. There are a few options t
 2. Run `export OPENAI_API_KEY=<your key here>` prior to running Mentat
 3. Place the previous command in your `.bashrc` or `.zshrc` to export your key on every terminal startup
 
-### Azure OpenAI
```
PCSwingle, line 73: Is any of this information anywhere else? I think it's important to keep somewhere

jakethekoenig: It should all be here: https://docs.mentat.ai/en/latest/user/alternative_models.html

README.md
```diff
@@ -1,5 +1,6 @@
 [![Twitter Follow](https://img.shields.io/twitter/follow/bio_bootloader?style=social)](https://twitter.com/bio_bootloader)
```
biobootloader, line 1:
```suggestion
[![Twitter Follow](https://img.shields.io/twitter/follow/AbanteAI?style=social)](https://twitter.com/AbanteAI)
```

biobootloader (approved):

README.md
PCSwingle: Could we link here here? I think it's important that it's easy for people to find the docs for alternative models

PR #466, opened 2024-01-10
add amend command, allow non command autocompleted prompts

### LINKED ISSUE #462: Add new command: /retry or /amend
I'm not sure the right name for this command, something like `/retry`, `/amend`, or `/go-back`

You would run it after Mentat has given you a bad set of edits or failed to understand what you wanted. Instead of leaving your original prompt and it's bad response in context, you'd run the command, which would remove both those message from history and prefill your prompt with your last prompt, allowing you to amend it and run again.

PCSwingle:
Fixes #462. Adds amend command and adds kwarg for having command autocomplete on a prompt.
## Pull Request Checklist


granawkins (approved):
LGTM - performs as expected and really useful to have.

1 question: I couldn't parse exactly when/why `command_autocomplete` would be False?

1 comment: In the future I wonder if it'd make sense to build our own wrapper around the ChatCompletionMessage, which could include a `get_text_content` method, and also track message metadata for history/sampler like the latest commitID, the git diff since the last user message (this would simplify `undo` a lot), etc.

PCSwingle:
We don't want it to be true when asking y/n questions, when asking y/n/i, etc. We only want it to be true when we're actually checking for commands.

biobootloader (approved):
Nice, works well!

Before merging make sure to update `docs/source/user/commands.rst`

mentat/command/commands/amend.py
```diff
@@ -0,0 +1,38 @@
...
+    def argument_autocompletions(
+        cls, arguments: list[str], argument_position: int
+    ) -> list[str]:
+        return []
+
+    @override
+    @classmethod
+    def help_message(cls) -> str:
+        return (
+            "Removes all context since last user request. Will not undo any edits made."
```
biobootloader, line 37:
```suggestion
            "Used to amend a previous user request. Works by reseting contex to the state it was at the last request and prefills user input with the last request. Does not undo any edits."
```

PR #467, opened 2024-01-10
fix returntouser being thrown outside try except bug

PCSwingle:
## Pull Request Checklist

### DIFF
--- a/mentat/code_context.py
+++ b/mentat/code_context.py
@@ -110,6 +110,7 @@ class CodeContext:
         prompt: Optional[str] = None,
         expected_edits: Optional[list[str]] = None,  # for training/benchmarking
         loading_multiplier: float = 0.0,
+        context_guard: bool = True,
     ) -> str:
         """
         Retrieves the current code message.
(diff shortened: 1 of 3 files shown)


mentat/conversation.py
```diff
@@ -82,7 +83,7 @@ async def display_token_count(self):
             }
             stream.send(
                 f"Included file{message[(_plural, _exceed)]} token limit"
-                f" ({tokens} / {context_size}). Truncating based on task similarity.",
+                f" ({tokens} / {context_size}).",
```
biobootloader, line 86: what was this?

mentat/code_context.py
```diff
@@ -149,7 +150,7 @@ async def get_code_message(
         )
 
         tokens_used = prompt_tokens + meta_tokens + include_files_tokens
-        if not is_context_sufficient(tokens_used):
+        if context_guard and not is_context_sufficient(tokens_used):
```
biobootloader, line 153: `is_context_sufficient` should probably be renamed, it sounds like it's asking if context is enough, not checking that it doesn't exceed max.

Looks like the only places it's called, the calling function just raises ReturnToUser if true? maybe we can move the `raise ReturnToUser` to it and rename to `raise_if_context_exceeds_max`

and we can rename `context_guard` to `suppress_context_limit_check` or something else more clear?

mentat/conversation.py
PCSwingle: We used to truncate, but we don't do that anymore. I realized it was saying this while testing this bug.

biobootloader (approved):

PR #472, opened 2024-01-11
Benchmarks moved out of test dir

jakethekoenig:
They no longer run with pytest. They are just python scripts. This is in preparation to make tests for the benchmark runners themselves that mock out just the OpenAi api.

This change also has a number of small benefits:
* The cli interface is simpler no longer requiring "pytest --benchmark -s test/"
* We don't have to worry about pytest capturing or suppressing errors or standard out.
* Some prompts wanted to be used in the scripts dir requiring tests to be a module. Not tests does not have to be one but benchmarks is.

I made sure benchmark_runner and the exercism benchmarks still work (you can test them with the commands in the readme) but some other benchmarks such as the context_benchmark seem a little broken at the moment. Which I guess shows the importance of testing.

## Pull Request Checklist


PCSwingle (approved):
Thanks for doing this! This makes it a lot cleaner

PR #475, opened 2024-01-11
add revisor using git diff

PCSwingle:
TODO: 
- [x] Add way to turn revisor on and off
- [x] Add tests
- [x] Add 'waiting for revisor...' text to make it clear what's happening

## Pull Request Checklist


mentat/revisor/revisor.py
```diff
@@ -0,0 +1,132 @@
...
+            ctx.stream.send(change_delimiter)
+            ctx.stream.send("\n".join(diff_diff))
+            ctx.stream.send(change_delimiter)
+        ctx.cost_tracker.display_last_api_call()
+
+
+async def revise_edits(file_edits: List[FileEdit]):
+    for file_edit in file_edits:
+        # We could do all edits asynchronously; of course, this risks getting rate limited and probably not worth effort
+        await revise_edit(file_edit)
```
biobootloader, line 132: it might be worth the effort since it'd reduce latency a lot (and I've actually already implemented a balancer in a different repo) But yeah not for this PR, for now let's just try to get an experimental feature, not perfected yet

biobootloader:
Doesn't seem reliable enough yet to be default-on but maybe we could merge as an experimental feature, since you've laid the groundwork for it and perhaps with more tweaks it could be good enough

biobootloader (approved):

PR #481, opened 2024-01-13
add full file_edits_to_llm_message + tests

granawkins:
This adds the missing `GitParser.file_edits_to_llm_message` method, which we'll use to convert file edits from your conversation history into git diffs for Sampler v.0.2.0.

## Pull Request Checklist

### DIFF
--- a/mentat/parsers/git_parser.py
+++ b/mentat/parsers/git_parser.py
@@ -14,8 +14,8 @@ for the LLM. We implement only those methods/properties which are necessary for 
 Things to be aware of:
 1. git diffs include SHA-1 hashes, which link the diff to a known record in the git repo. When
     we generate diffs from scratch, we won't have these, so the diff can't be reverse-applied via
-    e.g. `git apply myfile.diff`. The SHA-1 hashes should also be removed e.g. when comparing 
-    outputs in tests as they are not expected to match.
+    `git apply myfile.diff`. The SHA-1 hashes should also be removed when comparing outputs in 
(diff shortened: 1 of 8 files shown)


biobootloader:
@PCSwingle can you take a look at this

mentat/parsers/git_parser.py
```diff
@@ -62,6 +100,12 @@ def parse_string(self, git_diff: str) -> ParsedLLMResponse:
                 is_deletion=is_deletion,
                 rename_file_path=new_name,
             )
+            if not is_creation:
+                file_edit.previous_file_lines = (
```
PCSwingle, line 104: This is used specifically for undo's (when the fileedit is applied, it sets this so that the fileedit can be undone). It feels a bit weird using it here, but I guess it works.

mentat/parsers/git_parser.py
```diff
@@ -23,11 +39,33 @@ async def stream_and_parse_llm_response(
...
     ) -> ParsedLLMResponse:
-        string = ""
+        current_line = ""
+        conversation = ""
+        file_edits = ""
         async for chunk in response:
             for content in chunk_to_lines(chunk):
-                string += content
-        return self.parse_string(string)
+                # Wait for a full line
```
PCSwingle, line 47: Why are we parsing it this way? If we aren't streaming it, this just adds a bunch of complexity for no reason.

PCSwingle (approved):

PR #483, opened 2024-01-15
Benchmark_repos specified with a relative path

jakethekoenig:
The motivation for this PR is a fix for the benchmark github action. Now that the benchmarks are in a different package the aren't importing mentat code from where they did previously in the file system. This change should have no effect for people calling clone_repo from the mentat root dir.
I think in general it makes sense that the caller of setup_repo would expect the repo to be places in benchmark_repos of their cwd not relative to some script that they may not know about and may be in a strange unpredictable place depending on how mentat is installed.

## Pull Request Checklist

### DIFF
--- a/benchmarks/benchmark_runner.py
+++ b/benchmarks/benchmark_runner.py
@@ -286,7 +286,6 @@ async def run_benchmarks(retries, benchmarks):
             results.extend(await evaluate_sample(path))
 
     summary = BenchmarkResultSummary(results)
-    os.chdir("../..")
     with open("results.json", "w") as f:
         f.write(summary.to_json())
     summary.render_results()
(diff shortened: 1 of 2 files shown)


benchmarks/benchmark_runner.py
```diff
@@ -286,7 +286,6 @@ async def run_benchmarks(retries, benchmarks):
             results.extend(await evaluate_sample(path))
 
     summary = BenchmarkResultSummary(results)
-    os.chdir("../..")
```
jakethekoenig, line 289: This should have been removed when sample evaluation was added. cwd is set via argument not inferred from actual cwd. This originally changed the cwd back but now there's no need.

PR #487, opened 2024-01-16
remove unecessary example in block prompt

PCSwingle:
## Pull Request Checklist


biobootloader (approved):

biobootloader:
if anything this should increase performance. Notice any differences?

PR #488, opened 2024-01-16
Benchmark runner test

jakethekoenig:
While writing this I made the following changes:
* benchmark_repos was made a subdir of benchmarks so verify functions could import from generated code. Otherwise python complains it isn't part of a module.
* In fact setup_repo does change the current directory. evalute_py and evaluate_sample now change the directory back.
* The whole functions have a try now so spurious errors won't crash a run.
* The append action is removed from the parser.

It's a little less comprehensive than the exercism analogue but I think it'll still be valuable.

## Pull Request Checklist

### DIFF
--- a/testbed/exercism-python/exercises/practice/accumulate/accumulate_test.py
+++ b/testbed/exercism-python/exercises/practice/accumulate/accumulate_test.py
@@ -9,31 +9,33 @@ class AccumulateTest(unittest.TestCase):
 
     def test_pow(self):
         self.assertEqual(
-            accumulate([1, 2, 3, 4, 5], lambda x: x * x), [1, 4, 9, 16, 25])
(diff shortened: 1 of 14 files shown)


testbed/exercism-python/exercises/practice/accumulate/accumulate_test.py
```diff
@@ -9,31 +9,33 @@ def test_empty_sequence(self):
 
     def test_pow(self):
```
jakethekoenig, line 10: We can't let the actual accumulate get formatted because it could make the saved gpt output invalid but I'm inclined to let our linter change the test.

granawkins (approved):
It runs fine, I've just opened a big refactor to integrate the sample more so I'm fine merging this as-is.

PR #490, opened 2024-01-18
Integrate benchmark and sampler

granawkins:
This makes use of the Sampler as much as possible for benchmarks. Basically there's now a `class Benchmark`, which has a list of samples, and it evaluates each with `run_sample`. 

## Pull Request Checklist


benchmarks/benchmark_runner.py
```diff
@@ -116,11 +123,8 @@ async def compare_diffs(actual, generated):
     return await grade(prompt, comparison_prompt)
 
 
-async def grade_and_clean_diff(repo, response, result, comparison_diff=None):
+async def grade_and_clean_diff(diff, response, result, comparison_diff=None):
```
jakethekoenig, line 126: This doesn't clean anymore. I think we should rename to `grade_diff`. I like the change to having setup clean instead of teardown. It's good to have the benchmark repo easily inspectable after a benchmark.

jakethekoenig (approved):
Thanks so much! I really like the consolidation here between Sample and Modules. It makes things a lot cleaner. 3 small notes.

PR #492, opened 2024-01-18
Active diff snapshot not automatically run

jakethekoenig:
It's not very safe to be messing with git at startup. There are a lot of edge cases. Most users won't use the sampler at all and even with this change the sampler will still work well assuming the changes being made aren't related to the active diff.

## Pull Request Checklist


jakethekoenig:
The user said they had a workaround with .gitignore. I'll wait to merge until Grant's seen it.

PR #495, opened 2024-01-18
fix azure by removing model checking

### LINKED ISSUE #489: Error when deployment_name is different from model name in Azure
Microsoft is a genius in making simple things overcomplicated and confusing. This happened again that the deployment name and model name can be different in Azure OpenAI. Is it possible to allow adding an additional key `deployment_name` to `~/.mentat/.mentat_config.json`?  For example, my model name is `gpt-3.5-turbo` but my `depoyment_name` is `gpt-35-turbo`. This will be troublesome because currently mentat only allow using one variable `model`. If I set `model="gpt-3.5-turbo"`, the azure chat.completions.create() [call](https://github.com/AbanteAI/mentat/blob/main/mentat/llm_api_handler.py#L386) will fail because there is no deployment of `"gpt-3.5-turbo"`. However if I set `model="gpt-35-turbo"`, the _validate_key() [call ](https://github.com/AbanteAI/mentat/blob/main/mentat/llm_api_handler.py#L204)will get an error saying `"gpt-35-turbo"` doesn't exist in known model key.

Note: the culprit is really Azure's stupid way of defining the argument `model` in `client.chat.completions.create` is using deployment_name rather than model_name. See Azure's OpenAI 1.x [example](https://learn.microsoft.com/en-us/azure/ai-services/openai/how-to/migration?tabs=python-new%2Cdalle-fix).

PCSwingle:
Fixes #489 

## Pull Request Checklist

### DIFF
--- a/mentat/llm_api_handler.py
+++ b/mentat/llm_api_handler.py
@@ -42,7 +42,7 @@ from openai.types.chat import (
 from openai.types.chat.completion_create_params import ResponseFormat
 from PIL import Image
 
-from mentat.errors import MentatError, ModelError, ReturnToUser, UserError
+from mentat.errors import MentatError, ReturnToUser, UserError
 from mentat.session_context import SESSION_CONTEXT
 from mentat.utils import mentat_dir_path
 
@@ -184,9 +184,7 @@ class ModelsIndex(Dict[str, Model]):
 
     def _validate_key(self, key: str) -> str:
         """Try to match fine-tuned models to their base models."""
-        if super().__contains__(key):
-            return key
-        if key.startswith("ft:"):
+        if not super().__contains__(key) and key.startswith("ft:"):
             base_model = key.split(":")[
                 1
             ]  # e.g. "ft:gpt-3.5-turbo-1106:abante::8dsQMc4F"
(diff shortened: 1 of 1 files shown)


biobootloader:
If the user sets the model name as `gpt-35-turbo` (instead of `gpt-3.5-turbo`, like in the issue that prompted this), then will we not have cost estimates for them?

PCSwingle:
> If the user sets the model name as `gpt-35-turbo` (instead of `gpt-3.5-turbo`, like in the issue that prompted this), then will we not have cost estimates for them?

We won't but there's no way to do that; you name the azure model when you set it up, so it could be anything.

jakethekoenig:
Two things that might make sense to do (not as part of this pr):
1. Let the user set the costs in the config. Sort of like how they can set maximum_context.
2. Guess what prices we should use by embedding their model name and seeing what model it's closest to.

biobootloader:
how about let the user set a base model to use for cost estimate as well as the model name to call?

so there'd be an extra variable like `azure_deployment_true_model_name` or something. If they don't set it then they'd get a warning when using azure that we don't know model limits or costs unless they set that

PCSwingle:
I think this adds too much complexity for no good reason; if anything, we would want to add a config option to set the input and output token prices (but I also don't think we should do that since it's too complicated for so little gain).

biobootloader (approved):

PR #498, opened 2024-01-19
QOL updates from running benchmark samples

granawkins:
Changes that I made while working with samples/benchmarks.

### DIFF
--- a/benchmarks/arg_parser.py
+++ b/benchmarks/arg_parser.py
@@ -25,6 +25,13 @@ def common_benchmark_parser():
         ),
     )
     parser.add_argument(
+        "--directory",
+        "-d",
+        default="benchmarks/benchmarks",
+        type=str,
+        help="Directory from which to load benchmarks and/or samples",
+    )
+    parser.add_argument(
         "--max_benchmarks",
         default=1,
         type=int,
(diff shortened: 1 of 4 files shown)


benchmarks/arg_parser.py
```diff
@@ -24,6 +24,13 @@ def common_benchmark_parser():
             " depends on benchmark."
         ),
     )
+    parser.add_argument(
+        "--directory",
```
PCSwingle, line 28: Could we makes this --sample-directory? I think directory is too vague for this.

benchmarks/benchmark_result.py
```diff
@@ -83,5 +86,5 @@ def to_json(self):
         return json.dumps(self.to_dict())
 
     @classmethod
-    def from_json(cls, json_str):
+    def from_json(cls, json_str: str) -> "BenchmarkResult":
```
PCSwingle, line 89:
```suggestion
    def from_json(cls, json_str: str) -> Self:
```

jakethekoenig: Looks like this isn't a feature in python 3.11?

jakethekoenig (approved):

benchmarks/benchmark_result.py
granawkins: ya fwiw this broke the macos tests, I switch to `from __future__`..

PR #502, opened 2024-01-22
Add Textual

PCSwingle:
Adds textual

## Pull Request Checklist


waydegg:
image

I think the default background/theme should match whatever the terminal has its theme set to. Pretty sure there's a way to do this easily with Textual.

waydegg:
image
Should we be rendering the code blocks with syntax highlighting? Also Textual can render Markdown (for the model response that isn't a code edit) which could look nice.

I'm using the default Macos terminal here btw

waydegg:
image
Seeing weird line breaks on the sides of the input box, guessing it's due to the Macos terminal not being TrueColor and/or not supporting ligatures

PCSwingle:
>  Seeing weird line breaks on the sides of the input box, guessing it's due to the Macos terminal not being TrueColor and/or not supporting ligatures

Oh yeah probably since I don't have those.

> Should we be rendering the code blocks with syntax highlighting? Also Textual can render Markdown (for the model response that isn't a code edit) which could look nice.

We never render code blocks that aren't code edits; it would definitely be a nice feature though. I'll look into it; the only problem I can think of right now is that we're streaming the response, so it might behave weirdly (like it starts out streaming and you see the ```tsx up until it finishes the block, and then it finally renders everything). I'll test it out but I think I agree that it'll be better than what it currently is.

PCSwingle:
> I think the default background/theme should match whatever the terminal has its theme set to. Pretty sure there's a way to do this easily with Textual.

I think you're right and I'll look into it; my textual is always full screen though; why is yours not? Is it because you're using tmux? Either way I'll try and figure this out.

PCSwingle:
@waydegg https://textual.textualize.io/FAQ/#why-doesnt-textual-look-good-on-macos
Maybe we should add this to the docs?

waydegg:
>https://textual.textualize.io/FAQ/#why-doesnt-textual-look-good-on-macos
Maybe we should add this to the docs? Can you let me know how much it fixes?

Much better! Def worth adding in the docs 
image

waydegg:
>I think you're right and I'll look into it; my textual is always full screen though; why is yours not? Is it because you're using tmux? Either way I'll try and figure this out.

Yeah full screen looks great, and yup using Tmux so it looks a bit off. It looks like there might be a way to do this with Rich: https://rich.readthedocs.io/en/stable/syntax.html?highlight=terminal%20theme#background-color

The creator of Textual just opened an issue a couple days ago talking about this actually: https://github.com/Textualize/textual/issues/4119

waydegg:
>Would be nice to get some kind of "loading" spinner or progress bar after I submit a prompt

A couple users mentioned wanting this in the Discord server too. Worth adding later imo

mentat/terminal/terminal_app.py
```diff
@@ -0,0 +1,307 @@
...
+        if auto_context_tokens > 0:
+            context_header += (
+                f"\nAuto-Context: Enabled\nAuto-Context Tokens: {auto_context_tokens}"
+            )
+        else:
+            context_header += "\nAuto-Context: [yellow]Disabled[/yellow]"
+
+        context_header += "\nIncluded Files:"
+        if not features:
+            context_header += " [yellow]None[/yellow]"
```
waydegg, line 216: Can all of these inline colors/styles be send as `extra` items in the session stream? Would be easier to parse for the VSCode extension (and for any other extensions building ontop of Mentat)

mentat/terminal/patched_autocomplete.py
```diff
@@ -0,0 +1,191 @@
...
+
+    # By default the dropdown can only be below the input, but since our input is always at the bottom of the screen
+    # we want the opposite, so we patch this function to drop up instead of down
+    @override
+    def reposition(
+        self,
+        input_cursor_position: int | None = None,
+        scroll_target_adjust_y: int = 0,
+    ) -> None:
+        if self.input_widget is None:  # pyright: ignore
```
waydegg, line 165: Assuming we're ignoring IncompatibleOverride errors here?

PCSwingle: I think the one being ignored here is that self.input_widget has a type of Input, not Input | None (of course it could still be None if it isn't set, which is what we're checking for); there are a ton of pyright ignores in this file because I was mostly just copying the code from the classes I patched the code from and I didn't want to change it.

PCSwingle:
Just finished looking into all of your comments;

1. I tried displaying Markdown; it came with a lot of problems though. For example, a lot of the time the model wouldn't close off its markdown ``` tags and all the text after that would be marked down. Also, markdown doesn't support color. Overall I might be able to fix some of these problems, but it seems like too much work for very little gain to me (how often does the model output markdown formatted text anyways? Not too often).

2. I don't think there's a way to check the background of the terminal, but I did hook up the config.theme to Textual, so setting your theme to dark or light mode will set the background. I think this is fine, as most people will be opening it up in full screen, so it will look normal, and they can set their config to dark mode if they use a dark mode terminal (if they open it up in a non full screen mode somehow then hopefully they can figure it out haha). 

3. Added MacOS bug fix to docs!

4.

waydegg (approved):

mentat/command/commands/context.py
jakethekoenig, line 1: Can you delete `/context` from the [docs](https://docs.mentat.ai/en/latest/user/commands.html) too? Sorry that page should probably be generated from this dir.

jakethekoenig (approved):
Looks great to me! I think the following would be nice but not necessary to ship:
* x buttons in context pane to remove files.
* A microphone emoji in the entry box to trigger `/talk`
* The ability to change theme midsession with `/config`. The UX here is strange because presumably users will almost never switch back and forth. I wonder if there should be a sticky `/config` that saves it to the config or if some settings should be sticky and others not.
* This pr breaks reverse search. It'd be nice to get that back.

I'll keep using it this morning and let you know if I notice anything.

biobootloader:
when you exit should it display something in the terminal? like total conversation cost? 

image

biobootloader:
I can't select text (like copying the above error). Looks like this requires holding option / shift, but those don't work for me. Does option work for anyone else on a mac? maybe my window manager is messing it up - option moves the window around

image

granawkins:
> I can't select text (like copying the above error). Looks like this requires holding option / shift, but those don't work for me. Does option work for anyone else on a mac? maybe my window manager is messing it up - option moves the window around

Also not working for me on mac - tried lots of combinations, can't select/copy.

granawkins (commented):
Few small comments. My wish list would be:
1. Loading Bar issue
2. Copiable text
3. Dark theme (wasn't working out-of-the-box for me but maybe I'm missing sth?)

PCSwingle:
> Looks great to me! I think the following would be nice but not necessary to ship:
> 
> * x buttons in context pane to remove files.

This would be pretty difficult to do; I don't even know if it's possible. Right now I use the Tree textual widget and I looked at the docs and couldn't find anything that would do this. I'll make an issue though and I can look into it later, because I really do like this idea.

> * A microphone emoji in the entry box to trigger `/talk`

Would probably be easier, but I also think it would be good in another PR since I'm not entirely sure yet how I would do it. I'll put this in the same issue as the x buttons.

> * The ability to change theme midsession with `/config`. The UX here is strange because presumably users will almost never switch back and forth. I wonder if there should be a sticky `/config` that saves it to the config or if some settings should be sticky and others not.

This would be pretty nice; the reason I disabled mid session change is because then we need something watching the config to see when it changes and alert the terminal app (not because it's actually difficult to switch the theme midsession). A sticky config would be nice but then the question is where we put the stored config; we have like 3 different config files. Not sure how to address this, but I'll make another issue for it.

> * This pr breaks reverse search. It'd be nice to get that back.

I don't think this is the biggest deal and it would be very difficult to add (would have to program in our own reverse search). 

> I can't select text (like copying the above error). Looks like this requires holding option / shift, but those don't work for me. Does option work for anyone else on a mac? maybe my window manager is messing it up - option moves the window around

I can't really test this without a mac; if somebody else could try to figure this out and fix it that would be great. Thanks!

mentat/terminal/patched_autocomplete.py
```diff
@@ -0,0 +1,191 @@
...
+class CompletionDropdownItem(DropdownItem):
+    """
+    A dropdown item with a full completion to carry all of the information we need
+    """
+
+    content: str = ""
+    position: int = 0
+
+
+class PatchedDropdown(Dropdown):
```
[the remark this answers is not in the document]
PCSwingle: We can't change the text size (since we're working in a terminal), so I don't think this would end up looking very good. Maybe a different key would be better than tab?

PCSwingle:
> 3. Dark theme (wasn't working out-of-the-box for me but maybe I'm missing sth?)

What problems were you having with it? Did you change a config file or run with --theme dark?

granawkins:
> > 3. Dark theme (wasn't working out-of-the-box for me but maybe I'm missing sth?)
> 
> What problems were you having with it? Did you change a config file or run with --theme dark?

Ah ya I used `--theme dark` and it goes. I saw Wayde said it'd be nice if it picked up your theme automatically, so I was echoing that. But also, low priority.

mentat/terminal/patched_autocomplete.py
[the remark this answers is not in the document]
granawkins: could it just be Enter? that's what I do instinctually, and it's a lot of work to recover.

biobootloader:
Ok I figured out why holding Option key wasn't working for me - my window manager Yabai moves windows when you click on them while holding option. I turned Yabai off and selecting text now works - but selecting multiple lines gets you context from the right side, which isn't nice:

image

I'm going to go ahead and merge this though!

PR #508, opened 2024-01-31
Remove colored and highlight

PCSwingle:
Removes all uses of colored and highlight so that all coloring is handled entirely by the client (which is necessary for the Textual client). Ideally we eventually move completely over to styles instead of colors, but this is a step in that direction.

## Pull Request Checklist


mentat/edit_history.py
```diff
@@ -53,14 +51,15 @@ async def redo(self) -> Optional[str]:
...
-            return colored("No edits available to undo", color="light_red")
+            return ["No edits available to undo"]
 
         errors = list[str]()
         while self.edits:
             error = self.undo()
             if error:
-                errors.append(error)
-        return "\n".join(errors)
+                errors += error
```
waydegg, line 64: Should this still be `.append` (since it's a list)?

mentat/parsers/change_display_helper.py
```diff
@@ -108,16 +109,18 @@ def _get_code_block(
...
-):
+) -> FormattedString:
     lines = _prefixed_lines(line_number_buffer, code_lines, prefix)
-    if lines:
-        return "\n".join(colored(line, color=color) for line in lines.split("\n"))
+    if color is None:
+        return lines
     else:
-        return ""
+        return (lines, {"color": color})
```
waydegg, line 117: The return type is a `str` however a tuple is returned here

waydegg, line 117: `FormattedString` is the return type however a tuple is returned here

mentat/parsers/change_display_helper.py
```diff
@@ -134,42 +137,42 @@ def get_full_change(display_information: DisplayInformation, prefix: str = ""):
...
         ),
     ]
-    full_change = "\n".join([line for line in to_print if line])
-    prefixed_change = "\n".join(
-        (prefix + line) if line.strip() else line for line in full_change.split("\n")
-    )
-    return prefixed_change
+    for line in full_change:
+        if (isinstance(line, str) and line.strip()) or (
+            isinstance(line, Tuple) and line[0].strip()
```
waydegg, line 142: Is `line.strip()` in the if statement just checking if the line is empty or not? Might be more readable to do `line == ""` or `line != ""`

mentat/edit_history.py
PCSwingle: error is also a list

mentat/parsers/streaming_printer.py
```diff
@@ -0,0 +1,78 @@
+import asyncio
+from collections import deque
+from typing import Any, Dict, List, Tuple
+
+from mentat.session_context import SESSION_CONTEXT
+
+FormattedString = str | Tuple[str, Dict[str, Any]] | List[Tuple[str, Dict[str, Any]]]
```
waydegg, line 7: Might be better to make `FormattedString` an actual class instead of a union of all these types.

mentat/parsers/change_display_helper.py
PCSwingle: We want to make sure it has stuff other than just whitespace; otherwise I would just do if line

waydegg (approved):
Left a few comments/suggestions, otherwise LGTM

mentat/parsers/streaming_printer.py
PCSwingle: Maybe; I'll add a TODO but I want to get this PR merged quickly because I need it in order to work on my textual PR.

PR #511, opened 2024-02-01
add code file to daemon prompt

waydegg:
Adds the file contents the `@mentat` comment is in to the user prompt.


biobootloader (approved):
good to merge this so we can play around with it!

ideally it'd be integrated with the auto-context system to get more context

PR #515, opened 2024-02-02
Python client default mentat export

jakethekoenig:
You can now import the PythonClient via `from mentat import Mentat`.
* Docs updated
* Comments added to python_client
* tests and benchmarks updated to use new import

## Pull Request Checklist

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -4,8 +4,6 @@ from pathlib import Path
 import pkg_resources
 from setuptools import find_packages, setup
 
-from mentat import __version__
-
 readme_path = os.path.join(Path(__file__).parent, "README.md")
 with open(readme_path, "r", encoding="utf-8") as f:
     long_description = f.read()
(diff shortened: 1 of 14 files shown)


setup.py
```diff
@@ -4,16 +4,14 @@
...
-
 readme_path = os.path.join(Path(__file__).parent, "README.md")
 with open(readme_path, "r", encoding="utf-8") as f:
     long_description = f.read()
 
 
 setup(
     name="mentat",
-    version=__version__,
+    version="1.0.9",
```
waydegg, line 14: Why do we hardcode this over using `__version__`?

waydegg (approved):

setup.py
jakethekoenig: The reason I made this change was because we can't import from mentat in setup anymore because the default export will fail as requirements are not installed yet. The number has to be hardcoded somewhere and I thought here made sense but also considered putting it in a VERSION file which I've seen some packages do.

I'm down to do it a different way if you have opinions.

PR #519, opened 2024-02-12
Changelog added to docs and displayed

jakethekoenig:
I added CHANGELOG.rst to our project root and made it a page on the docs. If the user is using an outdated version the most recent changelog is displayed as part of the prompt to upgrade.

## Pull Request Checklist

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -6,7 +6,7 @@ In this changelog focus on user facing highlights and stick to the format. This 
 `1.0.9 <https://pypi.org/project/mentat/1.0.9/>`__
 --------------------------------------------------
 
-- Adds `/ammend` command: clear last message and prefill with last prompt.
+- Adds `/amend` command: clear last message and prefill with last prompt.
 - Experimental feature revisor. Turn on with `--revisor` flag. Attempts to fix edits that fail to conform to parser format.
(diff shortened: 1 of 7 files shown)


CHANGELOG.rst
```diff
@@ -0,0 +1,26 @@
+Changelog
+=========
+
+In this changelog focus on user facing highlights and stick to the format. This information will be used to motivate users to upgrade or after upgrading to inform them of features that might otherwise not be very discoverable.
+
+`1.0.9 <https://pypi.org/project/mentat/1.0.9/>`__
+--------------------------------------------------
+
+- Adds `/ammend` command: clear last message and prefill with last prompt.
```
PCSwingle, line 9: Typo

PCSwingle (approved):
Thanks for adding this! Looks great, although I think it would be nice to print the changelog for all versions after the current version (rather than just the most recent version). Would this be too difficult?

jakethekoenig:
> Thanks for adding this! Looks great, although I think it would be nice to print the changelog for all versions after the current version (rather than just the most recent version). Would this be too difficult?

I could do that. I worry it'd be strange if the user hasn't used it for a few versions to print a huge wall of text though. And in general we release infrequently enough weekly users shouldn't miss changelogs printed with this feature.

PR #531, opened 2024-02-22
Diff Context never null

jakethekoenig:
When run in a non git directory the diff context was set to null which made interacting with it complex. Now the diff context is responsible for determining it is not in a git context and returning sensible empty/none types when called.

## Pull Request Checklist

### DIFF
--- a/mentat/code_context.py
+++ b/mentat/code_context.py
@@ -53,21 +53,16 @@ class CodeContext:
     def __init__(
         self,
         stream: SessionStream,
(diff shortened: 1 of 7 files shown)


jakethekoenig:
I couldn't add the bot to this repo but you can see its review [here](https://github.com/jakethekoenig/mentat/pull/2) It actually tried to comment on a line that wasn't changed so not all the model comments are displayed.

PR #533, opened 2024-02-23
Conversation get_messages gets code_message

jakethekoenig:
Frequently in the codebase people will go to conversation for the messages and then independently get the code_message. It's simpler if the conversation handles that.

get_messages is also changed to accept a system prompt argument instead of having a include_system_prompt parameter to simplify agent's calling of it. If callers genuinely want no system prompt they can pass `[]`.

Move responsibility to raise if context too big from code_context to conversation.

## Pull Request Checklist

### DIFF
--- a/mentat/code_context.py
+++ b/mentat/code_context.py
@@ -28,7 +28,6 @@ from mentat.interval import parse_intervals, split_intervals_from_path
 from mentat.llm_api_handler import (
     count_tokens,
     get_max_tokens,
-    prompt_tokens,
     raise_if_context_exceeds_max,
 )
 from mentat.session_context import SESSION_CONTEXT
@@ -85,8 +84,7 @@ class CodeContext:
         git_diff_paths = [str(p) for p in self.diff_context.diff_files()]
         git_untracked_paths = [str(p) for p in self.diff_context.untracked_files()]
 
-        messages = await ctx.conversation.get_messages(include_code_message=True)
-        total_tokens = prompt_tokens(messages, ctx.config.model)
+        total_tokens = await ctx.conversation.count_tokens(include_code_message=True)
 
         total_cost = ctx.cost_tracker.total_cost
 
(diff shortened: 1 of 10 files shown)


mentat/code_context.py
```diff
@@ -154,29 +129,29 @@ async def get_code_message(
...
         code_message += ["Code Files:\n"]
-        meta_tokens = count_tokens("\n".join(code_message), model, full_message=True)
 
         # Calculate user included features token size
         include_features = [
             feature
             for file_features in self.include_files.values()
             for feature in file_features
         ]
-        include_files_message = get_code_message_from_features(include_features)
```
jakethekoenig, line 165: I moved all this to the if because it's only relevant if auto_context is on. Benchmarking shows its basically unimportant for performance but I think its easier to see how things are used and the flow of logic this way.

mentat/conversation.py
```diff
@@ -246,29 +271,10 @@ async def get_model_response(self) -> ParsedLLMResponse:
...
-                if isinstance(prompt, str)
-                else ""
-            ),
-        )
-        messages_snapshot.insert(
-            0 if config.no_parser_prompt else 1,
-            ChatCompletionSystemMessageParam(role="system", content=code_message),
-        )
+        messages_snapshot = await self.get_messages(include_code_message=True)
+        tokens_used = prompt_tokens(messages_snapshot, config.model)
```
granawkins, line 276:
```suggestion
        tokens_used = await self.count_tokens(include_code_message=True)
```
Can this new method be reused here?

granawkins (approved):
I like removing all the token checks from the context-building process (moving to llm_api_handler instead), and the conversation.count_tokens method. Made one small suggestion, otherwise LGTM. 🚀

mentat/session.py
```diff
@@ -151,7 +151,7 @@ async def _main(self):
         stream.send("Type 'q' or use Ctrl-C to quit at any time.")
```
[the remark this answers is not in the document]
biobootloader: chose the wrong line to comment on here?

mentat/conversation.py
jakethekoenig: It could be used but we still need to compute `messages_snapshot` to send to the llm anyway. So I think it is better as is.

PR #534, opened 2024-02-24
Add `load`/`save` context commands

### LINKED ISSUE #521: save/load context
image

galer7:
resolves #521

### DIFF
--- a/mentat/code_context.py
+++ b/mentat/code_context.py
@@ -437,6 +437,7 @@ class CodeContext:
 
     def to_simple_context_dict(self) -> dict[str, list[str]]:
         """Return a simple dictionary representation of the code context"""
+
         simple_dict: dict[str, list[str]] = {}
         for path, features in self.include_files.items():
             simple_dict[path.absolute().as_posix()] = [
@@ -447,10 +448,17 @@ class CodeContext:
     def from_simple_context_dict(self, simple_dict: dict[str, list[str]]):
         """Load the code context from a simple dictionary representation"""
 
-        self.include_files = {}
         for path_str, features_str in simple_dict.items():
             path = Path(path_str)
-            code_features = [CodeFeature.from_string(f_str) for f_str in features_str]
-            self.include_files[path] = code_features
+            features_for_path: List[CodeFeature] = []
 
-        self.refresh_context_display()
+            for feature_str in features_str:
+                feature_path = Path(feature_str)
+
+                # feature_path is already absolute, so cwd doesn't matter
+                current_features = get_code_features_for_path(
+                    feature_path, cwd=Path("/")
+                )
+                features_for_path += list(current_features)
+
+            self.include_files[path] = features_for_path
--- a/mentat/code_context.py
+++ b/mentat/code_context.py
@@ -440,9 +440,7 @@ class CodeContext:
 
         simple_dict: dict[str, list[str]] = {}
         for path, features in self.include_files.items():
-            simple_dict[path.absolute().as_posix()] = [
-                str(feature) for feature in features
-            ]
+            simple_dict[str(path.absolute())] = [str(feature) for feature in features]
         return simple_dict
 
     def from_simple_context_dict(self, simple_dict: dict[str, list[str]]):
--- a/mentat/code_context.py
+++ b/mentat/code_context.py
@@ -434,3 +434,21 @@ class CodeContext:
             return all_features_sorted
         else:
             return all_features_sorted[:max_results]
+
+    def to_simple_dict(self) -> dict[str, list[str]]:
+        """Return a simple dictionary representation of the code context"""
+        simple_dict: dict[str, list[str]] = {}
+        for path, features in self.include_files.items():
+            simple_dict[path.absolute().as_posix()] = [
+                str(feature) for feature in features
+            ]
+        return simple_dict
+
+    def from_simple_dict(self, simple_dict: dict[str, list[str]]):
+        """Load the code context from a simple dictionary representation"""
+        self.include_files = {}
+        for path, features in simple_dict.items():
+            self.include_files[Path(path)] = [
+                CodeFeature(Path(feature)) for feature in features
+            ]
+        self.refresh_context_display()
--- a/mentat/code_context.py
+++ b/mentat/code_context.py
@@ -435,7 +435,7 @@ class CodeContext:
         else:
             return all_features_sorted[:max_results]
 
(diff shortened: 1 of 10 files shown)


mentat/code_context.py
```diff
@@ -434,3 +434,23 @@ async def search(
...
+    def from_simple_context_dict(self, simple_dict: dict[str, list[str]]):
+        """Load the code context from a simple dictionary representation"""
+
+        self.include_files = {}
+        for path_str, features_str in simple_dict.items():
+            path = Path(path_str)
+            code_features = [CodeFeature.from_string(f_str) for f_str in features_str]
+            self.include_files[path] = code_features
+
+        self.refresh_context_display()
```
jakethekoenig, line 456: This is done once in the user input loop so it's unnecessary here.

mentat/code_feature.py
```diff
@@ -180,6 +180,28 @@ def count_tokens(self, model: str) -> int:
...
+    @staticmethod
+    def from_string(string: str) -> CodeFeature:
+        """
+        Create a CodeFeature from a string.
+        """
+
+        print("in CodeFeature.from_string", string)
+
+        # find last colon
+        colon_index = string.rfind(":")
```
jakethekoenig, line 192: The windows tests are failing because their paths start with `C:`. In general its not safe to split on ":" because paths can contain that character.

You can use `get_code_feratures_for_path` from `mentat/include_files.py` to do this.

mentat/session_stream.py
```diff
@@ -34,8 +34,8 @@ class SessionStream:
     default: Any data sent to the client over this channel should be displayed. Valid kwargs: color, style
 
     *session_exit: Sent by the client, suggesting that the session should exit whenever possible.
-    client_exit: Sent by the server, client should shut down when recieved.
-    session_stopped: Sent by the server directly before server shuts down. Server can't be contacted after recieved.
+    client_exit: Sent by the server, client should shut down when received.
```
jakethekoenig, line 37: Thanks

mentat/command/commands/save.py
```diff
@@ -0,0 +1,61 @@
...
+            try:
+                context_file_path = Path(args[0]).expanduser().resolve()
+            except RuntimeError as e:
+                raise PathValidationError(
+                    f"Invalid context file path provided: {args[0]}: {e}"
+                )
+
+        if len(args) == 0:
+            stream.send(
+                "No context file specified. Defaulting to context.json", style="warning"
```
jakethekoenig, line 36: In this command and in load I don't think its necessary to log a warning because we log where the context was saved in the end.

mentat/command/commands/load.py
```diff
@@ -0,0 +1,63 @@
...
+                context_file_path = Path(args[0]).expanduser().resolve()
+            except RuntimeError as e:
+                raise PathValidationError(
+                    f"Invalid context file path provided: {args[0]}: {e}"
+                )
+
+        with open(context_file_path, "r") as file:
+            parsed_include_files = json.load(file)
+
+        # TODO: Do we remove already-included files when loading new context file?
```
jakethekoenig, line 43: I don't think this is necessary. The user can always run `/exclude *` first or restart mentat if that's what they want. But if we make the default the other way there won't be any easy way to add the context from the file to context.

jakethekoenig (commented):
Thanks so much for doing this. It's looking pretty good. Left a comment on how to fix the windows test.

galer7:
Thanks for the review! I'll fix it this week

mentat/code_feature.py
galer7: Ah perfect, I was imagining you already have something for this but didn't find it when I first wrote the code

jakethekoenig (commented):
I commented the line which is causing the windows problem. After that change the tests should pass on windows. Both `/save` and `/load` work great for me, thanks for doing this. I have one note around error handling. But after those two things are done I think this should be good to merge.

mentat/command/commands/load.py
```diff
@@ -0,0 +1,59 @@
...
+            parsed_include_files = json.load(file)
+
+        code_context.from_simple_context_dict(parsed_include_files)
+
+        stream.send(f"Context loaded from {context_file_path}", style="success")
+
+    @override
+    @classmethod
+    def arguments(cls) -> List[CommandArgument]:
+        return [CommandArgument("required", ["path"])]
```
jakethekoenig, line 47:
```suggestion
        return [CommandArgument("optional", ["path"])]
```

jakethekoenig (commented):
One more thing. Also sorry about the butler bot. It's a work in progress.

Also it would be great if you updated `docs/source/user/commands.rst` with the new commands. I need to make that file autogenerated from the code but haven't gotten around to it.

galer7:
[Looks like an unlucky run](https://github.com/AbanteAI/mentat/actions/runs/8128131923/job/22213695406#step:8:9).

I found the same issue here: https://github.com/RobertCraigie/pyright-python/issues/162. Proposed fix is to run `brew install npm`, so I guess I'll just add a `actions/setup-node@v4` step in the CI.

biobootloader (approved):
Thank you for this! Looks great and works well for me.

I like that the behavior of load is to add to the current context, not replace it. Maybe the help message for the command should be explicit that it does that?

Maybe we should store context files somewhere other than the project root by default? Perhaps a `.mentat` directory we'd create in the project? That way users could easily add `.mentat` to their `.gitignore`. Users could already do this by just running `/save .mentat/context_name` but that'd be more work. @jakethekoenig what do you think?

galer7:
I agree, a `.mentat` folder for each project definitely sounds like a better default

---
Future idea: a context file per each new git branch, since that would be a new feature & you work on different aspects of your project with each feature

jakethekoenig:
> I like that the behavior of load is to add to the current context, not replace it. Maybe the help message for the command should be explicit that it does that?

I agree.
 
> Maybe we should store context files somewhere other than the project root by default? Perhaps a `.mentat` directory we'd create in the project? That way users could easily add `.mentat` to their `.gitignore`. Users could already do this by just running `/save .mentat/context_name` but that'd be more work. @jakethekoenig what do you think?

Context files aren't saved to the project root by default they're saved to `~/.mentat/context.json`. I think if the user does specify a name it would be confusing to append anything to it.

biobootloader:
> Context files aren't saved to the project root by default they're saved to ~/.mentat/context.json. I think if the user does specify a name it would be confusing to append anything to it.

Good point. Lets see how people use this. Maybe one context file is enough and naming it rare

PR #537, opened 2024-03-07
Anthropic message normalization

jakethekoenig:
System messages are converted to annotated user messages and repeated messages are concatenated.

This is sort of an ugly PR. I'd rather convert system messages to user messages natively and then have no difference between how gpt/claude are handled. I think a better philosophy than using system messages to inject information from the environment would be to tell the llm its talking to a bot that is relaying information from the end user and the environment and then have all system messages besides the parser prompt be user messages annotated with "Code message" and annotate the actual user messages as "End user message:".

But I kind of want to get this out quickly because it seems clear claude 3 is better than gpt-4 and I want to make it possible for other people to use it.

## Pull Request Checklist

### DIFF
--- a/docs/source/user/alternative_models.rst
+++ b/docs/source/user/alternative_models.rst
@@ -13,12 +13,36 @@ In addition, Mentat uses the :code:`gpt-4-1106-preview` model by default. When u
 .. warning::
     Due to changes in the OpenAI Python SDK, you can no longer use :code:`OPENAI_API_BASE` to access the Azure API with Mentat.
 
+Anthropic
+---------
+
+Mentat uses the OpenAI SDK to retrieve chat completions. This means that setting the `OPENAI_API_BASE` environment variable is enough to use any model that has the same response schema as OpenAI. To use models with different response schemas, we recommend setting up a litellm proxy as described `here <https://docs.litellm.ai/docs/proxy/quick_start>`__ and pointing `OPENAI_API_BASE` to the proxy. For example with anthropic:
+
+.. code-block:: bash
+
+    pip install 'litellm[proxy]'
+    export ANTHROPIC_API_KEY=sk-*************
+    litellm --model claude-3-opus-2024-0229 --drop_params
+    # Should see: Uvicorn running on http://0.0.0.0:8000
+
+.. code-block:: bash
+
+    # In ~/.mentat/.env
+    OPENAI_API_BASE=http://localhost:8000
+    # In ~/.mentat/.mentat_config.json
+    { "model": "claude" }
+    # or
+    export OPENAI_API_BASE=http://localhost:8000
+    mentat
+
+.. note::
+
+   Anthropic has slightly different requirements for system messages so you must set your model to a string with "claude" in it. Other than that it isn't important as the exact model is set by the litellm proxy server flag.
+
 🦙 Local Models
 ---------------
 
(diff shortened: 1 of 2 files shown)


docs/source/user/alternative_models.rst
```diff
@@ -13,12 +13,36 @@ In addition, Mentat uses the :code:`gpt-4-1106-preview` model by default. When u
 .. warning::
     Due to changes in the OpenAI Python SDK, you can no longer use :code:`OPENAI_API_BASE` to access the Azure API with Mentat.
 
+Anthropic
+---------
+
+Mentat uses the OpenAI SDK to retrieve chat completions. This means that setting the `OPENAI_API_BASE` environment variable is enough to use any model that has the same response schema as OpenAI. To use models with different response schemas, we recommend setting up a litellm proxy as described `here <https://docs.litellm.ai/docs/proxy/quick_start>`__ and pointing `OPENAI_API_BASE` to the proxy. For example with anthropic:
```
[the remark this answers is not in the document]
biobootloader: Isn't the example below with `OPENAI_API_BASE=http://localhost:8000` sufficient?

docs/source/user/alternative_models.rst
```diff
@@ -13,12 +13,36 @@ In addition, Mentat uses the :code:`gpt-4-1106-preview` model by default. When u
...
+    OPENAI_API_BASE=http://localhost:8000
+    # In ~/.mentat/.mentat_config.json
+    { "model": "claude" }
+    # or
+    export OPENAI_API_BASE=http://localhost:8000
+    mentat
+
+.. note::
+
+   Anthropic has slightly different requirements for system messages so you must set your model to a string with "claude" in it. Other than that it isn't important as the exact model is set by the litellm proxy server flag.
```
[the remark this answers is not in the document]
biobootloader: the users don't need to worry about how the system messages are handled, we handle that. the important thing here is that the users set the model to contain the name "claude" in it, so that know it's an anthropic model and can modify how we do system messages. Given that info, can you suggest an alternative for this line? us the suggestion syntax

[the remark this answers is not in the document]
biobootloader:
```suggestion
   To ensure compatibility with Anthropic models, it's crucial to include "claude" in your model's name. This naming convention signals Mentat to apply specific handling for system messages, aligning with Anthropic's requirements. You'll set the exact model when you start the llm proxy.
```

mentat/llm_api_handler.py
```diff
@@ -134,6 +137,74 @@ def count_tokens(message: str, model: str, full_message: bool) -> int:
...
+    elif current_role == "system":
+        concatenate_adjacent.append(
+            ChatCompletionSystemMessageParam(role=current_role, content=current_content)
+        )
+    elif current_role == "assistant":
+        concatenate_adjacent.append(
+            ChatCompletionAssistantMessageParam(
+                role=current_role, content=current_content
+            )
+        )
```
biobootloader, line 203: @abanteai-butler can you refactor, putting the if/elif/elif to append messages into a subfunction that's called in the loop and after the loop ends?

biobootloader: please write the suggestion in github suggestion syntax

biobootloader: you have to rewrite everything, can't just say `# Then replace the if/elif/elif blocks in the loop and after the loop with calls to this function.` or when I accept the suggestion it will overwrite everything

biobootloader: not quite, the suggestion is just replacing the code starting at the line `concatenate_adjacent = list[ChatCompletionMessageParam]()`, so you don't need to include the code from before that line, that will remain

biobootloader: you didn't actually make the suggestion

biobootloader (approved):
Looks good! I agree this should be temporary and I like your thoughts on a better approach.

PR #540, opened 2024-03-17
Ragdaemon

granawkins:
This aims to be the simplest possible integration of ragdaemon into mentat:
- CodeContext.daemon is added, updated initially (async) in Session._main, and then updated on each call to CodeContext.get_code_message. It replaces everything to do with filters (including embeddings).
- `include_files`, `CodeFeature` and `code_file_manager` are relatively undisturbed for simplicity. In future PRs we might refactor these to take better advantage of ragdaemon.
- By default, I mock out daemon.update for tests. For tests quick require full context, I added the `@pytest.mark.ragdaemon` flag, which will use (and cache) embeddings with chroma. Longer term this should also be mocked out, or tests of that functionality moved to the ragdaemon repo.
- Currently this doesn't chunk files. Ragdaemon does it with an LLM instead of ctags. I've turned it off for now (annotators doesn't include 'chunker'), but will aim to after some more optimization.

## Pull Request Checklist
- [x] Make ragdaemon a pip package
- [x] Remove unused files (filters, embeddings..)


mentat/code_context.py
```diff
@@ -116,41 +117,72 @@ async def get_code_message(
...
-            ]
+            code_message += [f"Diff References: {self.diff_context.name}\n"]
 
-        code_message += ["Code Files:\n"]
+        code_message += ["Code Files:\n\n"]
 
-        # Calculate user included features token size
-        include_features = [feature for file_features in self.include_files.values() for feature in file_features]
+        await self.daemon.update()
+        context_builder = self.daemon.get_context("", max_tokens=0)
```
granawkins, line 125: @PCSwingle - I had an idea about what you brought up on slack, re: included vs auto context. Check out this `ContextBuilder` class from Ragdaemon: https://github.com/AbanteAI/ragdaemon/blob/main/ragdaemon/context.py

You load paths, refs or diffs in, can add/remove with line numbers, then it has a `.render()` method that produces our `code_message`. Here, I load it up with `included_files`, then on line 167 I'll add auto-token files.

In a previous version of this commit, I made CodeContext.include_files a ContextBuilder straightaway, and maintained it like that. It has .add .remove methods to go with our commands. We could keep one, persistent ContextBuilder throughout the session, starting with your included_files (like I did before), and adding your auto-stuff to it, so you do just have one context. 

THis means your auto-context would be persistent, different from now, but I think that's a good thing. The *content* of the files your seeing should be up to date, but it doesn't need to re-select context after the first message.

granawkins:
Tests passing (but behind main). Todo here:
- [ ] redo Spice implementation w updates
- [ ] redirect to pip

Todo on `ragdaemon` before merging here:
- [ ] Redo `Spice` integration and mentat integration
- [ ] Reimplement daemon.load(), w/ checksum match
- [ ] Fix db records overlap/corruption issues
- [ ] Benchmark against Mentat's current auto-context
- [ ] Write documentation
- [ ] Add to pip

tests/code_context_test.py
```diff
@@ -202,22 +207,40 @@ def func_4(string):
...
-    async def _count_max_tokens_where(tokens_used: int) -> int:
-        code_message = await code_context.get_code_message(tokens_used, prompt="prompt")
-        return count_tokens(code_message, "gpt-4", full_message=True)
 
-    assert await _count_max_tokens_where(0) == 89  # Code
+    code_message = await code_context.get_code_message(0, prompt="prompt")
+    assert count_tokens(code_message, "gpt-4", full_message=True) == 95  # Code
+    assert (
+        code_message
+        == """\
```
jakethekoenig, line 220: nit: can you use dedent here?

mentat/diff_context.py
```diff
@@ -91,6 +91,9 @@ def annotate_file_message(code_message: list[str], annotations: list[DiffAnnotat
 
 
 class DiffContext:
+    target: str = ""
```
jakethekoenig, line 94: Does mentat use the diff for anything besides rendering annotations in the code message and putting information in the tree viewer? Should we totally delete this file and move the functionality to ragdaemon?

Also when I run `mentat --diff main` on this branch I get the following error:
image

jakethekoenig (commented):
Works for me and I appreciate the 600 line reduction. Also sorry about the butler mess I don't handle multiple commits very well right now.

I have a few notes but haven't had time to look at it closely.

PR #542, opened 2024-03-19
Fix Github Actions ctags install issue

biobootloader:
ctags installs were failing on github actions macos runners. Switching from `macos-latest` to `macos-14` seems to fix it (although I thought 14 was the latest??)

more details here: https://github.com/actions/runner-images/issues/4020

## Pull Request Checklist


biobootloader:
@abanteai-butler can you review again?

PR #571, opened 2024-04-18
add provider checking based on model name and provider

PCSwingle:


mentat/llm_api_handler.py
```diff
@@ -142,19 +155,48 @@ async def initialize_client(self):
...
+                from mentat.session_input import collect_user_input
+
+                match provider.name:
+                    case "open_ai" | "openai":
+                        env_variable = "OPENAI_API_KEY"
+                    case "anthropic":
+                        env_variable = "ANTHROPIC_API_KEY"
+                    case "azure":
+                        if os.getenv("AZURE_OPENAI_ENDPOINT") is None:
+                            raise MentatError(
```
jakethekoenig, line 183: I think we should prompt the user to enter the endpoint too.

mentat/llm_api_handler.py
```diff
@@ -142,19 +155,48 @@ async def initialize_client(self):
...
+                            f"No api key detected for provider {provider.name}. Create a .env file in ~/.mentat/.env or in your workspace root with your api key"
+                        )
+
+                ctx.stream.send(
+                    f"No {provider.name} api key detected. To avoid entering your api key on startup, create a .env file in"
+                    " ~/.mentat/.env or in your workspace root.",
+                    style="warning",
+                )
+                ctx.stream.send("Enter your api key:", style="info")
+                key = (await collect_user_input(log_input=False)).data
```
jakethekoenig, line 198: If I enter q here the program hangs

jakethekoenig (commented):
When I set my model to "test" and my provider to "openai" or "open_ai" I get the same exception
image

jakethekoenig:
This might not be totally connected to this PR but it would be nice if mentat didn't disable the user input box if they get a 404 from the api. For instance if I set my model to test I get the following (reasonable) error but then I can't enter anything into the box again
image

PCSwingle:
> This might not be totally connected to this PR but it would be nice if mentat didn't disable the user input box if they get a 404 from the api.

Agreed but it might be better to do in a separate PR

jakethekoenig (approved):
Looks good to me. I guess another thing to do in another PR is upgrade to 0.2 spice

PR #574, opened 2024-04-29
remove cron job for benchmarks

PCSwingle:


jakethekoenig:
I disabled it via the github UI:
image

